PR #158, opened 2025-02-24
Feat/fix appearance bugs

prachigarg19:
Fixes issues:
1. Replace branding logo everywhere with GoDAM logo

image

2. Fix color set by user not being displayed on hover over icons
3. Fix selected image for custom play button not being displayed


KMchaudhary (approved):
Screenshot 2025-02-24 at 6 36 31 PM

Can we use the default brand logo same as in above attachment background color, so that its looks good on every color theme.

You can use this logo as default brand logo
Screenshot 2025-02-24 at 6 35 49 PM

PR #179, opened 2025-02-28
Add Toggle button to disable DAM functionalities [develop]

Juzar10:
Screenshot 2025-02-28 at 5 43 57 PM


KMchaudhary (approved):
General settings section will only be visible if active license is there

PR #207, opened 2025-03-10
Add multisite plugin support

KMchaudhary:
- Update the use of REST API URLs across the plugin to support WordPress multisite
- Fix the retranscode bulk action for multisite setup

### DIFF
--- a/admin/godam-retranscode-admin.php
+++ b/admin/godam-retranscode-admin.php
@@ -287,16 +287,11 @@ class RetranscodeMedia {
 
 		$ids = implode( ',', $media );
 
-		$current_blog_id = null;
-		if ( is_multisite() ) {
-			$current_blog_id = get_current_blog_id();
-		}
-
 		// Can't use wp_nonce_url() as it escapes HTML entities.
 		$redirect_url = add_query_arg(
 			'_wpnonce',
 			wp_create_nonce( 'godam-tools' ),
-			get_admin_url( $current_blog_id, 'admin.php?page=godam-tools&goback=1&ids=' . $ids )
+			get_admin_url( get_current_blog_id(), 'admin.php?page=godam-tools&goback=1&ids=' . $ids )
 		);
 
 		wp_safe_redirect( $redirect_url );
--- a/admin/godam-retranscode-admin.php
+++ b/admin/godam-retranscode-admin.php
@@ -40,6 +40,13 @@ class RetranscodeMedia {
 	public $usage_info;
 
 	/**
+	 * Capability required to use this feature.
+	 * 
+	 * @var string
+	 */
+	public $capability;
+
+	/**
 	 * Functinallity initialization
 	 */
 	public function __construct() {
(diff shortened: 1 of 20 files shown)


assets/src/blocks/godam-player/render.php
```diff
@@ -101,7 +101,11 @@ function ( $layer ) {
 if ( ! empty( $ad_server ) ) :
 	$ad_tag_url = isset( $easydam_meta_data['videoConfig']['adTagURL'] ) ? $easydam_meta_data['videoConfig']['adTagURL'] : '';
 elseif ( ! empty( $ads_layers ) ) :
-	$ad_tag_url = rest_url( '/godam/v1/adTagURL/' ) . $attachment_id;
+	$site_id = null;
+	if ( is_multisite() ) {
```
Juzar10, line 105: you might not need this

```
		$site_id = null;
		if ( is_multisite() ) {
			$site_id = get_current_blog_id();
		}
```

you can directly do `get_rest_url( get_current_blog_is() )` and it should return the same result. as for single site the ID would be 1.

you can update it everywhere it's used.

admin/godam-retranscode-admin.php
```diff
@@ -39,6 +39,13 @@ class RetranscodeMedia {
 	 */
 	public $usage_info;
 
+	/**
+	 * Capability required to use this feature.
+	 * 
+	 * @var string
+	 */
+	public $capability;
```
Juzar10, line 47: "I noticed that the `$capability` variable was added, but it doesn’t seem to be used anywhere. Is it intended for a future change, or can it be removed?

pages/analytics/Analytics.js
```diff
@@ -21,10 +21,22 @@ const Analytics = ( { attachmentID } ) => {
 	const [ analyticsData, setAnalyticsData ] = useState( null );
 
 	const adminUrl = window.videoData?.adminUrl || '/wp-admin/admin.php?page=godam';
+	const restURL = window.godamRestRoute.url || '';
+	function pathJoin( parts, sep = '/' ) {
```
Juzar10, line 25: As this `pathJoin` function is used many time across the pages, we can make it a global function. in the admin.js file you can attach that function to window object.

`window.pathJoin = function( parts, sep = '/' )`

admin/godam-retranscode-admin.php
KMchaudhary: It's showing the Display warnings on wordpress admin, hence I added it as public variable and the previous transcoder plugin used it to assign transcoding related capabilities.

admin/godam-retranscode-admin.php
```diff
@@ -280,11 +287,16 @@ public function bulk_action_handler() {
 
 		$ids = implode( ',', $media );
 
+		$current_blog_id = null;
```
Juzar10, line 290: This still uses the `$current_blog_id` condition, can we maybe remove that.

Juzar10 (approved):

PR #213, opened 2025-03-14
Revamp Video Editor UI as per New Design [develop]

Juzar10:
## Description

This PR updates the video editor components to match the new design specifications.
design link: https://www.figma.com/design/qewE6AYuMQoSFnVPwjghKx/GoDAM?node-id=450-2153&t=PuW7IZFl43zWEm4B-4

- Add CSS classes as per the new design requirements
- Add those CSS classes to the component as needed
- Redesign video editor layout and it's changs
- Redesign the video seeker bar.

## Screenshots
Screenshot 2025-03-14 at 10 58 38 AM
Screenshot 2025-03-14 at 10 58 52 AM
Screenshot 2025-03-14 at 10 59 03 AM
Screenshot 2025-03-14 at 10 59 17 AM


inc/classes/rest-api/class-polls.php
```diff
@@ -0,0 +1,132 @@
...
+	public function get_rest_routes() {
+		return array(
+			array(
+				'namespace' => $this->namespace,
+				'route'     => '/' . $this->rest_base . '/polls',
+				'args'      => array(
+					array(
+						'methods'             => \WP_REST_Server::READABLE,
+						'callback'            => array( $this, 'get_polls' ),
+						'permission_callback' => '__return_true',
```
KMchaudhary, line 27: Should we add the permission_call to get a list of polls? This action should only be enabled for the roles like (author, editor, admin) 

Not sure about it!

inc/classes/rest-api/class-polls.php
```diff
@@ -0,0 +1,132 @@
...
+	 * @return \WP_REST_Response
+	 */
+	public function get_polls() {
+		global $wpdb;
+
+		if ( ! $this->is_poll_plugin_active() ) {
+			return new \WP_Error( 'poll_plugin_not_active', 'Poll plugin is not active.', array( 'status' => 404 ) );
+		}
+
+		$polls = $wpdb->get_results( "SELECT * FROM $wpdb->pollsq ORDER BY pollq_timestamp DESC" );
```
KMchaudhary, line 66: We can use the function provided by wp_polls plugin to fetch list of polls if they provide any.

KMchaudhary (approved):

inc/classes/rest-api/class-polls.php
Juzar10: I am also not sure. we need to check the permission callback values for other REST API created also.

We need to decide who can exactly make those changes, maybe we can also have custom capability for the plugin also.

inc/classes/rest-api/class-polls.php
Juzar10: I don't think there is any defined function for this one provided by the plugin. I tried to cache this but it might lead to some issues due to no proper cache invalidation, we can visit this later.

PR #216, opened 2025-03-19
Add error messages for plugin zip installation

prachigarg19:
When a user downloads plugin zip from repo, it gives a fatal error upon upload due to missing build files.
This PR deals with it by adding an error message to inform user about the need of generating plugin files instead of ambiguous fatal error message

When a user activates the plugin without generating the build files, below popup is displayed thus preventing user from activating the plugin :

image


Juzar10:
As this pull request is now outdated and no longer necessary, it might be best to close it.

@prachigarg19 , @KMchaudhary

PR #231, opened 2025-03-25
Redesign analytics [master]

prachigarg19:
This PR redesigns Analytics screen

### **Old:**
image

### **New:**
image

image

image


pages/analytics/Tooltip.js
```diff
@@ -1,7 +1,11 @@
+/**
+ * Internal dependencies
+ */
+import Info from '../../assets/src/images/info.png';
 const Tooltip = ( { text } ) => {
```
Juzar10, line 5: Add a single space after import statement.

pages/analytics/Analytics.js
```diff
@@ -72,78 +73,122 @@ const Analytics = ( { attachmentID } ) => {
...
+											<div className="analytics-single-info">
+												<div className="analytics-info-heading">
+													<p>
+														{ __( 'Average Engagement', 'godam' ) }
+													</p>
+													<Tooltip text="Video engagement rate is the percentage of video watched. Average Engagement = Total time played / (Total plays x Video length)" />
+												</div>
+												<p id="engagement-rate" className="min-w-[90px] engagement-rate">0%</p>
+											</div>
+											{/* <div className="analytics-stats">
```
Juzar10, line 101: There is some commented-out code in this PR. It is left temporarily for reference, but ideally, it should be removed or properly handled before merging.

Keeping unused code in the codebase can lead to long-term maintainability issues. I believe we should remove any code that is currently not needed, either because the backend feature is not complete or because it was temporarily removed. 

If we believe that this code might be useful, we should document this in the separate doc file and reference the relevant PR or commit.

Juzar10 (commented):
There are some eslint issues for the `video-analytics.js` and `Analytics.js` file. since you are already working on this, maybe we can also solve the eslint issue for these files.

Also for `Analytics.js` file..

can we move the `adminUrl` and `restURL` to outside the component.

prachigarg19:
@Juzar10 I've addressed your feedback, kindly check

Juzar10 (approved):
Thanks @prachigarg19 , LGTM 🚀

PR #235, opened 2025-03-27
Remove unused code

KMchaudhary:
- Remove the old `update_usage` function as usage data is now coming from a different endpoint and is handled differently.


Juzar10 (approved):
@KMchaudhary , I am assuming we are removing this because now the usage is calculate based on the API response correct?

PR #245, opened 2025-03-28
fix: Resolve audio transcoding issues in Media Library

subodhr258:


admin/class-rtgodam-transcoder-handler.php
```diff
@@ -628,13 +628,23 @@ public function add_transcoded_files( $file_post_array, $attachment_id, $job_for
...
+											array(
+												'ID' => $attachment_id,
+												'post_mime_type' => 'audio/mp3',
+											)
+										);
+									} else {
+										wp_update_post(
+											array(
+												'ID' => $attachment_id,
+												'post_mime_type' => 'video/mp4',
```
Juzar10, line 644: Why are we updating the mime_type in the first place? do you have any idea?

We are not changing the original video URL, just adding the transcoded `.mpd` URL. does any of the functionality break if we remove this whole logic to update the post?

Juzar10 (approved):

PR #258, opened 2025-04-11
Fix the conditional logic to load ad server and self-hosted ads

KMchaudhary:
**Issue:** Ad server and self-hosted ads are not loading properly because of the wrong conditional logic used to render ads.

**Solution:** Set the default adServer value and fix the condition to render ads.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ vendor/
 /pages/build/
 godam-build.zip
 godam-build
+godam.zip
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -38,4 +38,5 @@ vendor/
 /assets/build/
 /pages/build/
 godam-build.zip
-godam-build
\ No newline at end of file
+godam-build
```
Juzar10, line 41: We may not need this `godam.zip` now, as we are using the release base. for now we can remove this and then refactor the .gitignore file later.

KMchaudhary: Let's keep it for now, once we remove the `plugin-build.sh` then also remove from here.

Juzar10 (approved):

PR #264, opened 2025-04-14
feat: Send API key to fetch analytics endpoint for verification

subodhr258:


Juzar10 (commented):
Could you please clarify the purpose of including the `api_key` here, given that we're already sending the `account_token`? Just want to understand if both are necessary for authentication or validation.

Juzar10 (approved):

PR #322, opened 2025-05-05
Minor fixes to single analytics screen

prachigarg19:
1. Fix special characters in title
2. Prevent rendering duration if not available

Before:
image

After:
image

### DIFF
--- a/pages/analytics/Analytics.js
+++ b/pages/analytics/Analytics.js
@@ -16,6 +16,7 @@ import {
 	useFetchProcessedAnalyticsHistoryQuery,
 } from './redux/api/analyticsApi';
 import { calculateEngagementRate, calculatePlayRate } from './helper';
+import DOMPurify from 'isomorphic-dompurify';
 
 /**
  * WordPress dependencies
@@ -214,13 +215,6 @@ const Analytics = ( { attachmentID } ) => {
 		return 'left-greater right-greater';
 	};
 
-	function decodeHtmlCharCodes( text ) {
-		// Create a temporary DOM element
-		const textarea = document.createElement( 'textarea' );
-		textarea.innerHTML = text;
-		return textarea.value;
-	}
-
 	return (
 		<div className="godam-analytics-container">
 			<GodamHeader />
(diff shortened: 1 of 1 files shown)


pages/analytics/Analytics.js
```diff
@@ -242,11 +249,13 @@ const Analytics = ( { attachmentID } ) => {
...
 						<h2 className="text-2xl m-0 capitalize">
-							{ attachmentData?.title?.rendered }
+							{ decodeHtmlCharCodes( attachmentData?.title?.rendered ) }
 						</h2>
-						<span className="h-[26px] px-2 bg-white flex items-center rounded-sm">
-							{ attachmentData?.media_details?.length_formatted }
-						</span>
+						{ attachmentData?.media_details?.length_formatted &&
+							<span className="h-[26px] px-2 bg-white flex items-center rounded-sm">
+								{ attachmentData?.media_details?.length_formatted }
```
KMchaudhary, line 256: Are you sure this option will provide the video duration?

pages/analytics/Analytics.js
```diff
@@ -242,11 +249,13 @@ const Analytics = ( { attachmentID } ) => {
 				<div id="analytics-content" className="hidden">
 					<div className="p-10 flex gap-3 items-center">
 						<h2 className="text-2xl m-0 capitalize">
-							{ attachmentData?.title?.rendered }
+							{ decodeHtmlCharCodes( attachmentData?.title?.rendered ) }
```
KMchaudhary, line 252: We can either use the raw title if available (something like `title?.raw`)

Else you can use `<div dangerouslySetInnerHTML={{__html: attachmentData?.title?.rendered}} />` to render  title, also don't forgot to sanitise the title before rendering. We already have DOMPurify library in our project dependencies so you can use that.

pages/analytics/Analytics.js
```diff
@@ -214,6 +214,13 @@ const Analytics = ( { attachmentID } ) => {
 		return 'left-greater right-greater';
 	};
 
+	function decodeHtmlCharCodes( text ) {
```
KMchaudhary, line 217: I will not recommend to use this approach.

pages/analytics/Analytics.js
prachigarg19: Yes, it already does in the screenshot provided in above section of PR description

KMchaudhary (approved):

PR #360, opened 2025-05-21
Add option to Edit timestamp

prachigarg19:
Add an edit button to edit timestamp in video editing screen for layers.

image

image

image

Add checks for:
1. Prevention of adding similar timestamp for different layers
2. Adding value upto two decimal places
3. Adding value <= video duration
4. Allow only numbers and decimals
5. Prevention of empty value

### DIFF
--- a/pages/video-editor/VideoEditor.js
+++ b/pages/video-editor/VideoEditor.js
@@ -33,6 +33,7 @@ const VideoEditor = ( { attachmentID } ) => {
 	const [ currentTime, setCurrentTime ] = useState( 0 );
 	const [ showSaveMessage, setShowSaveMessage ] = useState( false );
 	const [ sources, setSources ] = useState( [] );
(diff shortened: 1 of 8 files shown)


subodhr258:
Found two minor issues:
1. We can trim the leading zeros in the number

Screenshot 2025-05-21 at 3 16 53 PM


2. When one tick is clicked, all the input fields are collapsed. We can either have only one field open at a time or collapse only one at a time

https://github.com/user-attachments/assets/314effb0-d14f-49a6-91c6-4db8b6affac9

KMchaudhary (commented):
Screenshot 2025-05-21 at 3 13 50 PM

Can't we add timestamp edit here? once user click on time it will show convert into input element and user can edit the layer timing there? 

@nayemDevs let us know your view here.

KMchaudhary (commented):
- Currently accepting negative values

PR #406, opened 2025-06-02
Feat: Woocommerce Layer

rudrakshi-gupta:
### Features added :

- Create new WooCommerce layer based on Hotspot layer.
- Create Products api to fetch products based on category, tag, brand, name id of product.
- Customize shop button with options like add to cart toogle button and button text.
- Render Product Hotspot with correct data in frontend.


assets/src/css/godam-player.scss
```diff
@@ -384,6 +384,162 @@
...
+.product-hotspot-woo-link:hover {
+	cursor: pointer;
+	pointer-events: auto;
+}
+
+.product-hotspot-box a {
+	overflow-wrap: break-word;
+	word-break: break-word;
+}
+
```
KMchaudhary, line 501: `.product-hotspot-box` is already declared, and since we're using SCSS, we should use SCSS nesting to style the layer elements. It would make the code more consistent.

Follow for other places as well.

inc/classes/rest-api/class-wc.php
```diff
@@ -0,0 +1,365 @@
...
+	/**
+	 * Get all WooCommerce products.
+	 *
+	 * @param \WP_REST_Request $request Request Object.
+	 * @return \WP_REST_Response
+	 */
+	public function get_products( $request ) {
+		$search     = sanitize_text_field( $request->get_param( 'search' ) );
+		$is_numeric = is_numeric( $search );
+		$term_info  = $this->godam_find_term( $search, array( 'product_cat', 'product_tag', 'product_brand' ) );
```
KMchaudhary, line 109: Performance improvement: We should not proceed with finding terms if the search term is numeric.

inc/classes/rest-api/class-wc.php
```diff
@@ -0,0 +1,365 @@
...
+	
+		$args = array(
+			'post_type'      => 'product',
+			'post_status'    => 'publish',
+			'posts_per_page' => 20,
+		);
+
```
KMchaudhary, line 116: **Improvements in serach strategy**
if 
   - $search is numeric and the check for post__in parameter
else
   - search all products with $seach name in it
   - search all product with matching $search taxonomies 
   - return all the products that matches search terms as well as taxonomies

pages/video-editor/components/woocommerce/FontAwesomeIconPicker.js
```diff
@@ -0,0 +1,161 @@
+/**
```
KMchaudhary, line 1: Why don't we import the Icon Picker component from `hotspot/FontAwesomeIconPicker.js` where ever required?

PR #408, opened 2025-06-02
fix: JS translation issue with translation plugins.

### LINKED ISSUE #407: Problem with string translations when using the translation plugin
## Description
While using translation plugins like [Loco Translate](https://wordpress.org/plugins/loco-translate/), the JS strings are not being translated.

## How
1. Install and activate [Loco Translate](https://wordpress.org/plugins/loco-translate/) plugin.
2. Switch the site’s language to `English (UK)` by navigating to `Settings > General > Language`.
3. Go to the `Translate Loco` and change the `General Settings` text to `General Settings UK` for the plugin.
4. Go to the `GoDam` menu page.

## Recording
https://github.com/user-attachments/assets/b9b9d7d5-40b4-44e6-b6f3-8e4162cc2cfc

mi5t4n:
## Description
This PR addresses an issue where JavaScript strings weren't being translated. The problem occurred because the JS files were minified `(*.min.js)`, and the `wp i18n make-pot` command always skips these files. As a result, the necessary strings were not included in the POT file. This fix ensures that unminified files are used for proper translation string extraction.

## How
1. Install and activate [Loco Translate](https://wordpress.org/plugins/loco-translate/) plugin.
2. Switch the site’s language to `English (UK)` by navigating to `Settings > General > Language`.
3. Go to the `Translate Loco` and change the `General Settings` text to `General Settings UK` for the plugin.
4. Go to the `GoDam` menu page.

## Issue Recording
[Kazam_screencast_00004.webm](https://github.com/user-attachments/assets/d6ade9da-f00e-4a53-a947-d0c9445ed362)

## Solution Recording
[Kazam_screencast_00005.webm](https://github.com/user-attachments/assets/aff6c20a-0a7c-4f58-a356-58ae0f146364)

Closes #407

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,12 +13,13 @@
 	"config": {
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true
-		}
+		},
+		"memory-limit": "1G"
 	},
 	"scripts": {
 		"pre-commit": [
-		"npm run lint:staged"
+			"npm run lint:staged"
 		],
-		"pot": "WP_CLI_PHP_ARGS='-d memory_limit=1024M' ./vendor/bin/wp i18n make-pot . --exclude=\"assets/src,assets/node_modules,tests,vendor\" ./languages/godam.pot"
+		"pot": "./vendor/bin/wp i18n make-pot . --include=\"admin,assets/build,inc\" --exclude=\".husky,admin/css,admin/images,assets/src,bin,wp-assets,pages\" ./languages/godam.pot"
 	}
 }
--- a/composer.lock
+++ b/composer.lock
@@ -534,16 +534,16 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v5.4.0",
+            "version": "v5.5.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/nikic/PHP-Parser.git",
-                "reference": "447a020a1f875a434d62f2a401f53b82a396e494"
+                "reference": "ae59794362fe85e051a58ad36b289443f57be7a9"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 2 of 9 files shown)


subodhr258:
Hi @mi5t4n,
We had updated all the build files to `.min.js` files so that they could be excluded from the POT file generation process. We had to exclude them because they caused a memory exhaustion error. 

If WordPress.org can't generate the POT file (with the default make-pot command, no exclusion parameters) when we release a new version, then it gives the warning that we had seen earlier, saying "The plugin is not prepared for localization".

The solution was provided by developers in the #meta-i18n channel [here](https://wordpress.slack.com/archives/C0341VCNL/p1747659100982119).

mi5t4n:
@subodhr258 

After reviewing the Slack thread, it seems that wp-org has some limitations with .pot file generation. However, we still need to ensure our plugin is fully compatible with other plugins, especially since non-English users primarily rely on translation plugins.

While inspecting the released version of the plugin, I noticed that many source JS files from the assets/src and pages directories are included. These files should be excluded from the final release build. Doing so should help resolve the memory exhaustion issue we’ve been seeing.

I will go through WooCommerce, and explore how they do things.

mi5t4n:
@subodhr258 Okay, it seems I found the answer. We can only include the build js and css files, as long as we link the source version to our public GitHub repository. 

Slack Thread: https://wordpress.slack.com/archives/C1LBM36LC/p1748927631053179

## WooCommerce
Note, WooCommerce also does the same thing in the plugin.



WooCommerce: https://wordpress.org/plugins/woocommerce/

## GoDAM
We also do the same thing for our plugin.


GoDAM: https://wordpress.org/plugins/godam/

> [!NOTE]
> So, we can remove the source js and scss files, and include only the build ones, doing so should speed up the pot generation in the wp-org.

Tagging @nayemDevs for the input.

hbhalodia:
Hi @mi5t4n, Is this PR needed? If yes can we resolve conflicts or if not then we can close this PR.

Issue #438, opened 2025-06-12
Implement theme support for GoDAM video player


KMchaudhary:
## Brand settings
![Image](https://github.com/user-attachments/assets/4f705dc0-ea45-4257-ab7c-de7fed6df900)

- [ ] Add a custom hook that will allow users to register new themes and enqueue styles for that theme.
- [ ] Implement the rtcamp theme for the rtcamp.com site.
- [ ] Prepare developer documentation on GoDAM themes, which will help the community to build a video player theme for their own site.

nayemDevs:
Another Skin suggestion - 

<img width="838" alt="Image" src="https://github.com/user-attachments/assets/5eb77175-c8ab-40ab-8fe2-37fddc5f0435" />

prachigarg19:
@kishan-gondaliya-7270 The feature is ready for testing

kishan-gondaliya-7270:
<img width="502" height="675" alt="Image" src="https://github.com/user-attachments/assets/ae910366-2f82-4fe2-a112-3c6658b23535" />

kishan-gondaliya-7270:
Nope @prachigarg19

kishan-gondaliya-7270:
Released in 1.3.0

PR #444, opened 2025-06-13
Generate skins for Video Player

prachigarg19:
This PR is still in WIP.

Resolves issue : https://github.com/rtCamp/godam/issues/438

This PR adds multiple skins, allowing user to switch between themes for their video player. 

It adds a tab under settings which allow user to switch between themes, set brand color and brand color globally for all the videos integrated using GoDAM block along with an option to preview the changes in the player


prachigarg19:
This PR is closed as it adds the rtCamp skin style replication which will not be provided to the end user publicly. This PR can be referred for the code of the same.

PR #463, opened 2025-06-19
fix: Video players controls style missing

theMasudRana:
### Description:

This PR fixes the issue where the GoDAM Video block controls were missing CSS variables when styles were applied via the editor. The problem was caused by inline styles overriding the expected variables.

#### Screenshot:

| Before | After |
|--------|--------|
| image | image |

Closes: https://github.com/rtCamp/godam-core/issues/174

### DIFF
--- a/inc/templates/godam-player.php
+++ b/inc/templates/godam-player.php
@@ -164,18 +164,42 @@ endif;
 
 $instance_id = 'video_' . bin2hex( random_bytes( 8 ) );
 
+// Create custom inline styles in a more maintainable way.
+$custom_css_properties = array(
+	'--rtgodam-control-bar-color'      => $easydam_control_bar_color,
+	'--rtgodam-control-hover-color'    => $easydam_hover_color,
+	'--rtgodam-control-hover-zoom'     => 1 + $easydam_hover_zoom,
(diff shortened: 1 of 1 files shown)


KMchaudhary (approved):

inc/templates/godam-player.php
```diff
@@ -167,18 +167,43 @@ function ( $layer ) {
...
+
+if ( ! empty( $attributes['aspectRatio'] ) ) {
+	$custom_css_properties['--rtgodam-video-aspect-ratio'] = $attributes['aspectRatio'];
+}
+
+// Build the inline style string, escaping each value.
+$custom_inline_styles = '';
+foreach ( $custom_css_properties as $propertie => $value ) {
+	if ( ! empty( $value ) ) {
+		$custom_inline_styles .= esc_attr( $propertie ) . ': ' . esc_attr( $value ) . ';';
```
KMchaudhary, line 187: No need to escape individuale style property and value

theMasudRana: Thanks @KMchaudhary, I have updated the code as recommended.

KMchaudhary (approved):

Issue #342, opened 2025-05-15
Extend Folder Organization Feature


nayemDevs:
We can add below features to our Folder and media management feature on WP Backend - 

- [x] Add sticky folder
- [x] Add star/bookmark the folder 
- [x] Lock Folder 
- [x] Download Zip of each folder 
- [x] Show media number on each folder 

Quality-of-life suggestions

- [ ] Allow customization of folder icon/color
- [ ] Dynamic folders based on date and year of when media was uploaded.

Ref: https://www.youtube.com/watch?v=1SqDey4idlQ 

<img width="1017" alt="Image" src="https://github.com/user-attachments/assets/6836ff90-106c-48be-9f20-5c8b18d845e4" />

Ref: https://codecanyon.net/item/media-folders-manager-for-wordpress/21715379

Juzar10:
Regarding the "**Star vs Bookmark**" functionality for folders, since we're currently not implementing any folder sorting or prioritization logic, the purpose of a separate **Star feature** isn’t entirely clear. Both actions (**Star and Bookmark**) seem functionally similar at this stage, and without sorting or filtering based on stars, it may introduce unnecessary confusion for users.

**Recommendation**:

Let’s proceed with implementing only the **Bookmark** feature for now. We can revisit and analyze the value of a distinct **Star** functionality once we have sorting or prioritization use-cases in place.

Let me know your thoughts.

@prachigarg19 , @nayemDevs

nayemDevs:
Thanks, @Akash097531 We need to re-design the UI

nayemDevs:
@Juzar10 Star/Bookmark same feature just it's the icon some plugin using to bookmark.

Juzar10:
To ensure optimal performance, files larger than `500MB` are currently skipped during ZIP generation. However, I will provide a filter hook to allow developers to customize this size limit as needed.

We have to test various scenarios to make sure the implementation remains performant and is working as expected. I will provide few basic test cases we can check for.

nayemDevs:
Design File - https://www.figma.com/design/qewE6AYuMQoSFnVPwjghKx/GoDAM?node-id=3409-16166

nayemDevs:
@DarkMatter-999 - You can take up the re-design task - thanks

subodhr258:
One more doubt in the design @nayemDevs
Here the number beside **_Bookmarked_** shows the total number of items in all the bookmarked folders right? (and not the number of bookmarked folders)
[...]

PR #482, opened 2025-06-23
Extend Media Library features

Juzar10:
Issue: https://github.com/rtCamp/godam/issues/342

### Show media number on each folder

We are now showing the attachment count for each folder.

#### Screenshot

https://github.com/user-attachments/assets/87de093e-17e4-41b6-bde8-e306723d82d2

### Lock Folder

Allow the ability to lock the folder, so that no new attachment can be added to it, also added a context menu for the same.

#### Screenshot

https://github.com/user-attachments/assets/e416ca7c-fee7-46df-bc69-ac026665dfa8

### Add Bookmark

Allow the functionality to add bookmarks of the media folders.

#### Screenshot

https://github.com/user-attachments/assets/89bc3673-e573-4941-8b92-002578ec48cb

### Download as Zip

Allow media to be download as a zip.

**NOTE: We need to test this functionality and it's limitation thoroughly, I will optimize this functionality a little in following PR.**

#### Screenshot

https://github.com/user-attachments/assets/899d918d-adf3-4d6f-8d48-014b88c72d6f


inc/classes/media-library/class-media-folder-utils.php
```diff
@@ -0,0 +1,185 @@
...
+	 */
+	public function get_attachment_count( $folder_id, $force_refresh = false ) {
+		$folder_id = absint( $folder_id );
+
+		if ( $folder_id <= 0 ) {
+			return 0;
+		}
+
+		// Create a unique cache key for this folder.
+		$cache_key = 'attachment_count_' . $folder_id;
```
KMchaudhary, line 46: We should use the `rtgodam_` prefix.

KMchaudhary (approved):
Need to add `@since n.e.x.t` on block comments. We started following this practice to well documented the codebase, and help developers on which version specific function was added/modified.

We will replace the `@since n.e.x.t` with the upcoming version during new plugin version release.

@nayemDevs I will take care of this point for PR, as @Juzar10 is on week leave

nayemDevs:
@KMchaudhary Correct, when that option will be enabled, we may disable the zip creation option from backend.

Issue #462, opened 2025-06-19
Video already transcoded showing "not started" in GoDAM Media tab


hashibhere:
https://github.com/user-attachments/assets/5c481a43-3ba8-4d63-a934-18b7aadc1619

dishitpala:
@subodhr258 Can you help me by checking the status of the below job IDs?

<img width="535" alt="Image" src="https://github.com/user-attachments/assets/9560250c-bd51-4bc7-a9d2-ba936fcfbce4" />

Also, please, can you provide me access to where I can check the status of transcoding media?

hashibhere:
Yes @subodhr258 has mentioned that correctly.

PR #495, opened 2025-06-23
Fix Video already transcoded showing 'not started' in GoDAM Media tab

dishitpala:
Fix Video already transcoded showing 'not started' in GoDAM Media tab
Issue: https://github.com/rtCamp/godam/issues/462

Screenshot 2025-06-23 at 6 52 48 PM

### DIFF
--- a/inc/classes/class-media-library-ajax.php
+++ b/inc/classes/class-media-library-ajax.php
@@ -126,20 +126,11 @@ class Media_Library_Ajax {
 
 			$response = $body->message;
 
-			$output = array();
-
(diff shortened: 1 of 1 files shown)


KMchaudhary:
@elifvish can we add one parameter in media fetching API endpoint, that will only return the sucessfully transcoded medias in response.

ref: 
https://github.com/rtCamp/godam/blob/41c880eecd5402c5e6e3643c84f0aa5414f2c942/inc/classes/class-media-library-ajax.php#L77

PR #513, opened 2025-06-24
GH 510: Fix godam video settings, layer to restrict images on upload files selection

### LINKED ISSUE #510: Restrict “Upload Files” Tab to Images Only for Logo and CTA Image Fields
## 🐞 Issue

* In the **General Settings** of the plugin, there's an option to upload a custom logo. This is intended to accept only image files from the media library.
  However, when using the **Upload Files** tab, users are able to upload any file type that is permitted by the WordPress installation — not just images.

* Similarly, when adding a **video layer (CTA)** and selecting the image option, the **Media Library** correctly filters to show only image files.
  But again, the **Upload Files** tab accepts all file types allowed by the site.

## ✅ Expected Behavior

Only image files should be allowed for upload in these cases.

## 💡 Possible Solutions

* Restrict the file uploader to accept only image file types (e.g., JPG, PNG, SVG, WebP, etc.) in both the logo upload and CTA image layer fields.
* Alternatively, if restricting at the upload level is not feasible, show an error message when a non-image file is uploaded (e.g., "This file type is not allowed for this field") and prevent it from being used.

## 📷 Screencast

https://github.com/user-attachments/assets/2ffdf394-bfe1-4f22-bb72-6820ca6c6067

---

hbhalodia:
## 🐞 Issue

- Closes #510

---

## ✅ Solution

- After a media item is selected, check the attachment type.
- If the selected file is **not an image**, exit early and do not set the media.

---

### Before

https://github.com/user-attachments/assets/2ffdf394-bfe1-4f22-bb72-6820ca6c6067

### After

https://github.com/user-attachments/assets/d606d2cc-07b2-49dd-8261-083fa75bb094

### DIFF
--- a/pages/video-editor/components/cta/ImageCTA.js
+++ b/pages/video-editor/components/cta/ImageCTA.js
@@ -4,8 +4,8 @@
 /**
  * WordPress dependencies
  */
-import { Button, Icon, RangeControl, SelectControl, TextareaControl, TextControl } from '@wordpress/components';
-import React, { useEffect, useState } from 'react';
+import { Button, Icon, Notice, RangeControl, SelectControl, TextareaControl, TextControl } from '@wordpress/components';
+import React, { useEffect, useState, useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { __ } from '@wordpress/i18n';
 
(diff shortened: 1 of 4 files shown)


pages/video-editor/components/cta/ImageCTA.js
```diff
@@ -66,13 +89,14 @@ const ImageCTA = ( { layerID } ) => {
...
+				setSelectedImageUrl( media.source_url );
 			} );
-	};
+	},
+	[ restURL, setSelectedImageUrl ] );
 
 	useEffect( () => {
 		fetchOverlayMediaURL( layer.image );
-	}, [ layer ] );
+	}, [ layer, fetchOverlayMediaURL ] );
```
hbhalodia, line 99: Have passed the dependency that was missing to useEffect.

KMchaudhary (approved):

Issue #623, opened 2025-07-08
Missing Translation-Ready Strings


subodhr258:
[assets/src/js/deactivation-feedback.js](https://github.com/rtCamp/godam/blob/main/assets/src/js/deactivation-feedback.js)

Several UI strings in this file are not wrapped with translation functions, making them unavailable for localization. For example:
```jsx
 <a href="${ deactivateLink[ 0 ].href }">
    Skip & Deactivate
</a>
```

## Tasks
	1. Wrap all user-facing strings in translation functions using the 'godam' text domain.
	2. Verify translation readiness using the Loco Translate plugin.

PR #624, opened 2025-07-08
GH-623 Adds translations for untranslated strings.

vipul0425:
Closes: https://github.com/rtCamp/godam/issues/623

### DIFF
--- a/inc/classes/rest-api/class-media-library.php
+++ b/inc/classes/rest-api/class-media-library.php
@@ -253,14 +253,14 @@ class Media_Library extends Base {
 				$return = $this->remove_all_terms_from_id( $attachment_id, 'media-folder' );
 
 				if ( is_wp_error( $return ) ) {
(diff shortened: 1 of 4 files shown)


hbhalodia (approved):
LGTM!

subodhr258:
One thing I noted: 
`import { __ } from '@wordpress/i18n';` this is added in the JS files but `wp-i18n` is not added as dependency while enqueueing the files in PHP.
For example:
```
wp_register_script(
			'godam-deactivation-survey-script',
			RTGODAM_URL . 'assets/build/js/deactivation-feedback.min.js',
			array(),
			filemtime( RTGODAM_PATH . 'assets/build/js/deactivation-feedback.min.js' ),
			true
		);
```
If the PHP doesn't enqueue wp-i18n as a dependency, then `wp.i18n` will be undefined when the script runs

@vipul0425 can you also please verify with the Loco Translate plugin if the translations are working?

Issue #287, opened 2025-04-23
Improve Folder Organization  feature in Media Library


KMchaudhary:
###  1. Selecting a folder and expanding or collapsing its subfolders are interfering with each other.
- We can fix this by triggering an expansion or collapse of the subfolder only when the user clicks on `⌄` icon.
https://github.com/user-attachments/assets/859c60e4-dc92-4a4b-b0b0-aa7734e7a955

### 2. Drag and drop of media files should also work with the media library list view.
<img width="1310" alt="Image" src="https://github.com/user-attachments/assets/ba079b71-fd8b-4ef4-bbc2-9eed1a82e057" />
Note: Drag and drop of media from grid view is working as expected.

### 3. Display a proper error message if add, rename, delete any of the folder operations failed

Juzar10:
For the issue: **Display a proper error message if add, rename, delete any of the folder operations failed**

We've already implemented proper error handling for most folder operations like add, rename, and delete — these return and display appropriate error messages when they fail. 

If there are specific cases where the error message is unclear or not descriptive enough, I'd be happy to address those individually. Otherwise, if there's no concrete example of a missing or misleading error, we can consider closing this issue.

@KMchaudhary, @nayemDevs

kishan-gondaliya-7270:
- Folder created toast has abnormal height 
<img width="1301" height="793" alt="Image" src="https://github.com/user-attachments/assets/b4ba52c4-4fd7-4af0-b943-df0cb1092c97" />


- ux When there is more folders : 
<img width="1469" height="878" alt="Image" src="https://github.com/user-attachments/assets/8dd28e08-985c-4acd-b009-c5f449e73966" />

- ux when there is 10ish folder :  (Looking bad as if folder tab is conflicting with lower part of site)
<img width="1394" height="824" alt="Image" src="https://github.com/user-attachments/assets/033bf04c-f90b-491c-a386-9b0f83f6055e" />

- Folder is displayed only when we have grid view, when we click list view folder tab is not appearing 
https://github.com/user-attachments/assets/c9be1a34-6872-4557-9762-0f0b3f41fb9d

@nayemDevs @Juzar10

kishan-gondaliya-7270:
https://github.com/rtCamp/godam/pull/644 @CoderAbhinav please get this PR merged to resolve #287

subodhr258:
@kishan-gondaliya-7270 both the PRs are now merged

kishan-gondaliya-7270:
[...]

PR #643, opened 2025-07-10
Fix: Snackbar toast height

DarkMatter-999:
part of #287 

Fix Snackbar toast height bug

### Before:
image

### After
image

### DIFF
--- a/pages/media-library/index.scss
+++ b/pages/media-library/index.scss
@@ -12,28 +12,28 @@
 	flex-direction: column;
 	min-height: calc(100vh - 34px);
 	overflow: auto;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


pages/media-library/index.scss
```diff
@@ -23,8 +23,7 @@
 		border-bottom: 1px solid #E0E0E0;
 	}
 
-	> :nth-last-child(3) {
-        border: none;
+	.tree-container {
         padding-bottom: 2.5rem;
```
CoderAbhinav, line 27: Use `tabs` instead of spaces, please check the current file and make the change.

CoderAbhinav (approved):

Issue #550, opened 2025-06-26
Suggested Fixes and Enhancements by QA


subodhr258:
## Description
These are the suggestions by QA team during 1.1.4 release testing.
~~1. Enhancement : Transcoded CDN URL box can be hidden for free version or add indicator to show case as premium feature field~~
For free users, transcoding is not provided, and so the CDN URL will always be empty here. We can add a Dollar sign ($) or any other indicator to show that this is a paid feature.

![Image](https://github.com/user-attachments/assets/e8a747e5-3717-4683-b8a3-6d365a8c6658)

### 2. Enhancement : Entries added by GoDAM video embedded forms can have identifier field/metadata to differentiate form responses from different source.
This applied to all kinds of form integrations:
1. WP Forms
2. Contact Forms
3. Gravity Forms
4. Jetpack Forms

Currently there is no way to distinguish if the form was submitted from a separate page or GoDAM Video layer form.
We can follow a hidden fields approach for this.
For example:
```php
// For Gravity Forms
add_filter('gform_field_content', function($content, $field, $value, $entry_id, $form_id) {
    if ($field->type === 'hidden' && $field->inputName === 'godam_source') {
        $content = '<input type="hidden" name="godam_source" value="' . esc_attr(json_encode($godam_identifier)) . '">';
    }
    return $content;
}, 10, 5);

// For Contact Form 7
add_filter('wpcf7_form_elements', function($content) {
    $content .= '<input type="hidden" name="godam_source" value="' . esc_attr(json_encode($godam_identifier)) . '">';
    return $content;
});
```

### 3. Limitation : Animation of block won’t work at parent level - though it is working at field level (Jetpack Forms)
Block Editor Setting:

![Image](https://github.com/user-attachments/assets/2c9cd93a-efbf-465a-8c5f-969437cb2817)

Video Editor and Frontend Output (Inconsistent):

![Image](https://github.com/user-attachments/assets/a7a0e304-e97a-41fc-a1f5-25e6538f5ac0)

### 4. Issue : Able to save layer without having any form for gravity form after adding gravity form or any form. We need to add proper validation when saving forms. 
This needs to be validated when saving for all supported Forms (Gravity Forms, WP Forms, Contact Form 7, Jetpack Forms)

https://github.com/user-attachments/assets/4158a26b-8e47-4c04-8269-46979e6e4f5d

nayemDevs:
@Mukulsingh27 Second option looks good

KMchaudhary:
[...]

PR #656, opened 2025-07-14
Fix: Add form layers validation

nani-samireddy:
Part of: #550 

As we have a single `save` button for all the layers, Added a validation on all layers of type `form`.

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/d6f3e781-98a2-4250-871f-bd1b686bbf79"> </video> | <video src="https://github.com/user-attachments/assets/2153d0b3-d88f-4640-8453-bf6806256a3a"> </video> |

NOTE: For a better user experience, we can update the toast message to show which layer(s) are failing validation with their timestamp.

**IMPORTANT: As we are using different `field_id` for different form integrations. We need to update the `formIDMap` whenever we add a new form integration.**

### DIFF
--- a/pages/video-editor/VideoEditor.js
+++ b/pages/video-editor/VideoEditor.js
@@ -200,7 +200,7 @@ const VideoEditor = ( { attachmentID } ) => {
 		// Validate form layers before saving.
 		if ( invalidLayers.length > 0 ) {
 			const layerTimes = invalidLayers.join( ', ' );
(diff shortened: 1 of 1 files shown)


nani-samireddy:
@subodhr258 I have updated the toast message with to show the layer timestamps and It looks like following:

https://github.com/user-attachments/assets/650ad84a-9374-4849-9d38-a310d4794fbe

subodhr258:
@nani-samireddy it is a good idea to show the timestamp of the layers, but this is the current message:

> Please select a form for the following layers: 0.72, 1.59

We could show that they are timestamps with a message like this:

Please select a form for the layers at timestamps: 0.72s, 1.59s

PR #630, opened 2025-07-09
feat: Add GoDAM Video CPT

yashjawale:
Closes #616 

## What
- This PR adds support for `godam_video` CPT which will be used by Themes for creating templates
- Enables query loop for GoDAM videos
- Adds fallback content so that `index` template can display video even without explicit support by theme
- Additional setting for video slug is added to Plugin Settings -> Video Settings

## How
- Add new `base` cpt class [similar to taxonomies base class]
- Adds new `godam_video` CPT -- Hidden
- `godam_video` post is created when a video attachment is added/modified
- CPT content is updated as changes are made to attachment data [CPT slug is updated according to attachment title]
- Setting for video post slug provided in GoDAM settings under GoDAM -> Settings -> Video Settings
- Invalid value for slug is sanitized & a preview is shown under the input
- Content mapping
  - Attachment title -> CPT title
  - Attachment description -> CPT excerpt
  - Attachment video source -> GoDAM Video Player block markup added in CPT content as fallback
  - Attachment taxonomies -> CPT taxonomies

## Screencast

https://github.com/user-attachments/assets/adc350c2-08cb-41ce-a7b9-7869f4aa50d2

Video slug settings

image

PR #696, opened 2025-07-18
Add user settings for CPT visibility in links

yashjawale:
Follows up on #630 

## What
This PR adds user settings for controlling CPT pages visibility
New settings are added under Permalinks for controlling whether archive/single pages for videos are visible to public

## Why
After introduction of CPTs, the archive & single pages were visible by default
While this may work for websites like YouTube / Netflix, this behavior might not be ideal for other websites which use GoDAM just for hosting & serving their video content in their post

## How
- Two new options: `rtgodam_video_post_allow_archive` & `rtgodam_video_post_allow_single` are added to store user preferences
- Based on these settings, `exclude_from_search` & `has_archive` options are controlled during CPT registration
- `public` & `publicly_queryable` are always set to `true` to allow compatibility with Query Loop block regardless of user preferences
- Instead, to control visibility, `template_redirect` hook is utilized to check the value of options & redirect to 404 template accordingly

## Screenshot

image

### DIFF
--- a/inc/classes/class-video-permalinks.php
+++ b/inc/classes/class-video-permalinks.php
@@ -148,6 +148,8 @@ class Video_Permalinks {
 	/**
 	 * Video archive visibility field.
 	 *
+	 * @since n.e.x.t
+	 * 
 	 * @return void
 	 */
 	public function video_archive_visibility_field() {
@@ -175,6 +177,8 @@ class Video_Permalinks {
 	
 	/**
 	 * Video single page visibility field.
+	 * 
+	 * @since n.e.x.t
 	 *
 	 * @return void
 	 */
@@ -267,6 +271,8 @@ class Video_Permalinks {
 
 	/**
 	 * Sanitize video post settings.
+	 * 
+	 * @since n.e.x.t
 	 *
 	 * @param array $input The input array to sanitize.
 	 * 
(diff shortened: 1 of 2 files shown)


yashjawale:
Thanks for the review 🙌
I've addressed the changes, requesting re-review...

inc/classes/class-video-permalinks.php
```diff
@@ -108,7 +140,63 @@ public function video_slug_field() {
...
 		<p class='description'>
-			<?php echo esc_html__( 'This slug will be used in the URL for video archive and single video pages (e.g., yoursite.com/videos/). Only lowercase letters, numbers, hyphens, and underscores are allowed.', 'godam' ); ?>
+			<?php esc_html_e( 'This slug will be used in the URL for video archive and single video pages (e.g., yoursite.com/videos/). Only lowercase letters, numbers, hyphens, and underscores are allowed.', 'godam' ); ?>
+		</p>
+		<?php
+	}
+	
+	/**
+	 * Video archive visibility field.
+	 *
```
mi5t4n, line 150: We need to start adding `@since n.e.x.t` to newly introduced functions, files, and classes.

yashjawale: Added ✅

PR #708, opened 2025-07-21
FEAT: Restrict access to video editing features for non-owners in media library and video editor

patil-vipul:
## Summary
Added role-based restrictions to hide video analytics, edit video button, and thumbnail strip in the media popup for users without access to others' media.

On the video editor page:
- Blocked direct access to other users' media via URL.
- Redirects unauthorized access to the media grid view.
- Prevents loading videos the user does not own.

### Screencaptures
The first selected media item shown below is not owned by the user; hence, no video editor features are displayed (similar to how WordPress restricts access to media not owned by the user).

https://github.com/user-attachments/assets/c75a34f2-cf4e-4ca7-98c7-27a4adff857a

As you can see, in the video editor, only media owned by the user is accessible — others are grayed out. Even if the user tries to access the editor via a URL containing a media ID they do not own, they are redirected back to the media selection window.

https://github.com/user-attachments/assets/425f0d97-3a1d-4523-bb9b-82d481d4865f


pages/video-editor/VideoEditor.js
```diff
@@ -176,6 +180,14 @@ const VideoEditor = ( { attachmentID } ) => {
 		}
 	}, [ gravityForms, cf7Forms, wpForms, everestForms, isFetching, dispatch, sureforms, forminatorForms, fluentForms, ninjaForms ] );
 
+	const canManageAttachment = ( attachment ) => {
+		const currentUserId = Number( window?.easydamMediaLibrary?.userId );
+		const canEditOthersMedia = window?.easydamMediaLibrary?.canEditOthersMedia;
```
hbhalodia, line 185: I see this function is being called multilpe times, Can we add is somewhere in utils? and use it at all places?

hbhalodia:
Hi @patil-vipul, Can you please resolve the conflicts and check on the feedbacks?

Thanks.

pages/video-editor/VideoEditor.js
patil-vipul: Created a shared utility function.

Issue #730, opened 2025-07-23
Media Library UI and Functionality Issues


subodhr258:
There are various issues noted after the Media Library UI redesign:

### 1. Hide Folders button should not be visible in media selector if "Enable folder organization in media library" is disabled.

https://github.com/user-attachments/assets/2e7c4e52-f30b-4db0-98d2-636cce691223

### 2. "Not Transcoded" Status is inconsistent between Media Library and the Media Selector.

https://github.com/user-attachments/assets/2a7f5d8e-0101-4455-9e74-6ee7131cad94

### 4. All Transcoding Status text should be properly capitalized. The video for the above task shows status as "transcoding" rather than "Transcoding".

subodhr258:
1,2, and 4 are completed and QA is required @kishan-gondaliya-7270

kishan-gondaliya-7270:
Move 3 from 1.3.1 after checking feasibility

kishan-gondaliya-7270:
Adding featured image in WordPress will show in list view as well 

<img width="1282" height="408" alt="Image" src="https://github.com/user-attachments/assets/362e0803-dd88-41b6-b4cf-ad013ec3a65d" />

coderGtm:
QA Tested! LGTM!
@kishan-gondaliya-7270

kishan-gondaliya-7270:
It is resolved and released in 1.3.1

PR #744, opened 2025-07-24
Fix: Media library UI functionality

DarkMatter-999:
part of #730 

- [x] Fixed hide folders button, now respects setting.
- [x] FIxed transcode status inconsistentencies between Media Library and the Media Selector 
- [x] Fixed casing on Transcoding status text  

image

image

image

### DIFF
--- a/assets/src/js/media-library/utility.js
+++ b/assets/src/js/media-library/utility.js
@@ -103,21 +103,13 @@ async function getGodamSettings() {
 	const url = `${ baseUrl }godam-settings`;
 
 	try {
-		const response = await fetch( url, {
+		const response = await wp.apiFetch( {
+			path: url,
 			method: 'GET',
-			headers: {
-				'Content-Type': 'application/json',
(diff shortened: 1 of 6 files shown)


assets/src/js/media-library/utility.js
```diff
@@ -91,4 +91,25 @@ const getQuery = ( props ) => {
 	} );
 };
 
-export { isAPIKeyValid, checkMediaLibraryView, isUploadPage, isFolderOrgDisabled, addManageMediaButton, getQuery };
+/**
+ * Helper function to fetch media settings.
+ *
+ * @return {Promise<Object>} A promise that resolves with the JSON response from the API.
+ * @throws {Error} If the network request fails or the response status is not OK.
```
mi5t4n, line 98: Everything looks good—just one final comment. This function doesn't throw an error because it's wrapped in a try-catch block. Let's remove the try-catch so the behavior aligns with the function signature.

mi5t4n (changes requested):

PR #765, opened 2025-07-28
VideoSEO fix where the SEO data does not appear

Juzar10:
## ✨ Overview

This PR fixes the VideoSEO modal where default values were not saved for the VideoSEO.


hbhalodia (changes requested):
- Please review copilot changes and address them, looks good to accomodate.

hbhalodia (approved):

Issue #318, opened 2025-05-02
Migration of Existing Media/Video to DAM


nayemDevs:
We need to hanlde migration of existing media to GoDAM and replace the media/URL inside post content.

coderGtm:
<img width="1225" height="334" alt="Image" src="https://github.com/user-attachments/assets/de55e121-6ab5-4bbe-b92f-b71238f0f56a" />

Added note to clear the purpose of the migration.

cc: @subodhr258 @nayemDevs

kishan-gondaliya-7270:
This is still in progress, can you confirm it @subodhr258 ?

coderGtm:
@Juzar10 @KMchaudhary It is fixed by #905 but it got merged yesterday only. I tested it after migration and could only see 1 source:

<img width="427" height="222" alt="Image" src="https://github.com/user-attachments/assets/663ea6f7-2526-4eec-ba50-dd4cb27d8a7e" />

subodhr258:
Possible Future Enhancements:

- Add a toggle button to collapse the Vimeo Migration container on the page
- Disable the button when migration has not been done on GoDAM Central
- When a Vimeo block is added to post and saved, then we can enable the migration button (detecting on save_post hook)

cc: @KMchaudhary @coderGtm

KMchaudhary:
@coderGtm any update on this?

kishan-gondaliya-7270:
@coderGtm correct status for changes that are in or require review should be "Code Review"

KMchaudhary:
@coderGtm Did you work on other changes?
- Change button text to "Restart migration", migration is already done.
- When clicking on the "Start migration" button, it takes 1-2s to start migration, and the button status does not change during that time, which is confusing for users.

coderGtm:
@KMchaudhary I am working on it.

kishan-gondaliya-7270:
[B] Title text for vimeo migration on rtCamp site (sol - it is only required for 1 about us video page, so can be done manually)
 @KMchaudhary

KMchaudhary:
@coderGtm 
We discuss whether we should hide the Vimeo migration tool by tracking if any post contains Vimeo videos.
- Concludes that tracking may include unnecessary data processing and may lead to decreased site performance in the long term, so we should not include Vimeo video tracking.

**Solution that we will follow:** In the upcoming version, we may split the migration process into two parts. 1) Fetch Vimeo videos all across the site posts. 2) Perform migration after the video list is fetched.

[...]

PR #778, opened 2025-07-29
fix: use correct sources array during migration

coderGtm:
Core and Vimeo migration was using empty array for sources, relying on the editor to figure it out during runtime. This caused VideoJS errors causing the godam video block to not load sometimes. 

This PR fixes it by setting the sources during migration itself to prevent these errors.


subodhr258 (approved):

subodhr258:
Part of #318

Issue #747, opened 2025-07-24
[EPIC] Virtual Media ID Issues


subodhr258:
- [ ] Dependency: https://github.com/rtCamp/godam-analytics/pull/49

---

Virtual Media are created in cases when the media is already existing on Central DAM. In such cases the video ID is not created. Instead we use the Job ID.

But this causes the `data-id` attribute in the GoDAM video to be null.

## Null `data-id`
```php
data-id="<?php echo esc_attr( $attachment_id ); ?>"
```

This was caught during testing of an analytics request. Null video IDs were being sent, leading to 422 error response. This is causing issues in production since Analytics data is not being sent if the page contains any virtual media.

This was the `video_ids` array in the payload: `video_ids :[540, 9, 256, null]`
The response was:
```json
{
    "detail": [
        {
            "type": "int_type",
            "loc": [
                "body",
                "video_ids",
                3
            ],
            "msg": "Input should be a valid integer",
            "input": null
        }
    ]
}
```
A temporary solution for now can be to allow null values in an array on Analytics service.

## Virtual Media not being visible in Video Editor
The **Customize** button leads to just the video editor page and not the specific video because it does not exist on the Video Editor page. The id used here is the Job ID which is incorrect.

https://github.com/user-attachments/assets/48a9ff81-b355-49fc-a509-fb28181e1ee8

mi5t4n:
@subodhr258 Could you explain in what scenario and steps a virtual media ID will be created?

kishan-gondaliya-7270:
- [ ] Fix preview Template and Preview Video is not working for Virtual Media 

<img width="1528" height="745" alt="Image" src="https://github.com/user-attachments/assets/22796f15-8785-4d28-ab66-b8b8a801a7f5" />

geekofshire:
https://github.com/rtCamp/godam-analytics/pull/49 - Related Analytics PR

subodhr258:
Analytics of virtual media is being tracked [here](https://github.com/rtCamp/godam/issues/974), closing this ticket since all items are completed.

PR #780, opened 2025-07-29
Fix - Null Video IDs being sent for Analytics

subodhr258:
Part of #747 
## Overview
Before:
```
video_ids: [null, 199]
```
This led to 422 error:
```
Input should be a valid integer
```
After:
```
video_ids: [199]
```


geekofshire (approved):
LGTM

Issue #804, opened 2025-08-01
GoDAM player block video playback issues


hbhalodia:
## Issue 1

- We have player settings for each video defined, and we have settings to define the skip duration.
- On adding the skip duration as `10s`, frontend correctly render the skip duration and on click of that it correctly skips to that duration.
- Now, when user is using keyboard for navigation, it does not respects the skip duration, it always skips to 5sec, no matter what settings stored is.

## Possible solution
- Skip duration should be consistent and should be based on settings configured and it should not be different on mouse click and keyboard arrow skip.

## Screencast

https://github.com/user-attachments/assets/f74ac471-ee55-48be-b91b-c68b7271affb

---

## Issue 2 -- Add as needed.

Cc: @subodhr258 @Juzar10 @Mukulsingh27

Mukulsingh27:
I believe the default skip is 5 seconds when using the keyboard, similar to how YouTube handles video skipping. However, this PR will make it consistent with the value shown in the skip settings controls, right?

I think consistency here is a good improvement—it avoids confusion and matches the UI settings.

hbhalodia:
- QA required.

Cc: @subodhr258 @Mukulsingh27 @kishan-gondaliya-7270

kishan-gondaliya-7270:
@renishsurani verify it on Safari and IoS

subodhr258:
@renishsurani since @KMchaudhary has fixed layer display on iPhones, please test again with this [zip](https://github.com/rtCamp/godam/actions/runs/17325208359)
cc: @kishan-gondaliya-7270

renishsurani:
I have verified this issue, and it appears to be resolved as all the custom player controls are now visible.

However, there are still some issues with the skin, which will be reported separately in a different [ticket](https://github.com/rtCamp/godam/issues/1006) and are not part of this issue.

cc: @kishan-gondaliya-7270 @subodhr258

kishan-gondaliya-7270:
Released with 1.4.0

PR #811, opened 2025-08-04
Fix: Update the skip duration using keyboard arrow to respect selected settings.

hbhalodia:
- Issue - https://github.com/rtCamp/godam/issues/804
- Fix, use `videoControls` data to fetch the skip duration settings and use that for skip using keyboard buttons.


hbhalodia:
> LGTM Can you add a demo video for ref?

Sure 👍

mi5t4n (approved):
LGTM

Issue #812, opened 2025-08-04
GoDAM Player settings `Show Volume Slider` setting is not respected.


Juzar10:
## Description

The `Show Volume Slider` in the Player settings is not respected.

## Recording

https://github.com/user-attachments/assets/1b866ee8-ebbc-4aba-86ce-ba10acbd557d

Juzar10:
@kishan-gondaliya-7270, QA needed for this.

kishan-gondaliya-7270:
Fixed and Tested 

<img width="1780" height="846" alt="Image" src="https://github.com/user-attachments/assets/52df60d7-6060-4ff2-aef2-39742855cb55" />
<img width="1558" height="958" alt="Image" src="https://github.com/user-attachments/assets/f3c2e733-fdb6-4916-acee-638910688aa7" />

PR #782, opened 2025-07-30
Migrate to Container Queries, Enhance Player Compatibility, and Fix Branding Display

prachigarg19:
Resolves issue: #753 

This PR introduces structural improvements and fixes to the GoDAM video player and its styling behavior, ensuring it aligns with modern CSS practices and WordPress theme compatibility:

###  **Key Changes:**

1. **Replaced All Media Queries with Container Queries**

   * Migrated all concerned responsive logic to use CSS `@container` queries instead of `@media`, improving modularity and scoped responsiveness.

2. **Ensured Compatibility with WordPress Default Themes (2020–2025)**

   * Validated container-based styling and layout behavior across the default WordPress themes (`Twenty Twenty` through `Twenty Twenty-Five`).
   * Ensured the player renders responsively and maintains functional UI elements (control bar, overlays, skip buttons) in all themes.

3. ** Fixed Branding Logo Display Bug**

   * Resolved an issue where the custom branding logo was unintentionally being shown, even when disabled.

---

### Tested On:

* WordPress default themes from **Twenty Twenty (2020)** to **Twenty Twenty-Five (2025)**.
* Confirmed `volumePanel` and branding visibility logic across different skin types.

PR #813, opened 2025-08-04
Fix disable inline volume panel for Minimal and Pills player skins

Juzar10:
## ✨ Overview

Issue – https://github.com/rtCamp/godam/issues/812

This PR resolves the issue for the Volume Panel Slider not being respected for the GoDAM player.

### DIFF
--- a/inc/templates/godam-player.php
+++ b/inc/templates/godam-player.php
@@ -184,6 +184,17 @@ $video_setup = array(
 );
 if ( ! empty( $control_bar_settings ) ) {
 	$video_setup['controlBar'] = $control_bar_settings; // contains settings specific to control bar.
+
+	if ( isset( $control_bar_settings['volumePanel'] ) && empty( $control_bar_settings['volumePanel'] ) ) {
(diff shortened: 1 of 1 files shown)


Juzar10:
@prachigarg19, 

I noticed that this change was introduced in PR: https://github.com/rtCamp/godam/pull/782.

I couldn’t find any direct usage of it in the current codebase, and removing it seems to make the volume slider work properly. Could you please help confirm if this condition is still needed or if it's safe to remove now?

hbhalodia (approved):
LGTM! Code wise, Need input from @prachigarg19 because - https://github.com/rtCamp/godam/pull/813#issuecomment-3149289775

Juzar10:
@prachigarg19, understood. I have updated the code to only have false value when it's set to false. can you please check and test this.

hbhalodia:
Hi @prachigarg19, Do we have an update on this? Were you able to test the changes and can you please help review the PR.

prachigarg19:
Everything works, can be merged

Issue #821, opened 2025-08-04
Fix compatibility with Elementor


kishan-gondaliya-7270:
ELEMENTOR : 
- [ ] URLs missing for GoDam tab Videos 
- [x] GoDam Icon missing for GoDam Video in Video Upload Picker 
- [x] Fix alignment for GoDAM logo
- [x] Selected video not appearing for Media Library Selected Video  (can be related to Virtual Media)
- [x] Audio File Not Visible 

Refer : https://docs.google.com/document/d/1KDpp3WxD6g8c0_J-p9SR5PzO148xZbgF3V97gyMmWOs/edit?tab=t.j635df7fw91i

### Additional Issues
- [x] SEO settings are not getting autofilled with the GoDAM Tab Videos

PR #849, opened 2025-08-06
Elementor plugin issues

Juzar10:
## ✨ Overview
**Issue** – https://github.com/rtCamp/godam/issues/821
This update improves the GoDAM Elementor audio and video widgets, focusing on SEO handling, media support, and general usability.

## ✅ Changes Made

- Renamed widget type to 'godam-video' across the codebase for consistency and to fix SEO and editor issues.
- Updated godam-audio to support both WordPress and external media sources.
- Improved video SEO metadata by correctly fetching and formatting duration and other info.
- Added checks to prevent errors when media attachments are missing or invalid.
- Replaced audio loader SVG with the GoDAM logo.
- Cleaned up unused options and simplified video widget rendering logic.

### DIFF
--- a/assets/src/js/media-library/views/attachment.js
+++ b/assets/src/js/media-library/views/attachment.js
@@ -1,6 +1,11 @@
 /* global jQuery */
 
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
  * Internal dependencies
  */
 import { isAPIKeyValid, isFolderOrgDisabled } from '../utility';
@@ -106,7 +111,7 @@ const Attachment = wp?.media?.view?.Attachment?.extend( {
 			if ( undefined !== virtual && virtual ) {
 				this.$el.append( `
 					<div class="transcoding-status__loader" data-percent="100">
-						<img src="${ GodamLogo }" alt="GoDAM Logo" width="24" height="24" />
+						<img src="${ GodamLogo }" alt="${ __( 'GoDAM Logo', 'godam' ) }" width="24" height="24" />
 					</div>
 				` );
 				this.$el.addClass( 'transcoding-status--completed' );


assets/src/js/elementor/widgets/godam-video.js
```diff
@@ -87,3 +110,121 @@ window.addEventListener( 'elementor/frontend/init', () => {
...
+		};
+
+		video.onerror = () => {
+			reject( 'Failed to load DASH video metadata.' );
+		};
+
+		// Set a timeout to avoid hanging
+		setTimeout( () => {
+			reject( 'Timeout loading video metadata' );
+		}, 10000 );
```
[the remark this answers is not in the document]
Juzar10: It's not being resused, so this is not necessary I believe.

Juzar10:
Some functions are shared between the Gutenberg block and the Elementor block, and it’s likely the same will apply to other integrations as well. Currently, using these functions would require importing them from the Gutenberg implementation, which creates an awkward dependency and a non-intuitive import path.

Since each block should be self-contained and not rely on functions defined within another block, this approach is not ideal.

In the future, we can extract these commonly used functions into a global-level utility file (e.g., utils.js). This will allow us to import and reuse them across blocks without introducing inter-block dependencies.

assets/src/js/elementor/widgets/godam-video.js
```diff
@@ -87,3 +110,133 @@ window.addEventListener( 'elementor/frontend/init', () => {
...
+						const duration = parsePT( mpdElement.getAttribute( 'mediaPresentationDuration' ) );
+						resolve( duration );
+						return;
+					}
+
+					// Fallback: try to create video element with MSE if available
+					if ( window.MediaSource ) {
+						createVideoElementForDash( videoUrl ).then( resolve ).catch( reject );
+					} else {
+						reject( 'Unable to determine video duration from MPD file' );
```
hbhalodia, line 145: Do we need to localize this?

assets/src/js/media-library/views/attachment.js
```diff
@@ -105,25 +106,7 @@ const Attachment = wp?.media?.view?.Attachment?.extend( {
...
-							<stop stop-color="#AB3A6C"/>
-							<stop offset="1" stop-color="#E6533A"/>
-							</linearGradient>
-							<linearGradient id="paint2_linear_76_96" x1="11.1281" y1="21.5288" x2="17.202" y2="15.4549" gradientUnits="userSpaceOnUse">
-							<stop stop-color="#AB3A6C"/>
-							<stop offset="1" stop-color="#E6533A"/>
-							</linearGradient>
-							</defs>
-						</svg>
+						<img src="${ GodamLogo }" alt="GoDAM Logo" width="24" height="24" />
```
hbhalodia, line 109: Alt needs to be localized.

hbhalodia (changes requested):

hbhalodia (approved):

Issue #824, opened 2025-08-04
Fix pagination for GoDAM videos on WordPress, it should be showing x of y


kishan-gondaliya-7270:
Fix pagination for GoDAM videos on WordPress, it should be showing x of y

<img width="630" height="348" alt="Image" src="https://github.com/user-attachments/assets/69367a08-cf3a-49dc-b923-cf85b05850e7" />

Akashshelke07:
Hello @nayemDevs ,
I've finished the work on the pagination fix. The PR has been updated to resolve all conflicts and is ready for the next steps.
Please let me know if any changes are needed.
Best regards,
Akash Shelke

Akashshelke07:
Hi @hbhalodia ,

I’ve completed the requested change and switched to the develop branch as asked.

Thanks!

hbhalodia:
Hi @Akashshelke07, Our team have left the comment on the PR you have raised, Could you please confirm that - https://github.com/rtCamp/godam/pull/875#issuecomment-3173294668

Juzar10:
This issue requies changes from the godam-core side, where API should return the count of the attachment present along with the attachments for proper pagaination calculation.

subodhr258:
Since it has a dependency on Central, removing 1.4.0 label 
cc: @kishan-gondaliya-7270

PR #850, opened 2025-08-06
Fix/pagination count

Akashshelke07:
**Resolves Issue #824**

**Description:**
This pull request fixes a bug in the WordPress media library (`upload.php`) where the pagination count was displaying incorrectly (e.g., "Showing 160 of 160") even when more items were available to be loaded.

The issue was that the custom GoDAM `AttachmentsBrowser` view was not reading the total item count provided by WordPress and therefore could not display the "Showing X of Y" text correctly.

**Changes Made:**
- Modified `assets/src/js/media-library/views/attachment-browser.js`.
- Added a custom `updateCount()` method to the `AttachmentsBrowser` view.
- This new method correctly reads the total number of attachments from `this.collection.props.get('total')` and updates the text display accordingly.
- This ensures the loaded item count and the total item count are displayed correctly, providing an accurate user experience.

### DIFF
--- a/assets/src/blocks/godam-gallery/block.json
+++ b/assets/src/blocks/godam-gallery/block.json
@@ -71,12 +71,18 @@
 	},
 	"supports": {
 		"anchor": true,
-		"align": [ "left", "center", "right", "wide", "full" ],
+		"align": [ "wide", "full" ],
 		"spacing": {
 			"margin": true,
 			"padding": true
 		}
 	},
+	"parent": [
+		"core/group",
+		"core/columns",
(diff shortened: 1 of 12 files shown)


yashjawale:
Hey @Akashshelke07 thanks for submitting the PR! 🙇‍♂️
Although, the PRs need to have `develop` as base branch instead of `main`
It also allows automated tasks to check the PRs
More information available in [CONTRIBUTING.md](https://github.com/rtCamp/godam/blob/main/CONTRIBUTING.md)

Issue #716, opened 2025-07-22
Create Recorder Field for Everest Forms


mi5t4n:
Create Recorder Field similar to WPForms, Gravity Forms, etc.

nayemDevs:
@mi5t4n / @subodhr258 It was done and released, right?

mi5t4n:
@nayemDevs No, it's still in-progress. I am facing issues with the way EVF is sending form data through ajax. Our way implementation that worked in previous forms is not working here, finding a workaround the issue.

hbhalodia:
Sure @nayemDevs, I can pick this up. 

@mi5t4n, Let me know if we can connect on this related to blocker and progress.

Thanks.

renishsurani:
Please note that I have used [this](https://github.com/rtCamp/godam/actions/runs/17259469030) ZIP file of the GoDAM plugin.

---

- [ ] When the GoDAM API is set in the backend and the form is submitted, an AJAX error is displayed instead of the success submission message. However, the form entries are saved successfully in the backend. (If I remove the API key from the backend, the success message displays correctly.)

<img width="400" height="1578" alt="Image" src="https://github.com/user-attachments/assets/82ba4ec8-b4cb-4a99-9c86-b99fdefbeb55" />

---

- [ ] The 'Record Video' button style looks broken on the full screen.

<img width="500" height="1057" alt="Image" src="https://github.com/user-attachments/assets/a61ab387-39ed-4db9-80dc-c790358a8c54" />

---

- [ ] When we click on the 'Record Video' button in full screen, the 'Import files from:' screen is not visible. It opens behind the full screen.

https://github.com/user-attachments/assets/7700cfbd-0901-4e65-a055-c53b37e7b2d6

---

- [ ] While the video is uploading, the 'Submit' button should be disabled.

<img width="500" height="700" alt="Image" src="https://github.com/user-attachments/assets/2b1a85e4-955a-40f9-b9c0-b9294a28929a" />

---

Hi @mi5t4n @hbhalodia, 

While testing the record feature, I noticed that the submitted media is not stored in the Media Library tab or in Central. I just wanted to confirm if this is the expected behaviour or an issue. Please let me know so that I can verify and report those issues accordingly.

cc: @kishan-gondaliya-7270

kishan-gondaliya-7270:
Released with 1.4.0

PR #861, opened 2025-08-07
Add recorder field for the everest forms

hbhalodia:
- Issue - #716


admin/class-rtgodam-transcoder-rest-routes.php
```diff
@@ -332,6 +332,33 @@ public function handle_callback( WP_REST_Request $request ) {
 			}
 		}
 
+		if ( ! empty( $job_for ) && 'everestforms-godam-recorder' === $job_for && ! empty( $job_id ) && class_exists( 'EverestForms' ) ) {
```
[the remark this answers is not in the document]
hbhalodia: False positive.

inc/classes/everest-forms/class-everest-forms-integration.php
```diff
@@ -0,0 +1,250 @@
...
+
+		/**
+		 * Bail early if no file to send.
+		 */
+		if ( empty( $file_url ) ) {
+			return;
+		}
+
+		/**
+		 * Form Title.
```
mi5t4n, line 204: We may skip these comments, as the code itself reveals the intent. The comment is only making the function larger. For single line comment, better to use `//` .

assets/src/js/godam-recorder/index.js
```diff
@@ -223,6 +224,18 @@ class UppyVideoUploader {
 				},
 			),
 		);
+
+		this.fileInput.dispatchEvent(
```
mi5t4n, line 228: There is already an event dispatch, similar to this, only the name is different. Is it possible to use that ?

inc/classes/everest-forms/everest-forms-field-godam-record-frontend.php
```diff
@@ -0,0 +1,83 @@
...
+ * @package GoDAM
+ *
+ * @since n.e.x.t
+ */
+
+// Get the primary field data.
+$primary = $field['properties']['inputs']['primary'] ?? array();
+
+// Get the required data.
+$max_file_size          = $field['max_size'] ?? (int) ( wp_max_upload_size() / ( 1024 * 1024 ) );
```
mi5t4n, line 14: When the max_size is not set, the default is an empty string. The null coalescing operator operator does not check for empty string. So, if the max_size is not set, the $max_file_size will be 0, which is supposed to be the max upload size supported the environment.

inc/classes/everest-forms/class-everest-forms-field-godam-video.php
```diff
@@ -0,0 +1,599 @@
...
+				}
+			}
+
+			// Add override style for everest forms.
+			$style = '<style>#everest-forms-entry-fields:not(.postbox) table tbody tr td span {margin: 0 !important;}</style>';
+
+			// Override style for video preview.
+			$style .= '<style>.evf-godam-video-preview .easydam-video-container{height:100%;}.evf-godam-video-preview .easydam-video-container .easydam-player.video-js{margin-top:0;}</style>';
+
+			/**
```
mi5t4n, line 227: Again, code is pretty much self explanatory. Let's avoid comment similar to this, if we are using single line better to use `//`

admin/class-rtgodam-transcoder-rest-routes.php
```diff
@@ -332,6 +332,38 @@ public function handle_callback( WP_REST_Request $request ) {
...
```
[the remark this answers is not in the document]
hbhalodia: Resolved.

inc/classes/everest-forms/class-everest-forms-integration.php
hbhalodia: Resolved. And address at other places as well.

inc/classes/everest-forms/everest-forms-field-godam-record-frontend.php
hbhalodia: Resolved.

Issue #862, opened 2025-08-07
Fix UX For GoDAM Plugin tabs


kishan-gondaliya-7270:
Responsiveness issue for 
<img width="874" height="654" alt="Image" src="https://github.com/user-attachments/assets/a968fd6f-28b8-475a-909d-4c1d56e77614" />

https://github.com/user-attachments/assets/8c843671-632b-49ac-b6e2-52ec2a9c114c

Steps : 
- Turn on debugger 
- Observe responsiveness and UX experience in GoDAM Plugin

hbhalodia:
- The Fix is added in the PR - https://github.com/rtCamp/godam/pull/870.

subodhr258:
@kishan-gondaliya-7270  The fix for the warning is done, and since it is an important fix we are planning to add it to the patch release.

kishan-gondaliya-7270:
Issue is Fixed and Released with 1.3.3, Issue related to alignment is tracked as sub issue

PR #870, opened 2025-08-08
Fix: Remove unused wp_register_script for JS file that is not present in build - #862

hbhalodia:
- Issue - #862 
- Fixes the issue related to the JS file being enqueue which is removed, but forget to remove register_script for the same, hence unable to fetch the filemtime for that JS file because build process does not include to add that script.

### DIFF
--- a/inc/classes/wpforms/class-wpforms-integration.php
+++ b/inc/classes/wpforms/class-wpforms-integration.php
@@ -62,14 +62,6 @@ class WPForms_Integration {
 		);
 
 		wp_register_script(
-			'wpforms-godam-recorder',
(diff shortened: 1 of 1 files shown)


Mukulsingh27:
[This](https://github.com/rtCamp/godam/blob/main/assets/src/js/wpforms-godam-recorder-editor.js) file?

hbhalodia:
> [This](https://github.com/rtCamp/godam/blob/main/assets/src/js/wpforms-godam-recorder-editor.js) file?

No, this is for editor and this is being registered below what we have removed.

So the pick here is, while working on wpforms and other integrations with recorder forms, we have build a common recorder JS file which needs to be enqueued and at that time wpforms were also being worked on, and after that the file was removed (as I think) but the php still was enqueing the script.

Issue #790, opened 2025-07-31
Refactor frontend.js file code implementation


hbhalodia:
## Issue

- [ ] The file - https://github.com/rtCamp/godam/blob/main/assets/src/js/godam-player/frontend.js, needs refactoring as it would become heavy to maintain the file, as new developers directly adding code and making it messy.
- [ ] Should convert to class based implementation.
- [ ] Define proper function for utils, condition checks.
- [ ] Should be efficient in using `if..else`.

## Try
- Can use github copilot, ask it to refactor file with DRY, Class based implementation and make sure it does not breaks.

Cc: @subodhr258 @Mukulsingh27

Juzar10:
@hbhalodia, I agree.

I recently created a PR: https://github.com/rtCamp/godam/pull/768, where I tried to modularize the code for something like hoverManager.

I was hoping that we could interactively refactor the `frontend.js`, which is just a big JS file with almost no separation, making it harder to debug any issues.

hbhalodia:
Here is the PR - https://github.com/rtCamp/godam/pull/802

Mukulsingh27:
Thanks @hbhalodia @Juzar10 for the cleanup, when are planning to merge this in develop?

hbhalodia:
No @Mukulsingh27, Not for now, and we are not planning to add it on the `1.3.1` release. It should we well tested, as refactoring is quite large.

hbhalodia:
Hi Team, Need some through testing on the PR - https://github.com/rtCamp/godam/pull/797, Aims to modularise the frontend.js code which is more scalable and easy to maintain.

Juzar10:
@hbhalodia, yes good idea. For above data can you tell me how many videos were on the page? I want to check if number of videos makes any change or not.

Also we should compare before and after the refactor execution time, although I don't expect much changes.

KMchaudhary:
Refactoring is already done. Outdated issue

PR #871, opened 2025-08-08
fix: JS translation issue with translation plugins

mi5t4n:
## Description
This PR addresses an issue where JavaScript strings weren't being translated. The problem occurred because the JS files were minified `(*.min.js)`, and the `wp i18n make-pot` command always skips these files. As a result, the necessary strings were not included in the POT file. This fix ensures that unminified files are used for proper translation string extraction.

## Reference
https://developer.wordpress.org/cli/commands/i18n/make-pot/#options

```log
[--exclude=<paths>]
The following files and folders are always excluded: node_modules, .git, .svn, .CVS, .hg, vendor, *.min.js.
```

## How
1. Install and activate [Loco Translate](https://wordpress.org/plugins/loco-translate/) plugin.
2. Switch the site’s language to `English (UK)` by navigating to `Settings > General > Language`.
3. Go to the `Translate Loco` and change the `General Settings` text to `General Settings UK` for the plugin.
4. Go to the `GoDam` menu page.

## Issue Recording
[Kazam_screencast_00004.webm](https://github.com/user-attachments/assets/d6ade9da-f00e-4a53-a947-d0c9445ed362)

## Solution Recording
[Kazam_screencast_00005.webm](https://github.com/user-attachments/assets/aff6c20a-0a7c-4f58-a356-58ae0f146364)

Closes - #407 

Duplicate - #408

### DIFF
--- a/languages/godam.pot
+++ b/languages/godam.pot
@@ -9,7 +9,7 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"POT-Creation-Date: 2025-08-13T06:12:19+00:00\n"
+"POT-Creation-Date: 2025-08-13T07:33:58+00:00\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "X-Generator: WP-CLI 2.12.0\n"
 "X-Domain: godam\n"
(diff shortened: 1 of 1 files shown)


hbhalodia:
I guess this is because of large file that is being generated after the build. This is getting exhausted `frontend-player.js` at my end. I guess we should create chunk of that file, from webpack and then enqueue the chunks could be beneficial.

I have'nt tried it but that could be probably a solution to go because previously it was failed because for large `.map` files, and I see the large files, it passed earlier because it were min.js which were excluded.

Cc: @mi5t4n

hbhalodia:
> Currently, if the memory limit is set to 1GB, we don't have a memory exhaustion issue. But, still large assets files needs to be minized. I will look into this.

Yes agreed, but we cannot gurantee that but - https://wordpress.slack.com/archives/C0341VCNL/p1753405130757349?thread_ts=1747659100.982119&cid=C0341VCNL (It suggests to have 1GB on WP platform when running the command). But definately need to look on solution for the same.

KMchaudhary (approved):

mi5t4n:
@subodhr258 Have we created an issue for that?

subodhr258:
@mi5t4n this is one related issue created by @hbhalodia : https://github.com/rtCamp/godam/issues/790

Issue #884, opened 2025-08-11
QA: UI Broken issues


renishsurani:
While testing other tickets, I found a few UI issues.

subodhr258:
2 is duplicate of https://github.com/rtCamp/godam/issues/872

kishan-gondaliya-7270:
@renishsurani pick up this task

renishsurani:
I have verified that the three reported issues are resolved in the development environment.
cc: @kishan-gondaliya-7270 @subodhr258

renishsurani:
- [ ] The media library section does not look good on small screens (1024px); it should be properly aligned for better user access.

<img width="1021" height="474" alt="Image" src="https://github.com/user-attachments/assets/bbb435f2-0172-462f-b8b4-003b59f50e94" />

When the option to enable folder organization in the media library is disabled, the media library buttons appear different in size and have misaligned spacing.

**Steps to reproduce the issue:**

1. Navigate to the General Settings tab of the GoDAM plugin.
2. Disable the toggle for Enable folder organization in media library.
3. Open the Media CPT and verify the buttons.

<img width="767" height="680" alt="Image" src="https://github.com/user-attachments/assets/cfcb7442-971b-43c6-af32-740a7923a453" />

kishan-gondaliya-7270:
Released with 1.4.0

Issue #872, opened 2025-08-08
Video Editor Forms Dropdown UI


subodhr258:
Video Editor Dropdown to select the form has a bug in the hover effect: 

https://github.com/user-attachments/assets/2f1d81ae-b37d-4536-a629-96241a8bdc94

Expected:

<img width="528" height="419" alt="Image" src="https://github.com/user-attachments/assets/69590464-a9a2-40b4-8f2e-0aefe3ba1bd4" />

IndiraBiswas:
This is working fine.

<img width="1461" height="780" alt="Image" src="https://github.com/user-attachments/assets/d0dfd288-e317-49dd-a88d-f6adecd177b8" />

PR #893, opened 2025-08-11
fix: Video Editor Accent Color

subodhr258:
This pull request makes a small visual update to the admin interface by changing the accent color used for WordPress components in the `#root-easydam` and `#root-video-editor` sections. The accent color is now set to `#ab3a6c` instead of using the `$godam-primary-text-color` variable.


subodhr258:
Fixed: https://github.com/rtCamp/godam/issues/872
Partially Fixed: https://github.com/rtCamp/godam/issues/884

Issue #819, opened 2025-08-04
Fix load more behavior for Folders in Plugin


kishan-gondaliya-7270:
Fix : https://github.com/rtCamp/godam/issues/478#issuecomment-3150498882

renishsurani:
Hi @Juzar10, as discussed, please make the necessary changes to hide the "Load More" button when there are no additional folders. 

Let me know once this is pushed to the develop branch.

cc: @kishan-gondaliya-7270

renishsurani:
I have verified the issue mentioned above. I have checked that the “Load More” functionality works as expected. 

Note: if the folder count is 20 or less, the Load More button is not visible; if the folder count is above 20, it becomes visible.

I have checked this feature for [this](https://github.com/rtCamp/godam/actions/runs/17155069088) published plugin version.

<img width="1300" height="874" alt="Image" src="https://github.com/user-attachments/assets/27c59cdc-3191-4bf9-a796-b66f2b8f676e" />

cc: @kishan-gondaliya-7270 @Juzar10

PR #896, opened 2025-08-12
Update default items per page to 20 in media library API

Juzar10:
## Summary:

ISSUE: https://github.com/rtCamp/godam/issues/819

This PR updates the default pagination size for media folders in both backend and frontend, increasing items per page from 10 → 20 for better consistency and user experience.
Changes:

**Backend**: Updated get_media_folders to default per_page to 20 when unset or invalid.
**Frontend**: Updated folderApi to default per_page to 20 for non-special cases, matching backend behavior.

### DIFF
--- a/inc/classes/rest-api/class-media-library.php
+++ b/inc/classes/rest-api/class-media-library.php
@@ -1363,7 +1363,7 @@ class Media_Library extends Base {
 			$per_page = (int) $request->get_param( 'per_page' );
 
 			if ( $per_page < 1 ) {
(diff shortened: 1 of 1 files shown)


subodhr258 (approved):

KMchaudhary (commented):
https://github.com/user-attachments/assets/47073550-5daa-455b-996d-e07af807d9da

We should hide the load more option if no more folder remains to display

Juzar10:
@KMchaudhary, to avoid one extra request to just get count of the folders, I have tried this method which fetches and if not found then hide it.

Issue #942, opened 2025-08-20
Enable hover-to-open behavior for captions & settings menus


prachigarg19:
Currently, the captions (CC) and settings menus in the GoDAM video player only stay open when the user explicitly clicks the button. This creates a usability problem: when hovering from the button to the submenu, the menu often closes prematurely.

Ref: https://app.godam.io/web/video/056gc7995o

PR #943, opened 2025-08-20
Enable hover-to-open behavior for captions & settings menus

prachigarg19:
Resolves issue #942 

1. Add a new MenuButtonHoverManager class.
2. Observe and attach listeners to both SubsCapsButton and SettingsButton.
3. Keep their menus open on mouseenter of the button or submenu.
4. Close the menu only when the mouse leaves both the button and submenu.
5. Integrate the manager into ControlsManager after control bar setup.


prachigarg19:
@hbhalodia You can add the captions using block editor

Ref: https://godam.io/docs/gutenberg-blocks/configuring-the-video-block/#how-to-upload-captions

PR #960, opened 2025-08-22
Add pagination count proper instead of normal optimistic pagination for media library

Juzar10:
## ✨ Updates

**ISSUE**: https://github.com/rtCamp/godam/issues/819#issuecomment-3210346854

- Backend now returns pagination headers (X-WP-Total, X-WP-TotalPages) 
- New helper get_total_media_folders_count for total count 
- Frontend parses headers → adds total & totalPages 
- Redux slice: hasNext ➝ totalPages
- React components (FolderTree, BookmarkTab, LockedTab) updated for new response
- Removed unused useEffect cleanup 🧹

### DIFF
--- a/inc/classes/rest-api/class-media-library.php
+++ b/inc/classes/rest-api/class-media-library.php
@@ -1347,7 +1347,7 @@ class Media_Library extends Base {
 			'hide_empty' => false,
 			'orderby'    => 'name',
 			'order'      => 'ASC',
-			'per_page'   => 100, // Default to 100 items per page.
+			'number'     => 100, // Default to 100 items per page.
 		);
 
 		// Initialize meta_query as empty array.
@@ -1371,6 +1371,9 @@ class Media_Library extends Base {
 			$args['meta_query'] = $meta_queries; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- Meta query is needed to filter by bookmark and locked.
 		}
 
+		$total_pages = 0;
+		$total_items = 0;
+
 		if ( ! $locked && ! $bookmark ) {
 			$page = (int) $request->get_param( 'page' );
 
@@ -1388,6 +1391,9 @@ class Media_Library extends Base {
 			$args['offset'] = ( $page - 1 ) * $per_page;
 
 			$args['parent'] = (int) ( $request->get_param( 'parent' ) ?? 0 );
+
+			$total_items = $this->get_total_media_folders_count();
+			$total_pages = ceil( $total_items / $per_page );
 		}
 
 		$terms = get_terms( $args );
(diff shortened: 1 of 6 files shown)


pages/media-library/components/folder-tree/FolderTree.jsx
```diff
@@ -412,11 +406,12 @@ const FolderTree = ( { handleContextMenu } ) => {
 						} ) }
 					</SortableContext>
 				</div>
-				{ page.hasNext && ( <button
+				{ ( ( currentPage < page.totalPages ) || ( isFetching && currentPage === page.totalPages ) ) && ( <button
```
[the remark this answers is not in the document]
Juzar10: I don't think it's that complex that it warrants another variable.

inc/classes/rest-api/class-media-library.php
```diff
@@ -1396,7 +1402,26 @@ public function get_media_folders( $request ) {
...
+
+		// Create the response.
+		$response = rest_ensure_response( $prepared_terms );
+
+		// Add headers only for paginated requests.
+		if ( ! $locked && ! $bookmark ) {
+			$response->header( 'X-WP-Total', $total_items );
+			$response->header( 'X-WP-TotalPages', $total_pages );
+			$response->header( 'X-WP-Current-Page', $page );
+			$response->header( 'X-WP-Per-Page', $per_page );
```
hbhalodia, line 1421:
```suggestion
			$response->header( 'X-Wp-Total', $total_items );
			$response->header( 'X-Wp-Totalpages', $total_pages );
			$response->header( 'X-Wp-Current-Page', $page );
			$response->header( 'X-Wp-Per-Page', $per_page );
```

Because, response is converting it to these headers name, So we can make this consistent.

Image

inc/classes/rest-api/class-media-library.php
```diff
@@ -1434,6 +1459,28 @@ private function get_all_children_terms( $terms, $taxonomy ) {
 		return $all_terms;
 	}
 
+	/**
+	 * Get the total count of media folders.
+	 *
+	 * @return int Total media folders count.
+	 */
+	private function get_total_media_folders_count() {
```
hbhalodia, line 1467: Can we add a comment such that it would return count for only top level parents? I was confused because I was having 25 folders in total, but some of them were child, but for load more functionality since we are only counting the parent folders, so comment would help others to indentify why only parent were considered.

pages/media-library/redux/api/folders.js
```diff
@@ -52,6 +52,18 @@ export const folderApi = createApi( {
 					},
 				};
 			},
+			transformResponse: ( responseData, meta ) => {
+				// Extract headers from meta.response.headers
+				const headers = meta.response.headers;
+				const totalItems = headers.get( 'X-WP-Total' ) || headers.get( 'x-wp-total' );
+				const totalPages = headers.get( 'X-WP-TotalPages' ) || headers.get( 'x-wp-totalpages' );
```
hbhalodia, line 59: Needs to updated with backend when you will look at first review comment.

hbhalodia (changes requested):
Have added some feedbacks, Can you please look on that?

inc/classes/rest-api/class-media-library.php
Juzar10: Yes, I agree. also updated function name to specify parent.

Juzar10:
@hbhalodia, updated the code according to review. please check.

hbhalodia (approved):
LGTM!

Tested and working as expected.

subodhr258 (approved):

PR #968, opened 2025-08-22
dragging folder to create child of parent folder gives 400 response

om-ghante:
Problem

When dragging and dropping folders, the app was sending wrong data (like "true"/"false" or 1/0 instead of real booleans).
Because of this, the backend API was rejecting the request with a 400 Bad Request error.

Solution

Made sure isOpen and locked values are always sent as proper booleans (true/false).

Wrapped the API call in try/catch so errors show in the snackbar.

Only update the folder state after the API call succeeds.


hbhalodia:
Hi @om-ghante, Thanks for the PR. Unfortunately we already had the PR which fixed the issue https://github.com/rtCamp/godam/pull/960 and is merged. 

Thanks again for raising the PR, For now I am closing this PR. You can lookout for other issues in the repository to contribute.

PR #996, opened 2025-08-28
Implement custom fullscreen feature for iOS devices

KMchaudhary:
On iOS devices, the GoDAM video player cannot enter fullscreen mode using the default Video.js option. This causes the video to appear blank without layers, ads, or the GoDAM player skin.

Ref: https://github.com/rtCamp/godam-core/issues/124

**Solution:**
Detect if the user is on iOS, then replace the default Video.js fullscreen button with a custom one. The custom button will handle full-screen behavior manually, allowing us to simulate fullscreen mode on iOS without triggering the native video player.

- [x] Add custom fullscreen button.
- [x] Make sure that the custom fullscreen UI does not change compared to the default fullscreen button on all available skins.
- [x] Add a check to detect if an iOS device, then replace the default full-screen button with a custom full-screen button.
- [x] Fix the pill skin fullscreen breaking UI

**Before**
iOS default video player
Image

**After**
| Mimic fullscreen | with layers | Classic skin |
|----------|----------|----------|
| WhatsApp Image 2025-08-27 at 18 06 32 | WhatsApp Image 2025-08-27 at 18 06 33 | WhatsApp Image 2025-08-27 at 18 08 07

### DIFF
--- a/assets/src/js/godam-player/managers/controlsManager.js
+++ b/assets/src/js/godam-player/managers/controlsManager.js
@@ -112,7 +112,9 @@ export default class ControlsManager {
 				}
 			}
 		}
-		videojs.registerComponent( 'CustomFullscreenButton', CustomFullscreenButton );
+		if ( ! videojs.getComponent( 'CustomFullscreenButton' ) ) {
+			videojs.registerComponent( 'CustomFullscreenButton', CustomFullscreenButton );
+		}
 		return new CustomFullscreenButton( this.player );
 	}
 
@@ -145,7 +147,9 @@ export default class ControlsManager {
 				}
 			}
 		}
-		videojs.registerComponent( 'CustomFullscreenExitButton', CustomFullscreenExitButton );
+		if ( ! videojs.getComponent( 'CustomFullscreenExitButton' ) ) {
+			videojs.registerComponent( 'CustomFullscreenExitButton', CustomFullscreenExitButton );
+		}
 		return new CustomFullscreenExitButton( this.player );
 	}
 
--- a/assets/src/js/godam-player/managers/controlsManager.js
+++ b/assets/src/js/godam-player/managers/controlsManager.js
@@ -44,7 +44,7 @@ export default class ControlsManager {
 		this.setupControlBarComponents( controlBarSettings );
 		this.setupCustomPlayButton( controlBarSettings );
 
-		this.setupFullscreenButton();
+		this.setupCustomFullscreenButton();
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -68,6 +70,108 @@ export default class ControlsManager {
...
+				if ( godamVideoContainer ) {
+					godamVideoContainer.classList.remove( 'godam-video-fullscreen' );
+				}
+			}
+		}
+		videojs.registerComponent( 'CustomFullscreenExitButton', CustomFullscreenExitButton );
+		return new CustomFullscreenExitButton( this.player );
+	}
+
+	checkIOSDevice() {
```
Juzar10, line 152: Add function comment here.

assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -68,6 +70,108 @@ export default class ControlsManager {
...
+		const userAgent = window.navigator.userAgent.toLowerCase();
+		return /iphone|ipad|ipod/.test( userAgent );
```
Juzar10, line 154: Need to confirm this userAgent with a wider range of device, QA can help with that. but if any issue we can expand this check by using `navigator.maxTouchPoints > 1`

assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -68,6 +70,108 @@ export default class ControlsManager {
...
+	}
+
+	/**
+	 * Setup fullscreen button for iOS devices
+	 */
+	setupFullscreenButton() {
```
Juzar10, line 160: As this function is only really for ios, maybe it would be better to rename it to something more representative of that.

something like `maybeSetupFullSCreenButton` or `setupFullScreenButtonForIOS`.

let me know your thoughts. so while future debugging we know that this function would not do anything on other devices.

Juzar10 (approved):
Looks mostly fine, but few suggestions added.

Also, just so you know, we need to test this on IOS device and an iPad. I could not test from my end as I have no IOS devices at the moment 💸📱❌.

So approving without testing.

assets/src/js/godam-player/managers/controlsManager.js
KMchaudhary: This issue is only occurring on iOS devices. I can ask Manoj to check the list of devices on which full-screen open **device default video player**, and later I can customize this function for that devices as well.

assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -68,6 +70,114 @@ export default class ControlsManager {
...
+	}
+
+	/**
+	 * Check if the device is iOS
+	 *
+	 * @return {boolean} - True if iOS device, false otherwise
+	 */
+	checkIOSDevice() {
+		const userAgent = window.navigator.userAgent.toLowerCase();
+		return /iphone|ipad|ipod/.test( userAgent );
```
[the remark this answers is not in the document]
KMchaudhary: Incorrect input

assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -68,6 +70,114 @@ export default class ControlsManager {
...
+			createEl() {
+				const el = super.createEl();
+				el.classList.add( 'vjs-fullscreen-control', 'vjs-control', 'vjs-button' );
+
+				return el;
+			}
+
+			handleClick( e ) {
+				const videoContainer = e.target.closest( '.video-js' );
+				const godamVideoContainer = e.target.closest( '.easydam-video-container' );
```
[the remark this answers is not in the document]
KMchaudhary: Not required

Issue #1022, opened 2025-09-02
Replace Default Preview Media Player in Media Library with GoDAM video player


KMchaudhary:
Currently, on the WordPress media page, when a user tries to see the preview of Virtual Media, it shows media format not supported errors for `.mpd` and `.m3u8` videos. The default WordPress media player does not support those formats, so it would be great if we replaced the default media player with the GoDAM video player.

<img width="1470" height="723" alt="Image" src="https://github.com/user-attachments/assets/7cc95957-2367-4799-9037-6fca001d5c9f" />

<img width="1470" height="723" alt="Image" src="https://github.com/user-attachments/assets/36673f46-8ef8-463a-a9a3-e0ce43d243f7" />

KMchaudhary:
Outdated issue. Virtual media video preview started appearing.

<img width="1470" height="733" alt="Image" src="https://github.com/user-attachments/assets/ba69740a-b2f4-425d-8078-f6d64a99b471" />

PR #1033, opened 2025-09-03
Fix: Replace Default video player in attachment details

DarkMatter-999:
part of #1022

- [x] add VideoJS player support for attachment-details page
- [x] add VideoJS player support for attachment-details-two-column page

### DIFF
--- a/assets/src/js/media-library/utility.js
+++ b/assets/src/js/media-library/utility.js
@@ -198,4 +198,49 @@ function createVideoJsPlayer( container, { videoId, sources, playerOptions = {} 
 	}, 100 );
 }
 
-export { isAPIKeyValid, checkMediaLibraryView, isUploadPage, isFolderOrgDisabled, addManageMediaButton, getQuery, getGodamSettings, canManageAttachment, createVideoJsPlayer };
+/**
+ * Adds transcoded sources (MPD/HLS/MP4) to an existing MediaElement player element.
+ *
+ * @param {HTMLMediaElement} mediaElement       - The <video> DOM element used by MediaElement player.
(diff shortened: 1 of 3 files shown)


DarkMatter-999:
Would a simple toggle work, we can keep it enabled by default.

Also should it be part of "General Settings" or "Video Player settings" since for the latter we only give customization option and it may be confusing to the user.
 
image

@mi5t4n @KMchaudhary

KMchaudhary:
@mi5t4n there are some edge cases, like in the case of virtual media, the user will not be able to see the preview for those videos, as videos are in `.mpd` and `.m3u8` formats, which are not supported by the default media player. 

According to me, we should not provide any admin option for the same.

@hashibhere what's your view on whether we should provide any admin option for the same? If the option is set to false, then the user will not be able to see the preview.

DarkMatter-999:
@KMchaudhary since we are overriding default WordPress functionality, we should give an option to opt out of it which can be kept enabled by default.

When disabled we can have the default player for the local attachments, while since it cant play virtual attachments, there we can use our custom player.

mi5t4n:
@KMchaudhary, my perspective on this is that we are modifying core behavior without providing the user with a proper fallback. If the player fails to work with attachments or conflicts arise with other plugins, it could negatively impact the user experience. In such cases, the only option available to the user would be to disable our plugin, which is not ideal, especially since most videos on the frontend may function correctly.

I still believe we should continue using the default player for virtual attachments as well. For the time being, we can adopt @DarkMatter-999’s solution: provide a disable option that turns off our player for local attachments while continuing to use our player for virtual attachments. We can revisit the handling of virtual attachments in a future update.

cc @hashibhere

PR #1071, opened 2025-09-09
fix: keep captions/settings menu open on hover until user fully leaves

prachigarg19:
This update improves the behavior of Video.js menu buttons (captions, settings, etc.) by ensuring their dropdown menus remain open while the user hovers over either the button or the menu itself.

Previously, the menu would remain stuck open when moving horizontally across the button area, since the .vjs-menu element is nested inside the button’s DOM. The old logic relied on relatedTarget containment checks, which failed in this structure.

The new implementation uses simple hover state tracking (overBtn, overMenu) with a small transition delay. The menu now closes only when the mouse fully leaves both the button and the menu, preventing accidental closures and fixing the “stuck open” bug.

Resolves issue: https://github.com/rtCamp/godam/issues/1088

https://github.com/user-attachments/assets/8f26f318-97de-476c-9d3e-7e2e404a9b46

Fix:

https://github.com/user-attachments/assets/a3b12150-743e-496d-832b-81648f837a36

### DIFF
--- a/assets/src/js/godam-player/managers/menuButtonHover.js
+++ b/assets/src/js/godam-player/managers/menuButtonHover.js
@@ -27,6 +27,9 @@ class MenuButtonHoverManager {
 				return;
 			}
 
+			// initialize clicked state per button
+			btnEl.dataset.clickedOpen = 'false';
+
 			const menuEl = btnEl.querySelector( '.vjs-menu' );
 			if ( menuEl ) {
 				this.attachMenuListeners( btnEl, menuEl );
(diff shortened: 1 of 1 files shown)


assets/src/js/godam-player/managers/menuButtonHover.js
```diff
@@ -38,39 +38,52 @@ class MenuButtonHoverManager {
...
-				this.hideMenu( menuEl );
-			}
+		menuEl.addEventListener( 'mouseleave', () => {
+			overMenu = false;
+			setTimeout( update, 500 );
 		} );
 	}
 
-	// Hide other menus when entering a new one
-	closeOtherMenus( currentMenu ) {
```
KMchaudhary, line 60: Why are we removing this functionality?

Issue #1174, opened 2025-09-25
Console Errors Appearing with Jetpack Forms Integration


renishsurani:
I have verified the form behavior with Jetpack version 14.8. I didn't see any field-related error messages in the console. However, I did notice a few other errors (screenshot attached). The entries are being displayed properly as expected for both the video layer form and the page form.

<img width="1844" height="914" alt="Image" src="https://github.com/user-attachments/assets/55124130-26f3-4b24-a3b3-c8fd4aded8d0" />

<img width="1578" height="675" alt="Image" src="https://github.com/user-attachments/assets/038c16ae-569c-4f65-9386-70e3fb14d165" />

---

For Jetpack version 14.9, I can see the error message mentioned in the issue even without filling the form—it is displayed as soon as the page is opened, where the video with the form layer is added. However, while submitting the form, the entries are being displayed properly as expected for both the video layer form and the page form.

<img width="1853" height="909" alt="Image" src="https://github.com/user-attachments/assets/40a7ab5e-079d-4e2b-a358-09f2eb8e7df0" />

<img width="1711" height="623" alt="Image" src="https://github.com/user-attachments/assets/f06e1f6e-2c02-45cb-9a5e-1682167e90a7" />

---

I have checked this feature for [this](https://github.com/rtCamp/godam/actions/runs/17849240458) GoDAM plugin version.

cc: @kishan-gondaliya-7270 @nani-samireddy @subodhr258

subodhr258:
Submission works with no console errors. @kishan-gondaliya-7270 please test for 1.4.3

PR #1186, opened 2025-09-26
Fix: Jetpack Form console error

nani-samireddy:
Related issue: https://github.com/rtCamp/godam/issues/1174

### Summary
In Jetpack v14.9 and above, the Interactivity API is used to manage forms. When a Jetpack form is used inside a GoDAM video block, console errors related to field registration are thrown. This happens because the entire form is being replaced on the frontend, causing the Interactivity API scripts to initialize twice.

### Changes made
Added a check to replace the from only if the form is using it is using ajax form submission.

### Screenshots
**Before**

Screenshot 2025-09-26 at 12 36 37 PM

**After**
Screenshot 2025-09-26 at 12 35 06 PM

### DIFF
--- a/assets/src/js/jetpack-form.js
+++ b/assets/src/js/jetpack-form.js
@@ -123,7 +123,7 @@ document.addEventListener( 'DOMContentLoaded', function() {
 
 		let newForm = form;
 
-		// If the form doesn't have `is-ajax-form` class, skip form replacement.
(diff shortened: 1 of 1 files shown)


subodhr258:
@nani-samireddy while the error has disappeared, there is an issue while submitting the form:

Image

subodhr258 (approved):

Issue #1195, opened 2025-09-30
Fluent Form -  Audio Recording for GoDAM Recorder


nayemDevs:
Currently, the GoDAM recorder field supports only Webcam and Screen Recording.
Users have requested the ability to record Audio-only (e.g., voice notes, narration, podcasts).
Since Uppy already provides an @uppy/audio plugin, we can extend the existing recorder integration to support audio recording.

Follow the integration done with Gravity Forms

PR #1204, opened 2025-10-02
Feat: Fluent Form - Audio Recording for GoDAM Recorder

kapilpaul:
## Fluent Form - Audio Recording for GoDAM Recorder

Issue: #1195

#### Changes
* Added audio recording support within Fluent Form using the GoDAM Recorder.
* Fixed video preview issues to ensure smoother playback and preview experience.
* Fixed "Start Recording" button UI inside the editor for better visibility.

**Notes**
All updates are scoped to Fluent Form integration only.
Ensures a more consistent user experience for both audio and video recording features.

### DIFF
--- a/assets/src/css/admin.scss
+++ b/assets/src/css/admin.scss
@@ -691,6 +691,10 @@ $media-frame-width: calc(300px + 2 * 24px);
 	}
 }
 
+.wpf_entry_value:has(.gf-godam-video-preview) {
+	white-space: inherit;
+}
+
 .godam_no_license {
 
 	pointer-events: none;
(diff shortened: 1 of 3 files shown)


assets/src/css/admin.scss
```diff
@@ -871,3 +875,27 @@ to ensure proper responsiveness and avoid layout issues.
...
+	color: transparent;
+	border: none;
+	border-radius: 4px;
+	cursor: pointer;
+	font-size: 14px;
+	position: relative;
+	width: 130px;
+
+	&::before {
+		content: 'Start Recording';
```
mi5t4n: We need static strings to be localized, so we shouldn't do this.

AhmarZaidi: Looks like this issue is dependent on the [above issue](https://github.com/rtCamp/godam/pull/1204#discussion_r2813616414). Adding localized button text should work once the above issue is resolved.

PR #1208, opened 2025-10-03
Fix/video editor capabilities

KMchaudhary:
This pull request introduces an access control check to the video editor page and improves user feedback for unauthorized access. The main focus is to ensure that only users with the correct permissions can edit media items, and to provide a clear error message when access is denied. Additionally, new styles are added for consistent error message presentation.

Access control and user feedback:

* Added a permission check in the `render_video_editor_page` method of `class-pages.php` to ensure only users with the `edit_post` capability for the given media item can access the editor. If the user lacks permission, an error message is displayed.

## Related issue
- #1210

### DIFF
--- a/assets/src/css/admin.scss
+++ b/assets/src/css/admin.scss
@@ -38,6 +38,18 @@ $media-frame-width: calc(300px + 2 * 24px);
 	width: $media-frame-width;
 }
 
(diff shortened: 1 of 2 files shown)


patil-vipul:
Resolve the Copilot issues

PR #1299, opened 2025-11-01
Fix Localhost API Key verification requests

subodhr258:
Issue: https://github.com/rtCamp/godam-core/issues/489

This pull request introduces logic to enhance the security of API key verification by blocking requests made over HTTP or from localhost environments, unless a specific override is present. The main change is to prevent accidental use of production API keys in insecure or development environments, unless explicitly allowed via a constant in `wp-config.php`.

Security improvements for API key verification:

* Added a new function `rtgodam_should_block_api_request()` to determine if an API request should be blocked due to being over HTTP or from a localhost environment.
* Updated `rtgodam_verify_api_key()` to block API key verification on HTTP or localhost unless the `RTGODAM_API_KEY` constant is defined in `wp-config.php`, providing clear error messages for these cases.

## Screenshots
Screenshot 2025-11-01 at 5 51 42 PM
Screenshot 2025-11-01 at 5 57 57 PM

### DIFF
--- a/admin/godam-transcoder-functions.php
+++ b/admin/godam-transcoder-functions.php
@@ -362,6 +362,24 @@ function rtgodam_get_server_var( $server_key, $filter_type = FILTER_SANITIZE_FUL
 }
 
 /**
+ * Check if API request should be blocked due to HTTP or localhost environment.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if request should be blocked, false otherwise.
+ */
+function rtgodam_should_block_api_request() {
+	// Check if current request is over HTTP (not HTTPS).
+	$is_http = ! is_ssl();
+
+	// Check if request is from localhost using existing helper function.
+	$is_localhost = rtgodam_is_local_environment();
+
+	// Block if either condition is met (HTTP OR localhost).
+	return $is_http || $is_localhost;
+}
+
+/**
  * Helper function to verify the api key.
  *
  * @param string $api_key The api key to verify.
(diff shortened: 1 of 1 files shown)


subodhr258:
## Testing Instructions

#### Scenario 1: Verify Blocking on Localhost/HTTP
1.  **Setup:** Use a WordPress environment running on `localhost` or over `http://`.
2.  **Action:** Go to the plugin settings page where you enter the API Key.
3.  **Input:** Enter any API key and attempt to verify/save it.
4.  **Expected Result:**
    *   The verification should **fail**.
    *   You should see an error message similar to:
        > "API key verification is blocked for HTTP connections or localhost environments. To use production API keys on localhost, please add the following constant to your wp-config.php file: define('RTGODAM_API_KEY', 'your-api-key-here');"

#### Scenario 2: Verify Override with Constant
1.  **Setup:** Remain on the same `localhost` or `http://` environment.
2.  **Action:** Open your `wp-config.php` file.
3.  **Edit:** Add the following constant, replacing `12345-TEST-KEY` with the actual API key you intend to test:
    ```php
    define( 'RTGODAM_API_KEY', '12345-TEST-KEY' );
    ```
4.  **Action:** Go back to the plugin settings page.
5.  **Input:** Enter the **exact same key** into the input field and verify.
6.  **Expected Result:** The verification should **succeed** (or proceed to the actual server-side validation).

#### Scenario 3: Verify Mismatch Logic
1.  **Setup:** Ensure `RTGODAM_API_KEY` is still defined in `wp-config.php`
2.  **Input:** Enter a **different** key (e.g., `WRONG-KEY`) into the plugin settings and verify.
3.  **Expected Result:** The verification should **fail** because the input key does not match the defined constant.

Issue #1049, opened 2025-09-05
Manage media Button link point to the folder


nitun:
On the GoDAM plugin, we have the `Manage Media` button under the `Media Library` section. 
That button should redirect the user to the site folder created inside the `https://app.godam.io/web/media-library/.....` 
For the example of the rtCamp.com site, it should go here: `https://app.godam.io/web/media-library?page=1&viewMode=grid&tab=Folder&folder=36asig06vc` 

<img width="1430" height="423" alt="Image" src="https://github.com/user-attachments/assets/db318b5e-f704-43ae-9a09-c4fc01c9c0f4" />

OpuRockey:
> API change for this is added in Central Production. [@OpuRockey](https://github.com/OpuRockey) [rtCamp/godam-core#464](https://github.com/rtCamp/godam-core/pull/464)
> 
> * Can we check folder deletion, API Key added without any media being added cases for this feature?

@kishan-gondaliya-7270 , 
The button will now automatically check for a site-specific folder.
If the corresponding folder exists, the button will redirect directly to that folder.
If no matching folder is found, it will fallback to the root folder as before.

OpuRockey:
> Though @OpuRockey @subodhr258 can you confirm how much time it take to get folder link ?
> Steps : I had site w/o license, after applying license key it had redirection link to root, after uploading one video and refreshing for 2-3 minutes manage button still dont have redirection to that folder (Folder : https://app.godam.io/web/media-library?tab=Folder&folder=kg9bpafchn&page=1&pageSize=20, site : https://148release.godam-146-multi.rt.gw/wp-admin/upload.php)

Getting site specific data from site API, we have used server cache. Right now, it is set to 24hours for performance. I have set it long as site folder name is not usually updated.

Is there any recommendation for cache time?

PR #1319, opened 2025-11-10
Manage media Button link point to the site folder

OpuRockey:
## Previously, the “Manage Media” button always redirected users to the root media folder, regardless of whether a folder existed for the specific site.

With this update:

1. The button will now automatically check for a site-specific folder.
2. If the corresponding folder exists, the button will redirect directly to that folder.
3. If no matching folder is found, it will fallback to the root folder as before.

🎯 Purpose of the Change

1. This enhancement improves usability by:
2. Reducing navigation steps for users
3. Taking users directly to the relevant media folder

Improving overall media organization and workflow efficiency

Issue: https://github.com/rtCamp/godam/issues/1049


subodhr258 (changes requested):

subodhr258:
Will merge this before QA, for 1.4.8

Issue #1269, opened 2025-10-17
iOS Safari and Chrome show vertical scroll behind GoDAM player in fullscreen mode


AmishRanpariya:
When the GoDAM player enters full-screen mode on iOS (both Safari and Chrome), a vertical scroll appears on the page. This allows the user to see the underlying page content below the full-screen player, making the UI look broken and visually inconsistent.

The issue is limited to iOS devices and does not occur on desktop browsers or Android devices.

**Steps to reproduce:**

1. Open [https://xtime.com/demo/schedule/](https://xtime.com/demo/schedule/) on an iOS device (Safari or Chrome).
2. Play the embedded GoDAM video. [It will automatically play in full-screen]
3. Observe the page—notice that vertical scroll is visible, revealing content behind the video player.

**Expected behaviour:**
When in full-screen mode, the player should completely cover the viewport, with no visible scroll or background content. The underlying page should be fixed and hidden until full-screen mode is exited.

**Actual behaviour:**
On iOS Safari and Chrome, a vertical scroll is visible in full-screen mode, allowing background content to show beneath the video player. This breaks the full-screen visual experience.

**Environment:**

* Affected OS: iOS
* Browsers: Safari, Chrome
* Unaffected platforms: macOS, Windows, Android
* GoDAM version: 1.4.4

**Impact:**
The full-screen experience looks broken and unpolished on iOS. Users can see and scroll to the content behind the video, diminishing immersion and potentially interfering with interaction.

**Possible direction:**

* Review full-screen handling on iOS and verify whether the player container is properly fixed and isolated.
* Ensure the page body is locked (e.g., `overflow: hidden;`) or fullscreen is managed via the native iOS video fullscreen API.
* Check if iOS Safari’s known fullscreen handling differences require a specific workaround in the GoDAM player implementation.

AmishRanpariya:
cc: @joelabreo227

hashibhere:
@subodhr258 we need to look into this issue and come up with a solution.

subodhr258:
Issue is fixed by hiding scroll overflow for fullscreen mode.

Before:
https://app.godam.io/web/video/gepjusqim7

After: (No scroll of background page, made sure that form layer scrolling on fullscreen video is not affected)
https://app.godam.io/web/video/k9hf07n7dr

@KMchaudhary please review the PR, we can add this fix to 1.4.5

kishan-gondaliya-7270:
[...]

PR #1339, opened 2025-11-13
GH1269 - iOS Safari and Chrome show vertical scroll behind GoDAM player in fullscreen mode

subodhr258:
Issue - https://github.com/rtCamp/godam/issues/1269

This pull request improves fullscreen behavior for the Godam video player, especially on iOS devices, by addressing viewport height issues and preventing unwanted background scrolling. The main changes include dynamically adjusting the fullscreen container's height using a custom CSS variable and managing body scrolling when entering or exiting fullscreen mode.

**Fullscreen and viewport improvements:**

* In `godam-player.scss`, the fullscreen modal's height now uses a dynamic calculation based on the `--vh` CSS variable, set by JavaScript, to ensure proper sizing on iOS devices. It falls back to `100vh` for other platforms.
* In `controlsManager.js`, a new method `setupIOSViewportHeight` is added to set and update the `--vh` variable based on the actual viewport height, handling window resize and orientation changes for accurate fullscreen sizing on iOS.

**Scrolling behavior management:**

* When entering fullscreen, the body's `overflow` style is set to `hidden` to prevent background scrolling, particularly on iOS.
* When exiting fullscreen, the body's `overflow` style is restored, allowing normal scrolling again. [[1]](diffhunk://#diff-b43145871d9e65106d5e84ff45c8dbc1f1aa2336cd8626ab1fb0b3a066cf0217R107-R117) [[2]](diffhunk://#diff-b43145871d9e65106d5e84ff45c8dbc1f1aa2336cd8626ab1fb0b3a066cf0217R154-R156)

### DIFF
--- a/assets/src/js/godam-player/managers/controlsManager.js
+++ b/assets/src/js/godam-player/managers/controlsManager.js
@@ -218,10 +218,18 @@ export default class ControlsManager {
 		}, { passive: true } );
 
 		// Handle orientation changes
-		window.addEventListener( 'orientationchange', () => {
-			// Give iOS a moment to recalc innerHeight
-			setTimeout( setViewportHeight, 200 );
-		} );
+		if ( screen.orientation ) {
+			screen.orientation.addEventListener( 'change', () => {
+				// Give iOS a moment to recalc innerHeight
+				setTimeout( setViewportHeight, 200 );
+			} );
+		} else {
+			// Fallback for browsers that don't support ScreenOrientation API
+			window.addEventListener( 'orientationchange', () => {
+				// Give iOS a moment to recalc innerHeight
+				setTimeout( setViewportHeight, 200 );
+			} );
+		}
 	}
 
 	/**


assets/src/js/godam-player/managers/controlsManager.js
```diff
@@ -181,9 +188,42 @@ export default class ControlsManager {
...
+		let resizeTimer = null;
+		window.addEventListener( 'resize', () => {
+			if ( resizeTimer ) {
+				clearTimeout( resizeTimer );
+			}
+			resizeTimer = setTimeout( setViewportHeight, 100 );
+		}, { passive: true } );
+
+		// Handle orientation changes
+		window.addEventListener( 'orientationchange', () => {
```
KMchaudhary: `orientationchange` event is deprecated and no longer recommended.
Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/orientationchange_event

You can use the **ScreenOrientation** API
e.g.
```js
screen.orientation.addEventListener("change", (event) => {
  const type = event.target.type;
  const angle = event.target.angle;
  console.log(`ScreenOrientation change: ${type}, ${angle} degrees.`);
});
```

subodhr258: Applied

Issue #1433, opened 2025-12-31
Verify the use of Video Compression Quality setting and Overwrite thumbnail setting


subodhr258:
1. We are sending a `video_quality` value (which can be 100,80,60,40 or 20)
2. Overwrite thumbnails option may not be required because we should always overwrite the thumbnails if the video is retranscoded.

<img width="1079" height="445" alt="Image" src="https://github.com/user-attachments/assets/ef0ee6c6-cbce-41b4-896d-8ea8839d6e13" />

PR #1444, opened 2026-01-01
fix: Overwrite thumbnails by default, persist selected custom thumbnail

subodhr258:
Issue - https://github.com/rtCamp/godam/issues/1433

This pull request refactors the logic for handling video thumbnail overwriting during media transcoding and retranscoding. The main change is the removal of the "overwrite thumbnails" setting and its related code, simplifying how and when thumbnails are updated. Now, the system only overwrites the primary thumbnail if it is not one of the custom-uploaded thumbnails, regardless of whether the media is being transcoded or retranscoded.

**Backend Logic Simplification:**

* Removed the `overwrite_thumbnails` setting and all related code, including its retrieval, storage, and usage in both PHP and JavaScript code. [[1]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868L126) [[2]](diffhunk://#diff-53b2ebd47ae78063d191af5a38b72dcee08982e0803759099aedbe794c85c3fcL40) [[3]](diffhunk://#diff-53b2ebd47ae78063d191af5a38b72dcee08982e0803759099aedbe794c85c3fcL298) [[4]](diffhunk://#diff-f29b4cd63502a994cef6926d1eab320b3b83465c4d16dd6920e3d7162bd61f27L470) [[5]](diffhunk://#diff-2d08689c450aa3bfb908c4d4323356f13b2a0354ba9264b65912a0d947e38b3bL21)
* Updated the logic in `add_media_thumbnails` and `transcoded_thumbnails_added` to only overwrite the primary thumbnail if it is not currently set to a custom-uploaded thumbnail, regardless of job type. [[1]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868L630-L642) [[2]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868R637-R642) [[3]](diffhunk://#diff-0d4b5d9fe699dc8791e18e5ba3ea3c125885c17e6c1ff6815ee718e53481e7afL630-L642)
* Removed the `rtgodam_is_override_thumbnail` helper function, as it is no longer needed.

**Frontend/UI Cleanup:**

* Removed the "Overwrite video thumbnails" toggle from the Video Thumbnails settings UI and all related Redux state and selectors. [[1]](diffhunk://#diff-edb269501c981c15d822f92d6acc4741cf45f7a56ee4ae46d9a670b864ca5c00L9-L14) [[2]](diffhunk://#diff-edb269501c981c15d822f92d6acc4741cf45f7a56ee4ae46d9a670b864ca5c00L44-L54) [[3]](diffhunk://#diff-2d08689c450aa3bfb908c4d4323356f13b2a0354ba9264b65912a0d947e38b3bL21)

**Documentation and Comments:**

* Updated method documentation to reflect the new thumbnail overwrite behavior and removed references to the old setting.

These changes make the thumbnail overwrite behavior more predictable and reduce complexity for both users and developers.

#### Before Transcoding

https://github.com/user-attachments/assets/f11e2e84-a663-4ba1-b8a1-ed916801ca08

#### After Transcoding

https://github.com/user-attachments/assets/4f8d83c5-ce3e-4010-bee2-b9f4552a8c5b


admin/godam-transcoder-functions.php
```diff
@@ -195,36 +195,6 @@ function rtgodam_get_upload_dir() {
...
-
-	/**
-	 * Allow user to override the setting.
-	 *
-	 * @since 1.1.0
-	 *
-	 * @param boolean   $rtgodam_override_thumbnail     Number of thumbnails set in setting.
-	 * @param int       $attachment_id              ID of attachment.
-	 */
-	$rtgodam_override_thumbnail = apply_filters( 'rtgodam_is_override_thumbnail', $rtgodam_override_thumbnail, $attachment_id );
```
shreyasikhar: What if some users are using this filter? It will not work for them after this change.

subodhr258: This filter is applied inside the `rtgodam_is_override_thumbnail()` function, which we are no longer using. And so the filter will not be useful.
If the user retranscodes the video then we are enforcing thumbnail overriding by default.

Issue #1275, opened 2025-10-20
SureForms - Audio Recording for GoDAM Recorder


subodhr258:
Currently, the GoDAM recorder field supports only Webcam and Screen Recording.
Users have requested the ability to record Audio-only (e.g., voice notes, narration, podcasts).
Since Uppy already provides an @uppy/audio plugin, we can extend the existing recorder integration to support audio recording.

Follow the integration done with Gravity Forms

kishan-gondaliya-7270:
@subodhr258 Please check preview submission as well for SureForms 

Also can we make   GoDAM Record button consistence for All Forms 
currently in gravity form : "Start Recording",   Sure Form : "Record Video"]

Also if we can add separate Folder for all Forms the way it is for gravity form #1335 ?

Issue #1194, opened 2025-09-30
Ninja Forms


nayemDevs:

nayemDevs:
@umeshnevase - If you have started on this, please change the task status and share ETA.

umeshnevase:
@nayemDevs Yes, I've started working on this and its almost complete.
I'll raise the PR after testing it properly.

umeshnevase:
@nayemDevs PR is created and tested on my test site.

PR #1451, opened 2026-01-05
Add audio recording support for SureForms GoDAM Recorder

AhmarZaidi:
Issue: #1275  

Implements audio recording support for SureForms GoDAM Recorder field.

- Import Uppy audio CSS and add audio player styling
- Detect audio files and set correct transcoding job type
- Add audio file selector option in field settings
- Update button text from "Record Video" to "Start Recording"
- Handle `.webm` audio files using name-based detection
- Add `is_audio` check fucntion to `custom-functions`

Note: Referenced #1194 & #1274 for consistency.

## Demo

https://github.com/user-attachments/assets/c76f5f33-f731-419b-aa9b-9745130f1394

### DIFF
--- a/inc/helpers/custom-functions.php
+++ b/inc/helpers/custom-functions.php
@@ -427,6 +427,10 @@ function godam_is_audio_file_by_name( $filename ) {
  * @return bool True if the file is an audio file, false otherwise.
  */
 function godam_is_audio_file( $file_path_or_url ) {
+	if ( empty( $file_path_or_url ) || ! is_string( $file_path_or_url ) ) {
+		return false;
+	}
+
 	$file_type = wp_check_filetype( $file_path_or_url );
 	$mime_type = ! empty( $file_type['type'] ) ? $file_type['type'] : '';
 
@@ -437,7 +441,7 @@ function godam_is_audio_file( $file_path_or_url ) {
 
 	// Container formats that can hold both audio and video.
 	// These require filename-based detection to distinguish audio from video.
-	$ambiguous_extensions = array( 'webm', 'mp4' );
+	$container_formats = array( 'webm', 'mp4' );
 
 	// Handle ambiguous container formats that might be audio.
 	// Browser-specific behavior:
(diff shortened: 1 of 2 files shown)


inc/helpers/custom-functions.php
```diff
@@ -417,6 +417,32 @@ function godam_is_audio_file_by_name( $filename ) {
...
+ * @param string $file The file path or URL to check.
+ *
+ * @return bool True if the file is an audio file, false otherwise.
+ */
+function godam_is_audio_file( $file ) {
+	$file_type = wp_check_filetype( $file );
+	$mime_type = ! empty( $file_type['type'] ) ? $file_type['type'] : '';
+
+	// Check if the MIME type indicates an audio file.
+	if ( ! empty( $mime_type ) && strpos( $mime_type, 'audio' ) !== false ) {
```
mi5t4n:
```suggestion
	if ( ! empty( $mime_type ) && str_starts_with( $mime_type, 'audio/' ) ) {
```

The previous implementation checks for the presence of "audio" anywhere in the string, which would return true even if it appears at the end. A better approach is to use str_starts_with(), which checks if the string begins with a specific value.

This function was introduced in PHP 8.0. However, WordPress Core provides a polyfill for it starting from version 5.9.0. Since our WordPress requirement is 6.5+, we can safely use str_starts_with().

Reference: WordPress Core Polyfill for [str_starts_with](https://github.dev/WordPress/wordpress-develop/blob/35313faf1e455d8e63c7711048215b9f786b4a0b/src/wp-includes/compat.php#L427-L447)

inc/helpers/custom-functions.php
```diff
@@ -417,6 +417,41 @@ function godam_is_audio_file_by_name( $filename ) {
...
+	$mime_type = ! empty( $file_type['type'] ) ? $file_type['type'] : '';
+
+	// Check if the MIME type indicates an audio file.
+	if ( ! empty( $mime_type ) && str_starts_with( $mime_type, 'audio/' ) ) {
+		return true;
+	}
+
+	// Container formats that can hold both audio and video.
+	// These require filename-based detection to distinguish audio from video.
+	$ambiguous_extensions = array( 'webm', 'mp4' );
```
mi5t4n:
```suggestion
	$container_formats = array( 'webm', 'mp4' );
```

Rather than ambiguous words, container format may be better wording for it.

mi5t4n (commented):
Just a nitpick comment, rest all looks good.

mi5t4n (approved):
LGTM

PR #1454, opened 2026-01-06
Add audio recording support for Everest Forms GoDAM Recorder

AhmarZaidi:
Issue: #1276

Implements audio recording support for Everest Forms GoDAM Recorder field.

- Import Uppy audio CSS and add audio player styling
- Detect audio files and set correct transcoding job type
- Add audio file selector option in field settings
- Update button text from "Record Video" to "Start Recording"
- Handle .webm audio files using name-based detection
- Add is_audio check fucntion to custom-functions

Note: Referenced #1194, #1274 & #1275 for consistency.

## Demo

https://github.com/user-attachments/assets/d2ff3f59-76b0-4cee-ba50-0d49273c0274

### DIFF
--- a/inc/helpers/custom-functions.php
+++ b/inc/helpers/custom-functions.php
@@ -427,6 +427,10 @@ function godam_is_audio_file_by_name( $filename ) {
  * @return bool True if the file is an audio file, false otherwise.
  */
 function godam_is_audio_file( $file_path_or_url ) {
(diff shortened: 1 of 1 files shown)


mi5t4n:
@AhmarZaidi I noticed a few issues here that are similar to those addressed in https://github.com/rtCamp/godam/pull/1451. Could you please address those here as well?

PR #1460, opened 2026-01-07
Fix: aspect ratio in godam/video block

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam/issues/1461

This pull request improves how aspect ratios are handled for the Godam video player block, ensuring that aspect ratios are only set when valid and allowing for dynamic calculation based on video dimensions. The changes enhance reliability and flexibility in aspect ratio management both in the editor and at runtime.

**Aspect Ratio Handling Improvements:**

* Added logic in `VideoJS.js` to only set the player's aspect ratio if the `aspectRatio` option is in a valid `x:y` format, preventing invalid values from being applied.
* In `edit.js`, implemented a `useMemo` hook to calculate the aspect ratio dynamically from `videoWidth` and `videoHeight` when `aspectRatio` is set to `'responsive'`, or to validate an explicit `aspectRatio` string.
* Updated the construction of `videoOptions` in `edit.js` to only include the `aspectRatio` property if it is valid, removing the previous hardcoded default of `'16:9'`. [[1]](diffhunk://#diff-7f3eed49285d53dd5ca33b2e1543d8964dcce2aa03553b162df3c1a9068dfa66L153) [[2]](diffhunk://#diff-7f3eed49285d53dd5ca33b2e1543d8964dcce2aa03553b162df3c1a9068dfa66L163-R187)

**Attribute Management:**

* Passed new attributes (`aspectRatio`, `videoWidth`, `videoHeight`) into the `VideoEdit` component to support dynamic aspect ratio calculations.

## Screen recording
https://app.godam.io/web/video/smn7vpgtma

### DIFF
--- a/assets/src/blocks/godam-player/VideoJS.js
+++ b/assets/src/blocks/godam-player/VideoJS.js
@@ -57,6 +57,13 @@ export const VideoJS = ( props ) => {
 			// Verify if aspectRatio is in valid format x:y
 			if ( /^\d+:\d+$/.test( options.aspectRatio ) ) {
 				player.aspectRatio( options.aspectRatio );
+
+				// Get x and y from aspectRatio
+				const [ x, y ] = options.aspectRatio.split( ':' );
+				if ( playerRef.current && x && y ) {
+					const playerEl = playerRef.current.el_;
+					playerEl.style.paddingTop = `${ ( y / x ) * 100 }%`;
+				}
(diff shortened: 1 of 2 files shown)


subodhr258:
Issue is resolved, LGTM as well

subodhr258 (approved):

Issue #1080, opened 2025-09-11
Vimeo Thumbnails: Custom & Existing Thumbnail Selection Issues


IndiraBiswas:
**Description:**

There are two related issues with Vimeo thumbnails:

~~1. Custom Thumbnails Not Working~~

- Uploading/setting custom thumbnails is currently not functional.
- Selecting from existing thumbnails is also not working.

2. Disabled Custom Thumbnail Option

- When the custom thumbnail option is disabled, the UI should provide clarity.
- A tooltip should be displayed explaining why the option is disabled (e.g., maximum of 3 thumbnails allowed).

PR #1471, opened 2026-01-09
Fix custom & existing thumbnail tooltip

AhmarZaidi:
Issue: #1080
Part 2: Disabled Custom Thumbnail Option

This PR adds a tooltip similar to copy block's tooltip in video editor when the custom thumbnail option is disabled. It also ensures that the tooltip is rendered as soon as the user hover's over the add button.

## Demo

https://github.com/user-attachments/assets/ead9cb32-553a-43f7-bdfc-34039641dcd0


shreyasikhar (approved):
LGTM except one comment

Issue #1550, opened 2026-01-20
Unexpected Uncaught TypeError on Two column screen


OpuRockey:
## Bug Report

It displays an Uncaught TypeError: Cannot read properties of null (reading 'length') at MediaElementPlayer error while virtual media modal pops up at two column screen

<img width="1028" height="131" alt="Image" src="https://github.com/user-attachments/assets/ba658469-d805-4adf-b669-5454544e97ad" />

subodhr258:
@kishan-gondaliya-7270  The PR is merged and ready for QA.

### Testing instructions
1. Open attachment modal for virtual media
2. Check the console for the reported error, it should not appear, with no side-effects

KMchaudhary:
Outdated issue. Not reproducible.

PR #1552, opened 2026-01-20
Fixed console error on two column modal for virtual media

OpuRockey:
# 🛠️ Fix: Prevent Uncaught TypeError on Two-Column Modal for Virtual Media

## Issue

Fixes the console error reported in Issue #1550 where the two-column screen throws an Uncaught TypeError: Cannot read properties of null (reading ‘length’) when opening the Virtual Media modal.

## What’s Fixed

This PR introduces a guard to ensure that before attempting to read .length on the related element (or collection), the value is verified to be non-null. This prevents the JavaScript TypeError from occurring when the modal attempts to access a DOM node or data that isn’t present.

## Testing

- Steps to verify the fix:
- Navigate to the two-column screen in the application.
- Trigger opening the Virtual Media modal.
- Confirm that the modal opens without any Uncaught TypeError in the console.
- Validate expected functionality of the modal behavior post-fix.

## Issue

https://github.com/rtCamp/godam/issues/1550

### DIFF
--- a/assets/src/js/media-library/views/attachment-detail-two-column.js
+++ b/assets/src/js/media-library/views/attachment-detail-two-column.js
@@ -604,7 +604,7 @@ export default AttachmentDetailsTwoColumn?.extend( {
 			// If the attachment is virtual (e.g. a GoDAM proxy video), override default preview.
 			if ( undefined !== virtual && virtual ) {
 				const videoUrl = this.model.get( 'transcoded_url' ); // Ensure it's a valid .mp4
-				const $container = this.$el.find( '.wp-video' );
+				const $container = this.$el.find( '.wp-video > div' );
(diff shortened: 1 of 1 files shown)


OpuRockey:
@AhmarZaidi , I couldn't replicate this at my end. Could you please give it a try here?
https://rakib.rt.gw/wp-login.php
u- rtcamp
p - #goodwork#

OpuRockey:
@AhmarZaidi , Could you please give it a try on incognito?

KMchaudhary:
@OpuRockey @AhmarZaidi , in such rare edge cases where we frequentl opening/closing the modal and we got console error is okay. We can skip those types of errors before investing more time on unfixable and rarely occurring console errors.

AhmarZaidi (approved):

Issue #1097, opened 2025-09-12
Transcoding Job Failure Handling – Show “Failed” in Plugin and Support Retry


IndiraBiswas:
**Description:**

When a transcoding job fails (e.g., due to HTTP auth errors), the plugin and Central are not handling the failure state consistently:

**Plugin:**

- Displays “Not Started” instead of “Failed”.
- No clear way to retry from the plugin UI.

**Central:**

- Shows a “Retry Transcoding” option, but when retry is clicked, no progress bar is visible (progress is only visible in Vishal’s account, not others).
- Even after retry succeeds, the previous error message remains attached to the job, creating confusion.

**Note:**
Sometimes a video doesn’t receive the correct status for re-transcoding, which can result in the loss of likes and comments. However, when the same video was re-transcoded after updating the thumbnail settings, it correctly showed the “Transcoded” status, and the likes and comments were retained.

PR #1572, opened 2026-01-22
Fix transcoding job failure inconsistant state due to HTTP auth

AhmarZaidi:
Issue: #1097

Plugin makes API calls to a Central (backend) that fails when HTTP authentication is enabled on the WordPress site. Server-side detection methods (checking PHP variables, making REST requests) are unreliable because they run within the authenticated context. Media uploads would fail silently with incorrect error messages since the HTTP auth status was only checked on page load.

This PR adds client-side HTTP authentication detection using a browser-based fetch request with `credentials: 'omit'` and `cache: 'no-store'`. This detects HTTP auth by checking for 401 responses which is accuarte.

The detection runs:
1. On initial page load to update the stored status
2. Before each media upload via Plupload's `BeforeUpload` event
3. The result is saved to the database via AJAX for PHP functions to reference

### Uploads Page 

1. Disable HTTP auth and test if uploads functionality is working correctly
2. Enable HTTP auth (using easy engine configuration)
3. Navigate to Media > Library (upload page)
4. Upload a video file using either: "Add New" button or Drag and drop into the upload area
5. On the media file we see a 'Transcoding Failed' status on hovering
6. When page is reloaded we see a admin notice with title 'GoDAM Transcoding Blocked'
7. Disable HTTP auth (using easy engine configuration)
8. Again check that the uploads functionality is working for new files & notice disappears on reload
9. The previously failed media show 'Not Transcoded' status on hovering and can be re-transcoded by bulk selecting and clicking on 'Transcode Media'

### WP Media Modal (on a post, page, etc.)

1. Disable HTTP auth and test if uploads functionality is working correctly
2. Enable HTTP auth (using easy engine configuration)
3. Add the 'GoDAM Video' block and click on 'Select Video'
4. Go to 'Upload Files' Tab in the Media Modal and select a file to upload
5. After upload, on the media file we see a 'Transcoding Failed' status on hovering
6. Disable HTTP auth (using easy engine configuration)
9. Now, on the previously failed media files we see a 'Not Transcoded' status on hovering

## Demo

https://github.com/user-attachments/assets/5f3236ed-d987-420f-8cae-8883a55e19a8

### DIFF
--- a/assets/src/js/http-auth-detector.js
+++ b/assets/src/js/http-auth-detector.js
@@ -95,6 +95,11 @@
 			// Before files are uploaded, check for HTTP auth.
 			if ( this.uploader ) {
 				this.uploader.bind( 'BeforeUpload', function( up, file ) {
+					// If this file has already been checked, allow upload to proceed.
+					if ( file.httpAuthChecked ) {
+						return;
+					}
+
 					// If detection is already in progress, pause uploads and wait.
 					if ( up.httpAuthDetectionInProgress ) {
 						file.status = window?.plupload?.STOPPED;
@@ -108,6 +113,8 @@
 					// Run detection and resume uploads when done.
 					$.when( detectHttpAuth() ).always( function() {
 						up.httpAuthDetectionInProgress = false;
+						// Mark this file as checked so we don't run detection again for it.
+						file.httpAuthChecked = true;
 						up.start();
 					} );
 				} );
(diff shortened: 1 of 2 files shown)


OpuRockey (approved):
LGTM. Please fix the conflict.

AhmarZaidi:
@OpuRockey Added HTTP auth detection for image files here: https://github.com/rtCamp/godam/pull/1572/changes/1b14e7b5e376c91de4741cdd765ccd56042330ca

PR #1585, opened 2026-01-29
Ensure Card Size Compatibility in GoDAM Product Gallery Layout [WooCommerce]

rudrakshi-gupta:
## PR Overview

- Add 2 new Card Size settings (Tablet & Mobile) for Product Gallery in carousel
- Replace rem units with vw where applicable
- Add 2 new settings for grid columns based on device type
- Fix Issue with Resposible grid videos based on device with CSS
- Ensure images and videos scale proportionally in carousel and grid

Refs rtCamp/godam-core#732


integrations/woocommerce/blocks/godam-product-gallery/edit.js
```diff
@@ -72,6 +76,10 @@ export default function Edit( { attributes, setAttributes } ) {
 
 	const blockProps = useBlockProps();
 
+	const deviceType = useSelect( ( select ) => {
+		return select( 'core/edit-post' )?.__experimentalGetPreviewDeviceType?.();
```
sabbir1991: Why we are using experimental function here? Instead of we can use native js for detecting device. Like window navigator user agent approch or we can detect using inner width.

integrations/woocommerce/blocks/godam-product-gallery/edit.js
```diff
@@ -106,11 +114,13 @@ export default function Edit( { attributes, setAttributes } ) {
 
 	// Update card width on view change.
 	useEffect( () => {
-		const defaultWidth = parseFloat( getCardWidthForView( view, layout ) );
+		const defaultWidth = parseFloat( getCardWidthForView( view, layout, 'desktop' ) );
 
 		// If view has changed, reset to default width, else do nothing.
 		if ( previousViewRef.current !== view ) {
-			setAttributes( { cardWidth: defaultWidth } );
+			setAttributes( { desktopCardWidth: defaultWidth } );
```
sabbir1991: Instead of multiple attribute can we keep only one object with different device width. It will easy to handle?

integrations/woocommerce/blocks/godam-product-gallery/render.php
```diff
@@ -109,16 +112,19 @@ function godam_sanitize_rgba_color( $color ) {
 	'unmute_button_enabled'       => filter_var( $attributes['unmuteButtonEnabled'], FILTER_VALIDATE_BOOLEAN ),
 	'unmute_button_bg_color'      => godam_sanitize_rgba_color( $attributes['unmuteButtonBgColor'] ),
 	'unmute_button_icon_color'    => sanitize_hex_color( $attributes['unmuteButtonIconColor'] ),
-	'card_width'                  => intval( $attributes['cardWidth'] ),
+	'dektop_card_width'           => intval( $attributes['desktopCardWidth'] ),
```
sabbir1991: Do you think the value always return int val or it can be a float value? If it also return the float value then floatval function need to use

rudrakshi-gupta: Sure. I'll use floatval.

integrations/woocommerce/blocks/godam-product-gallery/edit.js
rudrakshi-gupta: Changed to a more stable API - `select( 'core/editor' ).getDeviceType();`

integrations/woocommerce/blocks/godam-product-gallery/render.php
```diff
@@ -109,16 +112,19 @@ function godam_sanitize_rgba_color( $color ) {
...
 	'arrow_border_radius'         => intval( $attributes['arrowBorderRadius'] ),
 	'arrow_visibility'            => sanitize_text_field( $attributes['arrowVisibility'] ),
-	'grid_columns'                => intval( $attributes['gridColumns'] ),
+	'grid_columns_desktop'        => intval( $attributes['gridColumnsDesktop'] ),
+	'grid_columns_tablet'         => intval( $attributes['gridColumnsTablet'] ),
+	'grid_columns_mobile'         => intval( $attributes['gridColumnsMobile'] ),
 	'grid_row_gap'                => intval( $attributes['gridRowGap'] ),
 	'grid_column_gap'             => intval( $attributes['gridColumnGap'] ),
-	'grid_card_alignment'         => sanitize_text_field( $attributes['gridCardAlignment'] ),
 	'cta_enabled'                 => filter_var( $attributes['ctaEnabled'], FILTER_VALIDATE_BOOLEAN ),
```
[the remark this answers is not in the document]
rudrakshi-gupta: Done.

PR #1601, opened 2026-02-04
Feat: GoDAM Product Gallery specific skin

rudrakshi-gupta:
## In the PR

Figma refrence - [GoDAM reel Player](https://www.figma.com/design/qewE6AYuMQoSFnVPwjghKx/GoDAM?node-id=8650-58199&t=9rEXKvmFEecWMYmM-0)

- Add a new SCSS file:godam-reels-skin for the Gallery
- Style out every control and main play button
- Add new godam-context attribute for product gallery and skin initalization

### DIFF
--- a/inc/classes/rest-api/class-dynamic-shortcode.php
+++ b/inc/classes/rest-api/class-dynamic-shortcode.php
@@ -128,7 +128,7 @@ class Dynamic_Shortcode extends Base {
 			$shortcode .= " godam_context='{$godam_context}'";
 		}
 
-		$shortcode .= "]";
+		$shortcode .= ']';
 
 		// Add filter for shortcode.
 		$shortcode = apply_filters( 'rtgodam_shortcode_output', $shortcode, $id );


inc/classes/rest-api/class-dynamic-shortcode.php
```diff
@@ -120,7 +122,7 @@ public function render_shortcode( WP_REST_Request $request ) {
 		$video_date  = apply_filters( 'rtgodam_shortcode_video_date', $video_date, $id );
 
 		ob_start();
-		$shortcode = "[godam_video id='{$id}' engagements='{$engagements}' sources='{$sources_with_placeholders}']";
+		$shortcode = "[godam_video id='{$id}' engagements='{$engagements}' sources='{$sources_with_placeholders}' godam_context='{$godam_context}']";
```
sabbir1991: what about if `$godam_context` is return null, do you think we need to print that attribute in shortcode?

Issue #1239, opened 2025-10-09
WordPress Permission bug fixes and tweak


joelabreo227:
> Needs discussion : 
> 
> 1. Bug : Infinite loading for GoDAM Video for Contributor role (Handle it the way it is handled by WP for gallery block), Also able to click on New Folder and create, gives undefined error
> 2. Bug : Access to GoDAM Gallery --> Access to engagement settings for contributor
> 3. ~Editor should have access to settings for configuring skins and thumbnails, branding related things~
> 4. If Author has access to single video analytics for own and others videos then why access to Dashboard needs to be restricted? (**or restrict access to other videos analytics**)
> 
> <img width="1907" height="951" alt="Image" src="https://github.com/user-attachments/assets/34a806da-f0bb-4971-96ca-317ab3b7c084" /> 

 _Originally posted by @kishan-gondaliya-7270 in [#1027](https://github.com/rtCamp/godam/issues/1027#issuecomment-3384236691)_

OpuRockey:
> Bug : Access to GoDAM Gallery --> Access to engagement settings for contributor

Can you please add some more details?

OpuRockey:
> If Author has access to single video analytics for own and others videos then why access to Dashboard needs to be restricted? (or restrict access to other videos analytics)

What will be the suggested solution here?

KMchaudhary:
> If Author has access to single video analytics for own and others videos then why access to Dashboard needs to be restricted? (or restrict access to other videos analytics)
> 
> What will be the suggested solution here?

@OpuRockey allow the `Author` role to access the Dashboard page.

subodhr258:
@kishan-gondaliya-7270 the PR is merged and ready for QA.
@OpuRockey please add testing instructions for the PR, thanks

rudrakshi-gupta:
This is done.

PR #1639, opened 2026-02-12
Fix WordPress permission issues and UX tweaks for GoDAM (Issue #1239)

OpuRockey:
## Description

This PR addresses multiple WordPress permission and UX issues reported in [#1239](https://github.com/rtCamp/godam/issues/1239).

1. Fixes an issue where users with the Contributor role experienced infinite loading when using the GoDAM video functionality.
2. Only loads GoDAM data when the user has sufficient capabilities.
3. Restricts access to GoDAM related blocks from editor for Contributors and below roles.
4. Authors are allowed to view analytics.

## Testing Instructions

1. Log in as a Contributor.
2. GoDAM related blocks from editor are not accessible.

### DIFF
--- a/inc/classes/wpbakery-elements/class-wpb-godam-audio.php
+++ b/inc/classes/wpbakery-elements/class-wpb-godam-audio.php
@@ -13,7 +13,7 @@ defined( 'ABSPATH' ) || exit;
 
 /**
  * Class WPB_GoDAM_Audio
- * 
+ *
  * @since 1.6.0
  *
  * @package GoDAM
@@ -23,7 +23,7 @@ class WPB_GoDAM_Audio {
 
 	/**
 	 * WPB_GoDAM_Audio constructor.
-	 * 
+	 *
 	 * @since 1.6.0
 	 */
 	protected function __construct() {
@@ -40,7 +40,7 @@ class WPB_GoDAM_Audio {
 
 	/**
 	 * Setup hooks.
-	 * 
+	 *
 	 * @since 1.6.0
 	 *
 	 * @return void
@@ -51,13 +51,17 @@ class WPB_GoDAM_Audio {
 
 	/**
 	 * Map audio element to WPBakery.
-	 * 
+	 *
 	 * @since 1.6.0
 	 *
 	 * @return void
 	 */
 	public function godam_audio() {
 		if ( ! function_exists( 'vc_map' ) ) {
+			return;
+		}
+
+		if ( is_admin() && ! current_user_can( 'publish_posts' ) ) {
 			return;
 		}
 
@@ -146,7 +150,7 @@ class WPB_GoDAM_Audio {
 							'not_empty' => true,
 						),
 					),
-					
+
 					// Caption Settings.
 					array(
 						'type'        => 'textfield',
--- a/inc/helpers/custom-functions.php
+++ b/inc/helpers/custom-functions.php
@@ -1129,28 +1129,3 @@ function rtgodam_convert_to_https_url( $urls ) {
 
 	return set_url_scheme( $urls, 'https' );
 }
-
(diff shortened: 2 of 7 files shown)


inc/helpers/custom-functions.php
```diff
@@ -1129,3 +1129,39 @@ function ( $url ) {
...
+		'upload_files',
+		'manage_options',
+		'publish_posts',
+	);
+
+	if ( ! in_array( $permission_type, $allowed_caps, true ) ) {
+		$permission_type = 'publish_posts';
+	}
+
+	return current_user_can( $permission_type );
```
rtBot: :warning: **Warning**: Couldn&#039;t determine the value passed to the $capability parameter in function call to current_user_can(). Please check if it matches a valid capability. Found: $permission_type (*WordPress.WP.Capabilities.Undetermined*).

inc/classes/wpbakery-elements/class-wpb-godam-audio.php
```diff
@@ -61,6 +61,10 @@ public function godam_audio() {
 			return;
 		}
 
+		if ( is_admin() && ! current_user_can( 'publish_posts' ) ) {
```
AhmarZaidi: Since the same code is being repated in 6 places, can we move it to `custom-functions` and use it everywhere?

inc/classes/class-pages.php
```diff
@@ -180,7 +180,7 @@ public function add_admin_pages() {
 			$this->menu_slug,
 			__( 'Dashboard', 'godam' ),
 			__( 'Dashboard', 'godam' ),
-			'edit_pages', // Dashboard is accessible to editors and above.
+			'publish_posts', // Dashboard is accessible to authors and above.
```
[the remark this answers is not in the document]
AhmarZaidi: > The capability 'publish_posts' is used for the Dashboard submenu, while other Author-accessible submenus (Video Editor at line 193 and Analytics at line 203) use 'upload_files'. Both are Author-level capabilities in default WordPress, so this works correctly for standard role configurations. However, in custom setups where these capabilities might be granted independently, this inconsistency could lead to unexpected menu visibility. For better maintainability and consistency, consider using the same capability for all Author-accessible menu items.

@OpuRockey This comment seems reasonable.

For `authors and above` capability check we are using `upload_files` elsewhere so here also we should use the same.

Unless there's a reason not to?

AhmarZaidi (commented):
2 minor changes, rest look good.

- https://github.com/rtCamp/godam/pull/1639#discussion_r2803140477
- https://github.com/rtCamp/godam/pull/1639#discussion_r2803206122

inc/classes/class-pages.php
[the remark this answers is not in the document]
OpuRockey: In line 161, GoDAM menu is restricted below author role. I actually use the same condition to make it consistent.

Image

inc/classes/wpbakery-elements/class-wpb-godam-audio.php
OpuRockey: I have mentioned it earlier here.
https://github.com/rtCamp/godam/pull/1639#discussion_r2802313804

OpuRockey: I think it would be better to keep it as it is.

AhmarZaidi (changes requested):
Will approve again after testing Elementor and WPBakery

cc: @OpuRockey

AhmarZaidi (approved):
Tested Elementor and WPBakery:
- Elementor: Correctly hides GoDAM Widgets for contributors and below roles
- WPBakery: Doesn't show wpbakery editor for contributors and below roles

LGTM

PR #1659, opened 2026-02-20
Added play button gloabl settings and rendered on frontend

sabbir1991:
A quick settings for changing the play button globally for all product videos. Here is the related ticket https://github.com/rtCamp/godam-core/issues/830


integrations/woocommerce/classes/class-wc-utility.php
```diff
@@ -339,6 +339,17 @@ public static function get_woocommerce_video_modal_css_variables() {
 		$settings = get_option( 'rtgodam-settings', array() );
 		$woo      = $settings['integrations']['woocommerce'] ?? array();
 
+		$woo = wp_parse_args(
```
rudrakshi-gupta: You don't need to separately parse the variables.

> $woo      = $settings['integrations']['woocommerce'] ?? array();

This above line already calls all the woocommerce integration settings as an array.

PR #1663, opened 2026-02-23
Fix: GoDAM WooCommerce Layer UI

rudrakshi-gupta:
## In the PR

- Remove shop button and add icons for Add to Cart button
- Refactor conditions for showing icons in editor and frontend
- Fix UI CSS for Product Hotspot Box

Refs rtCamp/godam-core#735

### DIFF
--- a/assets/src/css/godam-player.scss
+++ b/assets/src/css/godam-player.scss
@@ -601,6 +601,7 @@
 	transition: opacity 0.2s ease, visibility 0.2s ease;
 	word-wrap: break-word;
 	min-width: 20em;
+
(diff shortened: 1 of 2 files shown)


assets/src/css/godam-player.scss
```diff
@@ -609,7 +611,6 @@
 		border-style: solid;
 		border-color: rgba(255, 255, 255) transparent transparent transparent;
 	}
-
 	a {
```
sabbir1991: add more specific selector. like a.example-class
it helps to override the default anchor css if have any

PR #1702, opened 2026-03-06
Split WooCommerce features into free and premium

rudrakshi-gupta:
## In the PR

- Add PRO tags on premium features
- Disable all save and video section buttoons from settings when API is invalid
- Add Notices for Upgrading plans for users on PRO features
- Configure CSS for premium features in Product edit page
- Add dummy video cards for Product reels
- Disable JS clicks and save_post hooks if key is not valid
- Stop the rendering of premium features on frontend when key is not valid

## Issue

Refs Refs rtCamp/godam-core#900

## Demo

https://github.com/user-attachments/assets/2d3813d8-369c-41ab-9a2c-25ee48dc95b6

### DIFF
--- a/assets/src/css/admin.scss
+++ b/assets/src/css/admin.scss
@@ -943,3 +943,11 @@ to ensure proper responsiveness and avoid layout issues.
 	font-size: 12px;
 	background: rgb(243, 245, 247);
 }
+
+.text-\[\#AB3A6C\] {
+	color: #AB3A6C;
+}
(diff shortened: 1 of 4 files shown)


KMchaudhary:
Screenshot 2026-03-09 at 10 40 46 AM

Here instead of **Dummy video 1**, **Dummy video 2** can we replace the text with **Sample video 1**, **Sample video 2**

Issue #1532, opened 2026-01-19
Fix UI Issue with Forms outflowing from  Video (In gallery block, in mobile view), comment outflowing from Engagement box in mobile view


kishan-gondaliya-7270:
Step 👍🏼: 
- Add forms layerin video and view in GoDAM Gallery block 
- Observe same in mobile view
- Add long comment and observe comment box in mobile

<img width="1407" height="872" alt="Image" src="https://github.com/user-attachments/assets/c8b7a1ba-bac7-4803-9dc1-78bf115f3583" />
<img width="512" height="900" alt="Image" src="https://github.com/user-attachments/assets/19712c15-c01c-43a4-8290-3822228589c9" />

ref : https://github.com/rtCamp/godam/issues/1459#issuecomment-3760412277

KMchaudhary:
Forms UI issue is fixed, but custom HTML CTA layer's overflow is not yet fixed

<img width="1470" height="734" alt="Image" src="https://github.com/user-attachments/assets/e2318a8b-58c7-41d0-ac2b-f3842acd24fe" />

<img width="1470" height="734" alt="Image" src="https://github.com/user-attachments/assets/dc8c54fd-15d3-45c3-a235-1f03fa48e6b8" />

kishan-gondaliya-7270:
Released

PR #1721, opened 2026-03-11
Fix: prevent comment overflow.

Asteriskkkk:
Fixes: #1532

Wrap long comment text in the engagement panel to stop UI overflow.

| Before | After |
| ------ | ------ |
| image-before| image-after

| Before | After |
| ------ | ------ |
| image-before| image-after


Asteriskkkk:
@subodhr258  please review this when you get a chance. It’s a small fix. also needed feedback on above comment thanks!

PR #1733, opened 2026-03-16
fix: ensure WooCommerce cart store availability before adding items to cart

KMchaudhary:
**Fixes:** 
> Bug : In GoDAM Gallery Block, Video with Woo Layer has failure in add-to-cart activity. 

Ref: https://docs.google.com/document/d/1KDpp3WxD6g8c0_J-p9SR5PzO148xZbgF3V97gyMmWOs/edit?tab=t.y19cn2tdynpw

---

This pull request focuses on improving the integration between the Godam Player and WooCommerce, specifically ensuring reliable access to the WooCommerce cart store and handling cases where it may not be available. The changes also add necessary dependencies to support this functionality.

Enhancements to WooCommerce integration:

* Added a runtime check in `WooCommerceLayerManager` to verify that the WooCommerce cart store (`wc/store/cart`) and its `addItemToCart` method are available before attempting to add items. If unavailable, a user-friendly error message is shown and the UI is reset.
* Updated script registration in `class-godam-player.php` to include the `wc-blocks-data-store` dependency when WooCommerce is active, ensuring the cart store is loaded and available for frontend scripts.

Dependency updates:

* Added `@wordpress/data` as a project dependency in `package.json` to support interaction with WordPress data stores, such as the WooCommerce cart store.

## Demo

https://github.com/user-attachments/assets/8e629e73-e06e-4d27-9af4-418cf6c6e606


rudrakshi-gupta (approved):

subodhr258:
Found 2 issues with the Woo layer in Gallery currently:
1. When I clicked on the + icon without the Mini Cart, it opened in a new page. (This issue occurs on GoDAM Video block as well.) I need to know if this is expected. If yes then we should add a help text that if mini cart is disabled then "Add to Cart" button will redirect to the cart page.
2. When mini cart is opened in the GoDAM Gallery, it has a black background and accessibility issues.

### Recording:

https://github.com/user-attachments/assets/e9473303-3194-4fcc-ac46-f7faf0f7822b

rudrakshi-gupta (approved):

PR #1777, opened 2026-04-08
Feat: WooCommerce product gallery videos

rudrakshi-gupta:
Suggested by @joelabreo227 and team.

### Overview
This PR adds support for videos within the WooCommerce product gallery, allowing videos to be displayed and played inline alongside product images on the single product page.

---

### What’s Added

- Ability to add video attachments via the product gallery (admin)
- Videos are rendered within the existing gallery slider (same as images)
- Inline video playback inside the gallery (no separate modal required)
- Support for looping and muted playback
- Seamless integration with existing WooCommerce gallery structure
- Backward compatible with existing image-only galleries

---

### Implementation Details

- Extended the product gallery to accept both `image` and `video` media types
- Stored video attachments using the existing `_product_image_gallery` meta key
- Replaced thumbnail-only behavior with actual video rendering in gallery slides
- Updated frontend logic to:
  - Inject `<video>` elements into gallery items
  - Handle playback within WooCommerce’s gallery (Flexslider)
- Removed dependency on modal-based playback for gallery videos

---

### Behavior

- Videos appear as part of the product gallery carousel
- When a video slide is active:
  - Video is playable inline
  - Autoplay can be triggered when the slide becomes active (if enabled)
- When navigating away:
  - Video playback is paused to prevent background playing

---

### Access Control

- Existing API key validation is respected
- Video functionality remains restricted where required (e.g., Pro feature gating)

---

### Testing Notes

- Tested with:
  - Multiple images and videos in the gallery
  - Mixed ordering of media
  - Gallery navigation (next/prev thumbnails)
- Verified no regression in:
  - Image rendering
  - Zoom/lightbox behavior for images

---

### Before vs After

**Before:**
- Videos displayed as thumbnails
- Clicking opened a separate modal player

**After:**
- Videos render directly inside the product gallery
- Playback happens inline within the gallery slider

---

### Summary

This update enhances the WooCommerce product experience by making videos a first-class media type in the gallery, improving product storytelling and user engagement.

## Demo

https://github.com/user-attachments/assets/75286570-be6e-4853-add3-bcc846aba69f


integrations/woocommerce/classes/class-wc-featured-video-gallery.php
```diff
@@ -397,6 +398,7 @@ public function handle_send_empty_alts() {
 				}
 
 				$video_thumbnails[] = esc_url( get_post_meta( $video_id, 'rtgodam_media_video_thumbnail', true ) ?: self::FALLBACK_THUMBNAIL );
+				$video_urls[]       = wp_get_attachment_url( $video_id );
```
subodhr258: Bug: $video_id may be undefined or stale if the `if` block above never ran

Video URLs and Video Thumbnails should be assigned in the same place as Video ID, within the if condition at line 395.

integrations/woocommerce/assets/js/featured-video/wc-featured-video-gallery.js
subodhr258: `resetVideoModal()` was removed, so the import can also be removed.

Same with `loadNewVideo`, `registerEscapeHandler`, `unregisterEscapeHandler`

subodhr258: Since `data.videoUrls` is not checked, we can use optional chaining below, or check it here in the `if` condition.

subodhr258 (commented):
Left some comments, otherwise LGTM

subodhr258 (approved):

Issue #958, opened 2025-08-21
Media Replacement without changing URL


nayemDevs:
## Objective

Allow users to replace existing media files with updated versions **without changing the URL** or breaking any existing embeds (on WordPress, external sites, or via API).

## Suggested Workflow

Update 1 Dec 2025: https://docs.google.com/document/d/1vPk-CLEUv6p_09sHJYgTyiBZg-xqRTVaZ0DoyUWh8rM/edit?tab=t.0#heading=h.sgccb83hhdf0

Original workflow:

1. **Select Media to Replace**
   - User selects an asset in Media Library.
   - Clicks **"Replace Media"** option in media details view.

2. **Upload New Media**
   - User uploads a new file (same media type).
   - System validates:
     - File type matches (video → video, image → image, etc.)
     - Format supported (codec, resolution, duration checks).
     - Warning if mismatch.

3. **Versioning System**
   - Old file archived as **Version n-1**.
   - New file stored as **Version n** but retains **same media ID & URL**.
   - Rollback option available from version history.

4. **Storage & CDN Update**
   - New file uploaded to GoDAM storage.
   - CDN cache purged → fresh file available worldwide.
   - Metadata (title, description, tags) preserved unless user edits.

5. **WordPress Plugin Sync**
   - GoDAM triggers webhook → WP plugin updates local media record.
   - WP keeps same URL in embeds, shortcodes, and Gutenberg blocks.
   - Seamless for WooCommerce product media & blog posts.

6. **Notifications & Logs**
   - User notified: *“Media replaced successfully. All embeds updated.”*
   - Activity log entry created:
     ```
     Media ID 123 replaced by [user] on [date]
     ```
   - Rollback action available in the log.

### GoDAM Central
- [ ] Implement **Replace Media** action in Media Library.
- [ ] Add **file validation** (type, size, codec).
- [ ] Build **versioning system** with rollback support.
- [ ] Handle **storage overwrite + CDN purge**.
- [ ] Trigger **webhook event** (`media.replaced`).
- [ ] Add **activity logs** for audit trail.

### WordPress Plugin
- [ ] Listen for **`media.replaced` webhook**.
- [ ] Update local DB record but **retain same URL**.
- [ ] Ensure Gutenberg blocks & shortcodes remain functional.
- [ ] Add UI notice in WP Admin after replacement.
- [ ] Test WooCommerce product media sync.
- [ ] Provide optional **rollback UI** (restore previous version).

## Acceptance Criteria

[...]

PR #1784, opened 2026-04-14
feat: Add Virtual Media Registrar and Migration classes for handling virtual media integration

shreyasikhar:
### Summary

This PR introduces two classes that allow a WordPress site to integrate as a **virtual media site** with GoDAM Central. When a virtual media attachment (identified by `_godam_original_id` meta) is added to the Media Library, this site is automatically registered with GoDAM Central so it receives transcoder callbacks for that job. A companion REST API endpoint allows GoDAM Central to trigger a one-time migration that reconciles all existing virtual attachments.

---

#### `inc/classes/class-virtual-media-registrar.php` _(new)_

Singleton class that hooks into the WordPress attachment lifecycle to keep site registration with GoDAM Central in sync.

**Registration** (`add_virtual_media_site`)
- Fires on `added_post_meta` / `updated_post_meta` (priority 10) when `_godam_original_id` is written, and on `add_attachment` (priority 21) as a fallback — priority 22 ensures it runs after the transcoding is complete for the new attachment.
- Requires `rtgodam-api-key` option and `RTGODAM_API_BASE` constant; returns `WP_Error` early if either is missing.
- Calls `godam_core.api.transcoder_job.add_virtual_media_site` on GoDAM Central with `job_name`, `site_url`, `callback_url`, and `api_key`.

**De-registration** (`remove_virtual_media_site`)
- Hooks `before_delete_post` to detect when a virtual attachment is deleted.
- Calls `godam_core.api.transcoder_job.remove_virtual_media_site` on GoDAM Central with `job_name`, `site_url`, and `api_key`.
- Logs failures via `error_log` without blocking the delete operation.

#### `inc/classes/rest-api/class-virtual-media-migration.php` _(new)_

REST endpoint for GoDAM Central to trigger a migration pass over all existing virtual attachments on this site.

image

- **Route:** `POST /godam/v1/virtual-media-migration`
- **Auth:** `api_key` body param validated with `hash_equals` against the stored `rtgodam-api-key` option.
- **Response:** returns `site_url`, `callback_url`, and a deduplicated array of `job_ids` — all `_godam_original_id` values from attachments that have a non-empty value for that meta key.
- `get_job_ids_for_migration()` queries attachments in batches of 100 (paginated `WP_Query`) to avoid memory issues on large sites, then returns `array_values( array_unique( $job_ids ) )`.

---

### Meta keys used

| Key | Purpose |
|---|---|
| `_godam_original_id` | Stores the GoDAM job name for a virtual attachment (pre-existing) |

---

### Testing

1. Add a virtual attachment (one with `_godam_original_id` set) — confirm `godam_core.api.transcoder_job.add_virtual_media_site` is called
2. Add the same attachment again / update its meta — confirm the registration API call is **not** repeated.
3. Delete the virtual attachment — confirm `godam_core.api.transcoder_job.remove_virtual_media_site` is called.
4. `POST /godam/v1/virtual-media-migration` with a valid `api_key` — confirm the response includes `site_url`, `callback_url`, and a correct deduplicated `job_ids` array.
5. Call the migration endpoint with an invalid/missing `api_key` — confirm a `403` is returned.

## Issue

https://github.com/rtCamp/godam/issues/958#issuecomment-4191158343
https://github.com/rtCamp/godam/issues/958#issuecomment-4197660474

Slack thread for reference: https://rtcamp.slack.com/archives/C25QZJ50B/p1776062860628749?thread_ts=1776051004.034789&cid=C25QZJ50B


inc/classes/class-virtual-media-registrar.php
```diff
@@ -0,0 +1,271 @@
...
+			'job_name' => $job_name,
+			'site_url' => get_site_url(),
+			'api_key'  => $api_key,
+		);
+
+		$response = wp_remote_post(
+			$endpoint,
+			array(
+				'method'  => 'POST',
+				'timeout' => 30,  // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
```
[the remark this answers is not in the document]
shreyasikhar: Reduced it

inc/classes/class-virtual-media-registrar.php
```diff
@@ -0,0 +1,271 @@
...
+
+	/**
+	 * Boot hooks.
+	 *
+	 * @return void
+	 */
+	protected function setup_hooks() {
+		add_action( 'added_post_meta', array( $this, 'maybe_register_from_meta_change' ), 10, 3 );
+		add_action( 'updated_post_meta', array( $this, 'maybe_register_from_meta_change' ), 10, 3 );
+		add_action( 'add_attachment', array( $this, 'maybe_register_from_attachment' ), 22, 1 );
```
subodhr258: WordPress fires the `add_attachment` hook before `_godam_original_id` is set, so `register_site_for_attachment_if_needed` will always early return with `true` here. I think we can remove this hook for `maybe_register_from_attachment`.

inc/classes/rest-api/class-virtual-media-migration.php
```diff
@@ -0,0 +1,147 @@
...
+					'post_status'    => 'inherit',
+					'fields'         => 'ids',
+					'posts_per_page' => $posts_per_page,
+					'paged'          => $current_page,
+					'orderby'        => 'ID',
+					'order'          => 'ASC',
+					'meta_query'     => array(  // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+						'relation' => 'AND',
+						array(
+							'key'     => '_godam_original_id',
```
subodhr258: The hardcoded key can be replaced with this:
`'key' => \RTGODAM\Inc\Virtual_Media_Registrar::META_ORIGINAL_ID,`

inc/classes/rest-api/class-virtual-media-migration.php
```diff
@@ -0,0 +1,147 @@
...
+							'compare' => 'EXISTS',
+						),
+						array(
+							'key'     => '_godam_original_id',
```
subodhr258: The hardcoded key can be replaced with this:
`'key' => \RTGODAM\Inc\Virtual_Media_Registrar::META_ORIGINAL_ID,`

inc/classes/rest-api/class-virtual-media-migration.php
```diff
@@ -0,0 +1,147 @@
...
+			);
+
+			$query_count = count( $query->posts );
+
+			if ( empty( $query_count ) ) {
+				break;
+			}
+
+			foreach ( $query->posts as $attachment_id ) {
+				$job_id = get_post_meta( $attachment_id, '_godam_original_id', true );
```
subodhr258: Above we are getting all the IDs and then we're calling `get_post_meta()` for every single ID, O(N+1).

You can get the Job ID as well for this, in one single query, something like this:


```php
global $wpdb;

$job_ids = $wpdb->get_col(
    $wpdb->prepare(
        "SELECT DISTINCT meta_value
         FROM {$wpdb->postmeta} pm
         INNER JOIN {$wpdb->posts} p ON p.ID = pm.post_id
         WHERE pm.meta_key = %s
           AND pm.meta_value != ''
           AND p.post_type = 'attachment'
           AND p.post_status = 'inherit'",
        '_godam_original_id'
    )
);
```

subodhr258 (commented):
Reviewed the code, added some comments

inc/classes/class-virtual-media-registrar.php
shreyasikhar: We are setting the meta here https://github.com/rtCamp/godam/blob/main/inc/classes/rest-api/class-media-library.php#L1581-L1586 on the same `add_attachment` hook with priority 1.

inc/classes/rest-api/class-virtual-media-migration.php
shreyasikhar: This works, but it won't be cached.

Do you want me to use transients to avoid query hitting the DB directly.

```php
$job_ids = get_transient( 'rtgodam_virtual_media_job_ids' );

if ( false === $job_ids ) {
    global $wpdb;
    $job_ids = $wpdb->get_col( $wpdb->prepare( 
        "SELECT DISTINCT meta_value FROM {$wpdb->postmeta} pm 
         INNER JOIN {$wpdb->posts} p ON p.ID = pm.post_id 
         WHERE pm.meta_key = %s AND pm.meta_value != '' 
         AND p.post_type = 'attachment'", 
        '_godam_original_id' 
    ) );
    set_transient( 'rtgodam_virtual_media_job_ids', $job_ids, HOUR_IN_SECONDS );
}
```

shreyasikhar: Implemented the above

KMchaudhary:
@shreyasikhar, I have tested the PR changes with @geekofshire  and found that the virtual media deletion function `remove_virtual_media_site` itself is not triggering from our plugin end. I verified it by adding the debug log.

- Old virtual media migration is working fine.
- New virtual media registration is also working fine.
- Old VM and new VM deletion requests to Frappe are not being sent from our plugin.

In short, the `api/method/godam_core.api.transcoder_job.remove_virtual_media_site` call is not triggered from our plugin when the attachment is deleted.

shreyasikhar:
@KMchaudhary Thanks for the report. I have fixed the problem by updating the hook with the right one that runs for attachments. Can you please re-check?

KMchaudhary:
@shreyasikhar @geekofshire, there is one testcase we should consider. If the user adds/deletes virtual media in bulk, in that case we should make sure the operation registration/de-registration operation will not break due to the rate-limit on new endpoints. We used to get such errors for API Key verifications. 
cc: @elifvish

geekofshire:
@KMchaudhary These APIs are not rate-limited, so we should not face this error

KMchaudhary (approved):
@shreyasikhar 
It will be good if we can collect all the add/remove attachment requests and send an API call to Frappe to update those attachments.

**Suggestions:**
We can use the hooks with a higher priority to collect the added/deleted attachments data, and then use the low-priority hook to send api call for the collected data.

shreyasikhar:
Ok, suggestion LGTM.
But, since @geekofshire confirmed that these APIs are not rate limited and we should not face this error. I think we can skip this optimization as of now. 
If needed later, we can do it. Also, I think the API needs to accept multiple job ids, so with one API call we can send multiple job ids during bulk operation.

elifvish (changes requested):
1. Duplicate API calls per attachment — registration fires 2+ times
The hooks overlap for a single virtual media creation. In [class-media-library.php:1645-1672](vscode-webview://1uantt3of5i1f6v1b9bqbf98ulhfpr2146h0kcj3m38evi76k4uh/inc/classes/rest-api/class-media-library.php#L1645-L1672):

The pre-setter sets _godam_original_id at add_attachment priority 1 → triggers added_post_meta → HTTP call 1.
add_attachment priority 22 (this new class) → HTTP call 2.
update_post_meta at line 1668 may fire updated_post_meta if the row state differs → HTTP call 3.
There is no idempotency / "already registered" guard. Suggest tracking registration in a per-attachment meta (e.g. _godam_virtual_site_registered) and short-circuiting if present, or storing a hash of {job_name, site_url} so re-registration only fires when something material changes.

2. Errors from registration are silently dropped
[class-virtual-media-registrar.php:108](vscode-webview://1uantt3of5i1f6v1b9bqbf98ulhfpr2146h0kcj3m38evi76k4uh/inc/classes/class-virtual-media-registrar.php#L108) register_site_for_attachment_if_needed returns WP_Error but the two hook callbacks (maybe_register_from_meta_change, maybe_register_from_attachment) discard the return. A site that never registers will fail silently forever. Mirror the error_log you already added in maybe_remove_virtual_media_site_on_delete.

3. Synchronous HTTP in hot paths blocks WP requests
add_attachment priority 22 with a 10-second timeout blocks every upload.
delete_attachment blocks bulk deletes by up to 5 s each.
Consider deferring via Action Scheduler (as_enqueue_async_action) — the existing code already uses it elsewhere ([class-video-migration.php:260,396](vscode-webview://1uantt3of5i1f6v1b9bqbf98ulhfpr2146h0kcj3m38evi76k4uh/inc/classes/rest-api/class-video-migration.php#L260)).

4. Migration endpoint transient is never invalidated
[class-virtual-media-migration.php:104-118](vscode-webview://1uantt3of5i1f6v1b9bqbf98ulhfpr2146h0kcj3m38evi76k4uh/inc/classes/rest-api/class-virtual-media-migration.php#L104-L118) caches rtgodam_virtual_media_job_ids for 1 hour but the registrar never calls delete_transient when new virtual media is added or one is deleted. Central can receive stale lists for up to an hour after changes. Invalidate in both register_site_for_attachment_if_needed and maybe_remove_virtual_media_site_on_delete.

@subodhr258 the first point, I need you to look into as this seems to pre exsisting, I need to know the reason for this

shreyasikhar:
Thanks @elifvish for the feedback.

I have addressed your queries:
1. Added idempotency checks for duplicate registrations, also removed `add_attachment` hook callback since post meta cover this case.
2. Errors from registration will now be logged 
3. Used Action Scheduler to avoid HTTP request delays
4. Added invalidation to Migration endpoint transient

 I have tested the new changes. Please let me know if any more feedback.

elifvish (approved):

PR #1855, opened 2026-05-07
Fix iOS autoplay issue 

elifvish:
This pull request improves the autoplay and muted handling for the Godam video player, focusing on better compatibility with browser autoplay policies (especially on iOS) and more accurate player configuration. The main changes ensure that videos set to autoplay are always muted, update the way autoplay and muted attributes are set in both the backend and frontend, and refine the logic for when autoplay-on-view is enabled.

**Autoplay and Muted Handling Improvements:**

* The backend now sets the `autoplay` option to `'muted'` (instead of just `true`) and forces `muted: true` when autoplay is enabled, ensuring compliance with browser autoplay requirements. (`inc/templates/godam-player.php`)
* The `<video>` element now always includes the `muted` attribute if either autoplay or muted is enabled, and the `autoplay` attribute is only included if autoplay is enabled, ensuring the correct HTML attributes are present for browser compatibility. (`inc/templates/godam-player.php`)

**Frontend Player Logic Updates:**

* The player manager in JavaScript now always sets the player to muted before attempting to play, and centralizes the play logic to better handle autoplay restrictions, especially on iOS devices. (`assets/src/js/godam-player/managers/playerManager.js`)

**Autoplay-on-View Logic Refinement:**

* The `data-autoplay-on-view` attribute is now set to `true` only if autoplay is enabled and `preload` is not set to `'auto'`, preventing unwanted autoplay behavior when videos are preloaded. (`inc/templates/godam-player.php`)

### DIFF
--- a/inc/templates/godam-player.php
+++ b/inc/templates/godam-player.php
@@ -437,9 +437,9 @@ if ( $godam_placeholder_lookup_id > 0 ) {
 }
 $godam_video_setup = array(
 	'controls'    => $godam_controls,
-	'autoplay'    => $godam_autoplay,
+	'autoplay'    => $godam_autoplay ? 'muted' : false,
 	'loop'        => $godam_loop,
-	'muted'       => $godam_muted,
+	'muted'       => $godam_autoplay ? true : $godam_muted,
 	'preload'     => $godam_preload,
 	'poster'      => $godam_video_poster,
 	'fluid'       => true,
--- a/inc/templates/godam-player.php
+++ b/inc/templates/godam-player.php
@@ -437,7 +437,7 @@ if ( $godam_placeholder_lookup_id > 0 ) {
 }
 $godam_video_setup = array(
 	'controls'    => $godam_controls,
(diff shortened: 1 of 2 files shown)


inc/templates/godam-player.php
```diff
@@ -437,9 +437,9 @@ function () use ( $godam_player_wrapper_css ) {
 }
 $godam_video_setup = array(
 	'controls'    => $godam_controls,
-	'autoplay'    => $godam_autoplay,
+	'autoplay'    => $godam_autoplay ? 'muted' : false,
```
KMchaudhary: Here autoplay should use $godam_autoplay variable only.

We added `'muted'       => $godam_autoplay ? true : $godam_muted` condition below

KMchaudhary (dismissed):

subodhr258 (dismissed):

KMchaudhary (approved):

subodhr258 (approved):

PR #1893, opened 2026-05-25
fix: Video block editor player collapsed to play button (#1891)

### LINKED ISSUE #1891: Fix Video Player loading issue on Block Editor Page after selecting video in GoDAM Video Block
## Bug Report

**Current Behavior**
Broken editor with only play button visible 

**Expected behavior/code**
Video Player should appear with proper thumbnail and player option in video

**Steps to reproduce the bug**
Visit any Page
Add GoDAM Video Block
Select Video/Virtual Media
Observe broken player 

**Testing Environment (if applicable)**
https://single-kis.focused.rt.gw/wp-admin 
GoDAM 1.10.1

**Additional context/Screenshots**
Image

subodhr258:
## Summary

- The GoDAM Video block in the block editor was rendering as a tiny play button after a video was selected — see [#1891](https://github.com/rtCamp/godam/issues/1891).
- Two issues combined to collapse the player:
  1. The video container had inline `display: none` until 500 ms after Video.js initialised, so Video.js couldn't measure its parent and started up with zero dimensions.
  2. Video.js writes its fluid `padding-top` rule to the top-level document. The editor canvas runs in an iframe, so the rule never reached the player and `.video-js` kept `padding-top: 0` — leaving only the big-play button visible.
- Render the video container immediately, stack the loading overlay on top via `position: absolute`, and mirror Video.js's fluid padding inline on the player element so the height resolves inside the editor iframe. Add a proper `.hide` rule so the loading overlay actually fades out once the player is ready.

## Before / After

| Before (#1891) | After |
|---|---|
| Tiny grey box with only the play button | Full 16:9 player with centred play button |

Verified locally on `subodh-multisite.rt.gw` by inserting a `godam/video` block with a sample MP4 source — the player now renders at 645×363 (16:9) and the loading overlay fades cleanly.

## Test plan

- [ ] Open the block editor, insert a GoDAM Video block, pick any uploaded video — player renders at the configured aspect ratio with the big play button centred.
- [ ] Switch aspect ratios in the inspector (16:9 → 4:3 → custom) — player resizes accordingly.
- [ ] Reload the editor with a saved post containing the block — player still renders at the correct size on initial mount.
- [ ] Frontend rendering of the same post is unaffected (no changes to `render.php` / frontend JS).
- [ ] Loading overlay briefly shows on first mount, then fades out after Video.js is ready.

Fixes #1891

### DIFF
--- a/assets/src/blocks/godam-player/VideoJS.js
+++ b/assets/src/blocks/godam-player/VideoJS.js
@@ -17,6 +17,23 @@ import 'videojs-flvjs-es6';
  */
 import { useRef, useEffect, useMemo } from '@wordpress/element';
 
+// Mirror Video.js's fluid aspect-ratio padding inline on the player element.
+// Video.js generates this rule in the top document only, which the block-editor iframe
+// can't see — without an explicit padding-top the player collapses to zero height.
+const applyAspectRatioPadding = ( player, aspectRatio ) => {
+	if ( ! player || ! aspectRatio || ! /^\d+:\d+$/.test( aspectRatio ) ) {
+		return;
+	}
+	const [ x, y ] = aspectRatio.split( ':' ).map( Number );
+	if ( ! x || ! y ) {
+		return;
+	}
+	const playerEl = player.el_;
+	if ( playerEl ) {
+		playerEl.style.paddingTop = `${ ( y / x ) * 100 }%`;
+	}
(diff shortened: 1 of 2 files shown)


KMchaudhary:
Closing the PR as task was done by #1906

PR #1911, opened 2026-06-02
fix(analytics): cap engagement rate at 100% in dashboard charts

subodhr258:
## Problem (P0)

The GoDAM dashboard reported an **Engagement Rate of 318.81%**. Engagement is the share of a video watched and cannot exceed 100%.

## Fix (display-side guard)

The root-cause fix lives in the analytics service — rtCamp/godam-analytics#172 — which keeps `play_time` consistent with `plays` (orphan playback excluded from both). This PR adds the matching **display-side guard** in the WP plugin so the UI never renders above 100%, even from stale or legacy cached data:

- `calculateEngagementRate()` clamps to 100%.
- `PlaybackPerformance` dashboard-mode and single-video computations clamp to 100%.

## Notes

Once both PRs ship, engagement is bounded server-side *and* client-side. Existing dashboards displaying historical (pre-fix) rows will now show ≤ 100% instead of the 318% spike.

### DIFF
--- a/pages/analytics/PlaybackPerformance.js
+++ b/pages/analytics/PlaybackPerformance.js
@@ -85,7 +85,9 @@ export default function PlaybackPerformanceDashboard( {
 				if ( mode === 'dashboard' ) {
 					return {
 						date,
-						engagement_rate: +entry.avg_engagement?.toFixed( 2 ) || 0,
+						// Engagement is a share of the video watched: cap at 100%
+						// as a display guard (the backend already bounds it).
+						engagement_rate: Math.min( +entry.avg_engagement?.toFixed( 2 ) || 0, 100 ),
 						play_rate: +( entry.play_rate * 100 || 0 ).toFixed( 2 ),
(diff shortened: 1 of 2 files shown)


subodhr258:
Closing — this is purely an analytics-side data bug, not a plugin issue.

Root cause (fixed in rtCamp/godam-analytics#172): the analytics service summed *all* `type=2` watch ranges into `play_time`, including "orphan" playback (a `type=2` with no preceding `type=1` page_load — cached page / SPA / iframe). `plays` already excludes those, so `play_time` could exceed `plays × video_length` and engagement went past 100% (the 318.81% on the dashboard). #172 makes `play_time` count only ranges belonging to a counted play, so engagement is ≤100% by construction.

A display-side `Math.min(…, 100)` here would only mask the value, not fix it — and would silently swallow any future regression. With #172 plus a historical reprocess (`migrate_historical_data`), no data exceeds 100% anywhere, so the plugin cap isn't needed.

PR #1956, opened 2026-06-15
feat(layer-analytics): record sub-hotspot lifetimes + forward to analytics (#196 WP side)

subodhr258:
WordPress companion for **rtCamp/godam-analytics#196** (analytics side: rtCamp/godam-analytics#209). Implements **direction (b)** — record per-sub-hotspot lifetimes and forward them so the analytics service credits a sub's inherited `viewed` only for the days it actually existed.

## Problem
Sub-hotspots (hotspot items / Woo product hotspots) don't emit their own `viewed` — the parent layer fires one impression per session and the service copies the parent's **full-range** `viewed` onto every sub. So a **deleted** sub keeps gaining Views forever, and a **newly added** sub instantly shows Views from before it existed.

## What this PR does (3 coupled changes)
1. **Record lifetimes on save** — `class-meta-rest-fields.php`: the `rtgodam_meta` update_callback now diffs old-vs-new layers and records `added_at`/`deleted_at` per composite sub id (`parent::hotspot.id` / `parent::p<productId>`, matching what the player emits) into `rtgodam_meta['layer_lifetimes']`.
   - A sub that **appears** this save → `added_at = today` (covers re-adds).
   - A sub that **disappears** → `deleted_at = today` (the row persists past deletion).
   - A **pre-existing, untouched** sub gets **no entry** → the service treats it as full-history, so historical content is **not** retroactively truncated.
   - Dates are UTC `Y-m-d` to match the service's daily buckets.
2. **Forward the map** — `class-analytics.php`: for `hotspot`/`woo` layer types the proxy POSTs the lifetime map (JSON body) to `/processed-layer-analytics/`, preserving the existing query params + 3s timeout + soft-error handling. Atomic types (cta/form/poll) keep the GET path.
3. **Use the server's sliced `viewed`** — `useVideoLayerData.js`: the dashboard hook stops re-inheriting the parent's full-range `viewed` onto sub rows client-side (which would have undone the slice) and uses the server's per-sub value.

## Safety / sequencing
- **Forward-compatible & no-op until both sides are live:** a sub absent from the map = full history; and a service **without** rtCamp/godam-analytics#209 simply ignores the POST body (query params still apply). Change 3 is a no-op against today's service (sub `viewed` already equals the parent's) and only takes effect once the map + #209 are live.
- Recommended order: ship the analytics side (#209) first or together, then this.

## Scope
WooCommerce product hotspots are stored in **this plugin's** `rtgodam_meta` and saved through this plugin's pipeline, so **godam-for-woo needs no change** for direction (b). (Direction (c) — `active_subs` on the emitted event — is the only thing that would touch godam-for-woo; not in scope here.)

## Testing notes
No PHPUnit harness in this repo. Manual: create a hotspot/Woo layer with ≥2 subs, save; delete one sub, save → `rtgodam_meta['layer_lifetimes']` gets a `deleted_at` for it; add a new sub, save → it gets `added_at`. With #209 deployed, the removed sub's Views stop climbing and the new sub no longer shows pre-existence Views.

### DIFF
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -267,11 +267,37 @@ class Analytics extends Base {
 			$query_params['job_id'] = $job_id;
 		}
 
+		// #196: for layer types that have sub-hotspots, forward the per-sub
+		// lifetime map (added_at / deleted_at, recorded on save) so the service
+		// credits a sub's inherited `viewed` only for the days it existed. The
+		// endpoint accepts GET or POST and reads `layer_lifetimes` from the body;
+		// atomic layer types send nothing and keep the GET path. A sub absent
+		// from the map is treated as full-history, so this is a no-op until the
+		// map is populated.
+		$layer_lifetimes = array();
+		if ( $attachment_id && in_array( $layer_type, array( 'hotspot', 'woo' ), true ) ) {
+			$meta = get_post_meta( $attachment_id, 'rtgodam_meta', true );
+			if ( is_array( $meta ) && ! empty( $meta['layer_lifetimes'] ) && is_array( $meta['layer_lifetimes'] ) ) {
+				$layer_lifetimes = $meta['layer_lifetimes'];
+			}
+		}
+
 		$endpoint = add_query_arg( $query_params, RTGODAM_ANALYTICS_BASE . '/processed-layer-analytics/' );
 		// Bounded timeout: useVideoLayerData fires one of these per layer type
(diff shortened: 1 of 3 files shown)


subodhr258:
Superseded by #2064. Reopening this one was refused by GitHub because the branch was force-pushed after close (rebased onto current `develop`), so the work continues there on the same branch.

Closed here as a scope call: sub-hotspot data was not going to be displayed. That has since changed, per rtCamp/godam-plugin-wp#26, which puts per-hotspot rows on screen with revenue attached.

PR #1964, opened 2026-06-17
Add DependencyExtractionWebpackPlugin for pages and read deps from .min.asset.php

KMchaudhary:
This pull request updates how JavaScript assets are registered and enqueued throughout the plugin to make use of dynamically generated `.asset.php` files. This ensures that each script's dependencies and version are accurately set based on the build process, improving reliability and maintainability when assets change.

**Asset management improvements:**

* Updated all script registrations (both frontend and admin) to load dependencies and version information from corresponding `.asset.php` files generated during the build process, falling back to defaults if the files are missing. This affects scripts in `class-assets.php`, `class-deactivation.php`, `class-elementor-widgets.php`, `class-pages.php`, `class-video-embed.php`, `class-video-preview.php`, `class-godam-media.php`, `class-everest-forms-field-godam-video.php`, `class-recorder-field.php`, and `class-init.php`. [[1]](diffhunk://#diff-e329e454999381b4aad93a6c5811e683aed9bf08f0428fb7f991feb9975b791dR49-R64) [[2]](diffhunk://#diff-e329e454999381b4aad93a6c5811e683aed9bf08f0428fb7f991feb9975b791dR193-R208) [[3]](diffhunk://#diff-e329e454999381b4aad93a6c5811e683aed9bf08f0428fb7f991feb9975b791dR356-R371) [[4]](diffhunk://#diff-bd13381b748a6c124d6b366f2e43b0d584e15f652d9c0a0de38ea11f50a9b91fR44-R59) [[5]](diffhunk://#diff-36c6d43153787a95b423e72fa0a35cf7e3492244930adb8832ae91faff664391R107-R122) [[6]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR442-R457) [[7]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR476-R491) [[8]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR678-R693) [[9]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR764-R779) [[10]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR913-R928) [[11]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR958-R973) [[12]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR1061-R1076) [[13]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR1108-R1123) [[14]](diffhunk://#diff-8f6c0412582a633d4d6c9cbd98466c31f29eecb1e995a3118df0cad252fa827eR1148-R1163) [[15]](diffhunk://#diff-48a53a0255077bc13c0a1e7769d5ca1e356c5c46fe2d14cc85d390d061771fa4R77-R92) [[16]](diffhunk://#diff-786ef9e563a3529a67b8abf2e9933c972da53b5a073281871e7ba0b45636eac6R75-R90) [[17]](diffhunk://#diff-bf4f8109daa9549847e1dfd8d78a40832f7abae6af82b728ea31bb8695d779c7L79-R90) [[18]](diffhunk://#diff-639ca3702364e00212af65dd6fbdbd60cbeb33bf9dd1767c0dd2d86a725c10b7R500-R515) [[19]](diffhunk://#diff-29f0075a9d523c4aa39543e5658e0ca8277189ce5af874a41d47607d29c37908R313-R328) [[20]](diffhunk://#diff-5ceffc3c921130634b19da6bc6e886c0dcba4863fc74d9558192e3e5acedbaa0R225-R240)

**Dependency handling enhancements:**

* Where appropriate, dependencies from the asset files are merged with required runtime dependencies (such as `'godam-player-frontend-script'` or form plugin requirements), ensuring scripts are loaded in the correct order. [[1]](diffhunk://#diff-36c6d43153787a95b423e72fa0a35cf7e3492244930adb8832ae91faff664391R107-R122) [[2]](diffhunk://#diff-786ef9e563a3529a67b8abf2e9933c972da53b5a073281871e7ba0b45636eac6R75-R90)

**Fallback and compatibility:**

* If an `.asset.php` file is missing, the code falls back to an empty dependencies array and uses the file modification time for versioning, maintaining backward compatibility and preventing asset loading failures. (applies throughout all affected files and changes above)

These changes make asset loading more robust, especially as the codebase grows and asset dependencies change over time.

### DIFF
--- a/inc/classes/shortcodes/class-godam-player.php
+++ b/inc/classes/shortcodes/class-godam-player.php
@@ -145,11 +145,17 @@ class GoDAM_Player {
 		$player_analytics_path = RTGODAM_PATH . 'assets/build/js/godam-player-analytics.min.js';
 
 		if ( file_exists( $player_analytics_path ) ) {
+			$player_analytics_asset_file = RTGODAM_PATH . 'assets/build/js/godam-player-analytics.min.asset.php';
+			$player_analytics_asset      = file_exists( $player_analytics_asset_file ) ? include $player_analytics_asset_file : array(
+				'dependencies' => array(),
+				'version'      => filemtime( $player_analytics_path ),
+			);
+
 			wp_register_script(
 				'godam-player-analytics-script',
 				RTGODAM_URL . 'assets/build/js/godam-player-analytics.min.js',
-				array( 'godam-player-frontend-script', 'wp-i18n' ),
-				filemtime( $player_analytics_path ),
+				array_merge( $player_analytics_asset['dependencies'], array( 'godam-player-frontend-script' ) ),
+				$player_analytics_asset['version'],
 				true
 			);
 		}
--- a/inc/classes/class-assets.php
+++ b/inc/classes/class-assets.php
@@ -46,11 +46,22 @@ class Assets {
 	 */
 	public function enqueue_scripts() {
 
+		$main_asset_file = RTGODAM_PATH . 'assets/build/js/main.min.asset.php';
+		$main_asset      = array(
+			'dependencies' => array(),
+			'version'      => filemtime( RTGODAM_PATH . 'assets/build/js/main.min.js' ),
+		);
+
+		if ( file_exists( $main_asset_file ) ) {
+			// phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable -- file path is constant.
+			$main_asset = include $main_asset_file;
+		}
+
 		wp_register_script(
 			'rtgodam-script',
 			RTGODAM_URL . 'assets/build/js/main.min.js',
-			array(),
-			filemtime( RTGODAM_PATH . 'assets/build/js/main.min.js' ),
+			$main_asset['dependencies'],
+			$main_asset['version'],
 			true
 		);
 
@@ -179,11 +190,22 @@ class Assets {
 			wp_enqueue_media();
 		}
 
+		$admin_asset_file = RTGODAM_PATH . 'assets/build/js/admin.min.asset.php';
+		$admin_asset      = array(
+			'dependencies' => array(),
+			'version'      => filemtime( RTGODAM_PATH . 'assets/build/js/admin.min.js' ),
+		);
+
+		if ( file_exists( $admin_asset_file ) ) {
+			// phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable -- file path is constant.
+			$admin_asset = include $admin_asset_file;
+		}
+
 		wp_register_script(
 			'rtgodam-script',
 			RTGODAM_URL . 'assets/build/js/admin.min.js',
-			array(),
-			filemtime( RTGODAM_PATH . 'assets/build/js/admin.min.js' ),
+			$admin_asset['dependencies'],
+			$admin_asset['version'],
 			true
 		);
 
@@ -331,11 +353,22 @@ class Assets {
 
 		// Only enqueue HTTP auth detector on uploads page or pages where media uploading is possible.
 		if ( godam_should_load_auth_detector_script( $screen ) ) {
+			$http_auth_asset_file = RTGODAM_PATH . 'assets/build/js/http-auth-detector.min.asset.php';
+			$http_auth_asset      = array(
+				'dependencies' => array(),
+				'version'      => filemtime( RTGODAM_PATH . 'assets/build/js/http-auth-detector.min.js' ),
+			);
+
+			if ( file_exists( $http_auth_asset_file ) ) {
+				// phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable -- file path is constant.
+				$http_auth_asset = include $http_auth_asset_file;
+			}
+
 			wp_register_script(
 				'godam-http-auth-detector',
 				RTGODAM_URL . 'assets/build/js/http-auth-detector.min.js',
-				array( 'jquery' ),
-				filemtime( RTGODAM_PATH . 'assets/build/js/http-auth-detector.min.js' ),
+				$http_auth_asset['dependencies'],
+				$http_auth_asset['version'],
 				true
 			);
 	
--- a/inc/classes/class-deactivation.php
+++ b/inc/classes/class-deactivation.php
@@ -41,11 +41,22 @@ class Deactivation {
 	public function load_scripts() {
 		global $pagenow;
 
+		$deactivation_asset_file = RTGODAM_PATH . 'assets/build/js/deactivation-feedback.min.asset.php';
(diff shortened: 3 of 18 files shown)


inc/classes/shortcodes/class-godam-player.php
```diff
@@ -145,11 +145,17 @@ public function register_scripts() {
 		$player_analytics_path = RTGODAM_PATH . 'assets/build/js/godam-player-analytics.min.js';
 
 		if ( file_exists( $player_analytics_path ) ) {
+			$player_analytics_asset_file = RTGODAM_PATH . 'assets/build/js/godam-player-analytics.min.asset.php';
+			$player_analytics_asset      = file_exists( $player_analytics_asset_file ) ? include $player_analytics_asset_file : array(
```
rtBot, line 149: :warning: **Warning**: File inclusion using variable (`$player_analytics_asset_file`). Probably needs manual inspection (*WordPressVIPMinimum.Files.IncludingFile.UsingVariable*).

rtBot (commented):

**phpcs** scanning turned up:

:warning: 1 warning 


***


subodhr258:
## Code Review

### Overview
Replaces hardcoded script dependency/version arguments in `wp_register_script`/`wp_enqueue_script` with values read from build-generated `*.min.asset.php` files, falling back to `array()` deps + `filemtime()` when the asset file is absent. It also wires `DependencyExtractionWebpackPlugin` (DEWP) into the **`pages`** webpack config and adds the package to `package.json`.

Worth noting: the `assets/build/js/*` bundles already emit `*.min.asset.php` files today, because the shared JS config extends `@wordpress/scripts`' default config (which already bundles DEWP). The genuinely new build output is for `assets/build/pages/*`, which the `pages` config did not run DEWP on before. So the webpack change correctly closes that gap. 👍

The version-from-content-hash change (vs `filemtime()`) is a real improvement: content-based cache busting is stable across deploys/servers, whereas `filemtime()` varies per server and over-invalidates. The `array_merge(...)` calls also correctly preserve cross-script handles DEWP can't infer (`godam-player-frontend-script`, `backbone`, `rtgodam-script`).

### Issues

**1. `pages` config: DEWP + manual `externals` may emit empty dependency arrays — please verify before merge (high impact).**
The `pages` config in `webpack.config.js` manually externalizes `@wordpress/element|i18n|components|api-fetch` (and `react`/`react-dom`). DEWP also performs externalization, and it only writes a handle into the `.asset.php` for modules **it** externalizes — modules already claimed by a manual `externals` entry can be skipped. If that happens here, the generated `assets/build/pages/*.min.asset.php` could contain **no** `wp-element`/`wp-components`/etc.

Because the PHP fallbacks for the page scripts are mostly `array()` (only `video-editor` keeps `['wp-element','wp-i18n']`), a stripped asset file would silently drop `wp-element`/`wp-components` from the admin pages (dashboard, analytics, settings/godam, tools, help, whats-new, media-library) → React/components unavailable → blank/broken screens. Previously these were guaranteed via the hardcoded deps.

Please run the production build and inspect the generated `assets/build/pages/*.min.asset.php` to confirm each page's deps match (or exceed) the prior hardcoded lists. If they don't, remove the now-redundant manual `@wordpress/*` `externals` from the `pages` config and let DEWP own externalization (then re-check `react`/`react-dom` handling).

**2. `jquery` dependency dropped for scripts that use the jQuery global without importing it (medium).**
DEWP only emits `jquery` when jQuery is `import`ed; these scripts use the `$`/`jQuery` global, so the generated asset files omit it. Verified against the already-built `js/*.min.asset.php` plus source grep:

| Handle | Source uses jQuery | Old deps | New effective deps |
|---|---|---|---|
| `godam-http-auth-detector` | yes (`http-auth-detector.js`) | `['jquery']` | `[]` |
| `nf-godam-recorder-upload` | yes (`ninja-forms/index.js`) | `['backbone','jquery','wp-i18n']` | `['wp-i18n','backbone']` |
| `gf-godam-recorder-editor-script` | yes (`gravity-form/gf-godam-recorder-editor.js`) | `['jquery']` | `[]` |

Dropping the explicit `jquery` removes WordPress's guarantee that jQuery is registered and enqueued **before** these footer scripts. In practice jQuery is usually present (admin, or loaded by the forms plugins), so it may not break — but it's a real correctness regression in the dependency declaration. Suggest keeping it via the same merge pattern already used elsewhere, e.g. `array_merge( $asset['dependencies'], array( 'jquery' ) )` for those three handles.

For the record, the other `jquery` removals look **safe** — `everestforms`, `fluentforms`, `elementor` (frontend/editor), `lifterlms`, and `ninja-forms-submissions-list` sources have no jQuery usage, and `godam-player-analytics` losing `wp-i18n` is fine (its source uses no i18n).

**3. `@wordpress/dependency-extraction-webpack-plugin` should be a `devDependency` (minor).**
It's added to `dependencies`, but every other webpack tool here (`@wordpress/scripts`, `css-minimizer-webpack-plugin`, `webpack-remove-empty-scripts`) lives in `devDependencies`. This misplacement is also what produced the large `package-lock.json` churn — dozens of transitive packages lost `"dev": true` because they're now reachable from a production dependency. Moving it to `devDependencies` reverts that lockfile noise and keeps the production tree clean.

### Maintainability (non-blocking)

- **~25× copy-pasted boilerplate** across 15 files. A small helper, e.g. `rtgodam_get_asset_meta( $relative_js_path, $default_deps = array() )` returning `['dependencies' => ..., 'version' => ...]`, would cut most of the +354 lines and centralize the `file_exists`/fallback logic. It would also unify the two styles currently present (the verbose `if ( file_exists )` block vs. the compact ternary in `class-godam-player.php`).
- **Inconsistent fallback deps**: `video-editor` keeps `['wp-element','wp-i18n']` as its fallback while every other page uses `[]`. Given issue #1, a meaningful fallback for all pages is arguably the safer choice — pick one approach consistently.

### Summary
Solid, well-scoped direction and the content-hash versioning is a genuine win. Before merging, please (1) build and verify the generated `pages/*.min.asset.php` actually carry the WP deps, and (2) restore `jquery` for the three handles whose source still uses the jQuery global. Moving the new package to `devDependencies` will also clean up the lockfile diff.

PR #1972, opened 2026-06-19
Video Editor 2.0

KMchaudhary:
Fixes: https://github.com/rtCamp/godam-plugin-wp/issues/3

## Video Editor Redesign

**Scope reminder:** the design board has 10 labeled sections — *Adding a Layer, CTA Layer, Layouts for CTA, Hotspot Layer, Form Layer, Woo Layer (Product Hotspots), Display Settings, Generating Transcription, Uploading Transcription, Chapters*. Note there is **no standalone Poll panel** in the design (Poll appears only as an add‑layer menu entry), and there is **no video‑library/picker screen** anywhere in this PDF.

## 🟢 Extra (in the PR, not in this design) 

- **Transcript tab** — entirely missing. Will be added in a separate PR.

- **Hotspot** —  layer-level "Layer Duration (seconds)" field, generalized as we are supporting multiple hotspots in a single layer.

- **CTA — 3 extra card layouts** beyond the design's 5 (two "Full Height" cover variants + "Image Background"), plus an Advanced "Layer Background Colour" picker.

- **Display Settings — "Adjust Skip Duration" control mismatch.** Design = free‑form time input (`0:00`); code = fixed dropdown 5/10/30s with different help text. (VideoJS only supports skip icons for 5/10/30 sec duration)

- **Chapters — overlap/range validation messages.

- **Add-layer dropdown — Please expect the mismatch between the add layer dropdown icons and the video timeline icons, as some of the non-WordPress icons need SVG code modification for proper fill property rules.

## Demos

1. **Video list view** (Implemented using Gutenberg DataViews components, help in future to provide the list, table video of the same media list)

https://github.com/user-attachments/assets/95ea0029-3fbe-4b89-918e-9acdeba28bd5

2. **Video Editor Loading skeleton**

https://github.com/user-attachments/assets/39639810-fd94-4d80-b0ba-9f8661ce5a26

3. **Video Editor > Settings**
Screenshot 2026-06-25 at 2 19 23 PM

4. **Video Editor > Chapters**
Screenshot 2026-06-25 at 2 20 58 PM

Screenshot 2026-06-25 at 2 21 37 PM

Screenshot 2026-06-25 at 2 22 01 PM


subodhr258 (commented):
## Review — Video Editor 2.0

Big, well-structured redesign. The layer-style abstraction (`hotspotStyle.js` shared between admin preview and frontend player), the new trigger model in `formLayerManager`, and the DataViews list view are clean and well-documented. A few things worth addressing before merge.

### 🔴 Backward compatibility — existing "Text" CTAs

`godam-player.php` drops the `'text' === $godam_layer['cta_type']` render branch, and `TextCTA.js` is deleted. For videos already published with a text CTA:

- **Frontend:** the layer now falls through the `if/elseif` chain and renders an empty box — the saved `text` content silently disappears.
- **Editor:** `CTALayer.js` normalizes `cta_type` to `'image'` (`layer?.cta_type === 'html' ? 'html' : 'image'`) but the stored `text` value is never migrated into the card fields (`imageText`, etc.), so opening + saving an old text CTA loses its content rather than converting it.

Please add a migration (or a frontend fallback that renders the legacy `text` into the card/html template) so existing content doesn't break. If product has decided text CTAs are acceptable to drop, that should be called out explicitly.

### 🟠 REST endpoint — `class-video-editor.php`

- **Author scoping.** `get_videos()` runs a raw `WP_Query` for all `attachment`/`video` posts gated only by `current_user_can( 'upload_files' )`. `upload_files` is granted to Authors, who normally can't see others' media. The core `wp/v2/media` controller scopes results by author for users lacking `edit_others_posts`; this endpoint returns the whole library (titles, URLs, thumbnails) to any author-level user. Worth confirming this matches the intended visibility, and adding `author__in` scoping if not.
- **`build_filter_meta_query()` performance.** The `edited`/`unedited` filters do `LIKE '%"layers";a:%'` (+ `NOT LIKE`/`NOT EXISTS`) on `rtgodam_meta`. These are unindexed full scans of `wp_postmeta` and will get slow on large libraries. The serialized-signature match is also brittle if the meta ever changes serialization (e.g. JSON). The comment acknowledges the trade-off — at minimum flag it for scale; a dedicated boolean meta (`_rtgodam_has_layers`) written on save + backfill would index cleanly.
- **N+1 in the loop.** `prepare_video_item()` calls `wp_prepare_attachment_for_js()` + `get_post_meta()` per post (up to 100). Consider priming caches once before the loop (`update_meta_cache`, `_prime_post_caches`) to cut queries.

### 🟡 Other

- **`analyticsApi` added to the video-editor store** (`redux/store.js`). This pulls the whole analytics API slice into the video-editor bundle — intended? Confirm it's actually used (EditorStatsRow?) and not just bundle weight.
- **Stale doc comment.** `VideoEditorDataView.jsx`'s header says it "drives an infinite scroll via an IntersectionObserver sentinel," but the implementation uses a manual **"Load more"** button (`loadMore`) — no observer. Update the comment.
- **RTK `mutation` for a GET.** `getVideoEditorVideos` models a read as a mutation (acknowledged in the comment). Works, but a `query` with a `merge`/`serializeQueryArgs` for the infinite list would be the idiomatic fit and gets caching/dedup for free.
- **Description:** no **"Automation impact"** line. The PR does add `data-test-id`s (👍), so just add the impact note to the description per the QA convention on godam-plugin-wp#1.

### 🧪 Tests

There are no tests, and this PR introduces several pure, easily-unit-testable units: `build_filter_meta_query()` (PHPUnit), `resolveHotspotStyle()`, `time.js` (`formatClock`/`parseClock`), and the new trigger logic in `formLayerManager` (timestamp/watch_depth/on_pause/end_of_video ordering + one-shot). Given the size and the backward-compat surface here, a few targeted tests would be well worth it.

KMchaudhary:
@subodhr258, the PR feedback has been resolved.

KMchaudhary:
@subodhr258 
Backward compatibility — existing "Text" CTAs **(Intended)**
Resolve other code feedback comments.

subodhr258 (approved):

PR #1976, opened 2026-06-22
feat(godam-pdf): revamp Document block

rudrakshi-gupta:
# feat(godam-pdf): Document Block Revamp

Feat: Doc Block rtCamp/godam-plugin-wp#4

## Summary

Full revamp of the GoDAM Document (PDF) block — new editor UX, multi-tile custom cover picker, client-side page counting, card view meta line (pages + file size), and ESLint fixes.

---

### 1. Editor UX — Empty State & Upload Progress

The editor canvas now shows a purposeful empty state instead of a blank `MediaPlaceholder` when no document is attached.

- Dashed lavender wrapper (`#eef2ff` background, `#a5b4fc` border, `8px` radius) wraps the block in the editor for all states, making it visually distinct from frontend output.
- **Empty state:** centred document icon, "Attach a document here" heading, subtitle, and a full-width "Attach Document" CTA button.
- **Upload progress state:** a card with the file name, file size, and an animated indeterminate progress bar appears while the file is uploading. Replaces the default WordPress spinner.

### 2. Multi-Tile Custom Cover Picker (Inspector)

The inspector's cover section is replaced with a horizontally scrollable tile row.

| Behaviour | Detail |
|---|---|
| **First tile** | Auto-generated PDF thumbnail fetched from `rtgodam_media_pdf_thumbnail` post meta via REST API. Marked with a 4-pointed sparkle SVG badge (bottom-left corner, dark semi-transparent pill). |
| **Upload more** | "Upload cover image" button opens the media library. Users can upload as many tiles as they want — no limit. |
| **Auto-select on upload** | The last-uploaded tile is immediately selected. |
| **Hover → delete** | Each user-uploaded tile shows a `×` delete button outside the top-right corner on hover (dark circle, turns red on hover). |
| **Selected state** | Active tile gets a WordPress blue border (`--wp-admin-theme-color`) and a checkmark overlay (top-right, drop-shadow). |
| **Scrollable** | The row scrolls horizontally with a thin custom scrollbar; tiles don't wrap. |

### 3. REST API: `rtgodam_media_pdf_thumbnail`

`register_post_meta` for `rtgodam_media_pdf_thumbnail` with `show_in_rest: true` was missing. Added to `class-meta-rest-fields.php` so the editor can fetch the auto-generated PDF thumbnail via `useSelect` / `getEntityRecord`.

### 4. Client-Side PDF Page Counting

Page count is derived entirely on the client using `fetch()` on the raw PDF binary — no server-side processing or new meta key required.

**Algorithm:**
1. `fetch(src, { credentials: 'same-origin' })` → `arrayBuffer()`
2. Decode with `TextDecoder('latin1')` (safe for binary PDF streams)
3. Match all `/Count \d+` occurrences; take the **maximum** value (the root page tree node)
4. Fallback: count `/Type /Page` occurrences if no `/Count` found

Works for:
- PDFs uploaded directly (original file)
- PDFs already transcoded by GoDAM (transcoded URL stored in `rtgodam_transcoded_url`)
- Existing PDFs with no transcoding job (uses original URL as fallback)

The count is stored in the `pageCount` block attribute and rendered in the frontend.

### 5. "Show Cover" Toggle

- When the **Show cover** toggle is **off**, no cover image is rendered in either the editor preview or the frontend output — regardless of whether a custom cover or auto-generated thumbnail exists.
- Cover resolution order (when toggle is **on**): `customCover` (user-selected tile URL) → `rtgodam_media_pdf_thumbnail` (auto-generated) → no cover (placeholder icon).

### 6. Card View — Pages & File Size Meta Line

The card view's third line now shows contextual metadata.

```
4 pages • 140 KB
```

| Field | Source |
|---|---|
| Page count | `pageCount` block attribute (set client-side, see §4) |
| File size | `wp_get_attachment_metadata()['filesize']` (stored in DB — works with S3/CDN). Falls back to `filesize(get_attached_file())` for local disk installs. |

Both parts are optional — if neither is available, the filename is shown as a fallback (existing behaviour).

Page count uses `_n()` for correct pluralisation: `1 page` / `4 pages`.

---

## Files Changed

| File | Change |
|---|---|
| `assets/src/blocks/godam-pdf/block.json` | Added `coverImages` (array) and `pageCount` (number) attributes |
| `assets/src/blocks/godam-pdf/edit.js` | Multi-tile cover picker, empty state, upload progress, page count via fetch, effectiveCover logic |
| `assets/src/blocks/godam-pdf/editor.scss` | Dashed wrapper, tile row, tile states (badge, check, delete), empty state, uploading state styles |
| `assets/src/blocks/godam-pdf/style.scss` | Card view mobile breakpoint kept at original (480 px) |
| `assets/src/blocks/godam-pdf/render.php` | Page count + file size meta line, cover gated on `showCover`, `$godam_page_count` attribute |
| `inc/classes/rest-api/class-meta-rest-fields.php` | `register_post_meta` for `rtgodam_media_pdf_thumbnail` with `show_in_rest: true` |
| `package.json` | Added `@wordpress/notices` and `@wordpress/core-data` to devDependencies |
| `.eslintrc` | Added `import/no-unresolved` ignore rule for `^@wordpress/` |

---

### Editor
- [ ] Empty state shows when no document is attached
- [ ] Uploading a document shows the progress card UI
- [ ] After upload, the block renders with the dashed lavender wrapper
- [ ] Inspector → Cover section shows the auto-generated thumbnail tile with sparkle badge
- [ ] Uploading a custom cover image adds a new tile and auto-selects it
- [ ] Clicking a tile selects it (blue border + checkmark)
- [ ] Hovering a user-uploaded tile shows the × delete button; clicking it removes the tile
- [ ] Deleting the selected tile falls back to the auto-generated thumbnail
- [ ] "Show cover" toggle OFF hides cover in editor preview
- [ ] Page count populates after the PDF file is resolved (check block attributes in post editor)

### Frontend — Card View
- [ ] Meta line shows page count and file size when available
- [ ] Meta line shows filename when both are unavailable
- [ ] Cover image hidden when "Show cover" is toggled off
- [ ] Cover image shows correctly for auto-generated and custom covers
- [ ] File size correct for local files and S3-hosted files

---

## Notes

- `@wordpress/core-data` must be installed after merging: run `npm install` to pull the new devDependency.
- The `import/no-unresolved` ESLint ignore for `^@wordpress/` is intentional — the ESLint import resolver cannot resolve `@wordpress/*` package aliases even when they are installed, because they rely on webpack externals rather than standard Node resolution. The packages are correctly declared in `devDependencies` and installed in `node_modules`.
- Page count is **not** re-counted on every render — the `useEffect` guard ensures it only runs when `src` changes (i.e., when a new PDF is attached).
- Caption behaviour is **unchanged**: captions only appear when a document is attached via the GoDAM tab (which pre-populates `media.caption`). Standard media library / direct upload leaves caption empty by design.

## Demo

https://github.com/user-attachments/assets/aba26142-8744-48cd-9830-b56119b8e99e


KMchaudhary (commented):
## Code Review: revamp Document block

### Summary
A large, mostly front-end revamp of the `godam-pdf` block: new editor empty/upload states, a multi-tile cover picker, default/card preview modes, a client-side PDF page counter, and a card-view meta line. The PHP/REST surface is small, well-escaped, and consistent with existing conventions. The main concerns are in the editor React layer (invalid HTML nesting, a removed interaction guard) and in the page-counting approach, which is expensive and silently fails for the offloaded-media case the PR claims to support.

### Critical / High Issues

| # | File | Line | Issue | Severity |
|---|------|------|-------|----------|
| 1 | `assets/src/blocks/godam-pdf/edit.js` | ~616–645 | **Nested `<button>` inside `<button>`.** Each user-uploaded cover tile is a `<button>` that contains the delete `<button>`. This is invalid HTML — the browser reparents the inner button out of the outer one, triggering React's `validateDOMNesting` warning and making the delete click / `stopPropagation` behavior unreliable. | 🔴 High |
| 2 | `assets/src/blocks/godam-pdf/edit.js` | 163–205 | **Whole-PDF download on every editor open + silent CORS failure.** The `useEffect` does `fetch(src) → arrayBuffer()` on the *entire* PDF every time `src` is set (i.e. every editor load), then regex-scans it. For large PDFs this is a big download + full-file allocation in the editor. Worse: GoDAM offloads media to cloud/CDN, so `src` is typically cross-origin; `fetch` is CORS-blocked and the `.catch(() => {})` swallows it → `pageCount` stays `0`. So the feature silently does **not** work for the S3/CDN case the PR description says it supports. | 🔴 High |

### Suggestions

| 3 | `assets/src/blocks/godam-pdf/edit.js` | 411–424 | The default-view embed lost its old `<Disabled isDisabled={!isSingleSelected}>` wrapper. The embedded PDF now captures clicks/scroll in the editor, making the block hard to select. Restore the `Disabled` wrapper (or an overlay) when the block isn't selected. | Correctness/UX |
| 4 | `edit.js`:432 vs `render.php`:1265 | — | Editor meta line says **"N slides"**; frontend says **"N page/pages"** (`_n`). Same data, different label. Pick one (the PR description even shows "4 pages"). | Correctness/UX |
| 5 | `assets/src/blocks/godam-pdf/edit.js` | 163–205 | Even if CORS is fixed, guard the effect with `if ( pageCount > 0 ) { return; }` so it doesn't re-download already-counted files on every open (and doesn't mark a clean post dirty by re-`setAttributes` on load). Consider an `AbortController` instead of just the `cancelled` flag so the in-flight download is actually aborted on unmount. Better still: compute page count server-side at upload/transcode time and store it in meta. | Performance |
| 6 | `package.json` / `package-lock.json` | — | `@wordpress/core-data` is pinned at `7.48.1` while the rest of the WP packages are an older wave (`@wordpress/data 10.42.0`, `element 6.39.0`, `components 30.7.0`). This drags a whole newer nested tree (components 35, element 8, data 10.48) into the lockfile and risks linting/types against APIs newer than the runtime provides. Pin core-data (and notices) to the version matching the project's existing WP wave. | Maintainability |
| 7 | `assets/src/blocks/godam-pdf/edit.js` | 360–361 | Empty-state copy: "You can upload a pdf, **ppt or any document**" but `ALLOWED_MEDIA_TYPES = ['application/pdf']` — only PDFs are accepted. Misleading; align copy with what's allowed. | UX/Copy |
| 8 | `assets/src/blocks/godam-pdf/render.php` | 1165, 1221, 1237 | `customCover` / thumbnail are `esc_url`'d on assignment and then `esc_url`'d again at output (`src="<?php echo esc_url( $godam_cover_url ); ?>"`). Harmless but redundant — store raw, escape once at output. | Maintainability |
| 9 | `assets/src/blocks/godam-pdf/edit.js` | 616–622 | Uploaded cover tiles render `` with no `aria-label` on the selecting button → screen readers announce a nameless "button". Add an accessible name (e.g. `aria-label={ __( 'Select cover', 'godam' ) }`). | Accessibility |
| 10 | `assets/src/blocks/godam-pdf/edit.js` | — | The revamp drops all `data-test-id` hooks (`godam-pdf-canvas`, `…-panel-settings`, etc.). No test references them in this repo, so nothing breaks today — just confirm there's no external e2e suite relying on them. | Testing |

### Suggested fixes for the top two

**#1 — un-nest the delete button** (make the tile a non-button clickable so the delete `<button>` is a valid child):

```jsx
<div
    role="button"
    tabIndex={ 0 }
    className={ `godam-pdf-cover-tile${ customCover === cover.url ? ' is-selected' : '' }` }
    onClick={ () => onSelectTile( cover.url, cover.id ) }
    onKeyDown={ ( e ) => ( e.key === 'Enter' || e.key === ' ' ) && onSelectTile( cover.url, cover.id ) }
>
    <img src={ cover.url } alt="" />
    { /* check overlay */ }
    <button
        type="button"
        className="godam-pdf-cover-tile__delete"
        aria-label={ __( 'Remove cover', 'godam' ) }
        onClick={ ( e ) => { e.stopPropagation(); onDeleteCoverImage( i ); } }
    >…</button>
</div>
```

**#2 — make page counting tolerant and cheap.** Cross-origin `fetch` won't work for offloaded media regardless of code, so the robust answer is to count pages server-side during upload/transcode and expose it via meta (similar to `rtgodam_media_pdf_thumbnail`). If you keep the client approach short-term, at minimum guard against re-running and document that it only works for same-origin, uncompressed PDFs — note that PDF 1.5+ object streams compress the page tree, so the `/Count` and `/Type /Page` regexes miss many modern PDFs even same-origin.

### What Looks Good
- **PHP output is well-escaped** throughout `render.php` (`esc_url`, `esc_html`, `esc_attr`, `wp_json_encode`+`esc_attr` for `data-sources`); inputs sanitized with the right helpers (`sanitize_text_field`, `sanitize_textarea_field`). No injection surface introduced.
- New `register_post_meta` for `rtgodam_media_pdf_thumbnail` mirrors the existing `rtgodam_media_video_thumbnail` registration exactly (`auth_callback` = `current_user_can('edit_posts')`) — correctly scopes write access while exposing a non-sensitive thumbnail URL.
- File-size resolution via `wp_get_attachment_metadata()['filesize']` with a `get_attached_file()` disk fallback is the right pattern for S3/CDN setups.
- Correct `_n()` pluralization on the frontend, and graceful degradation (meta line → filename) when data is missing.
- Card link uses `target="_blank" rel="noopener noreferrer"`.

### Verdict
**Request Changes** (advisory — posted as a comment). No security blockers, but issues #1 (invalid DOM that affects the delete UX) and #2 (page counter that's expensive and silently no-ops for cloud-hosted media — the product's normal case) should be addressed before merge, along with the editor `Disabled`-wrapper regression (#3) and the slides/pages label mismatch (#4).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

KMchaudhary:
@rudrakshi-gupta please don't allow user to upload non-PDF files, show appropiate message.
Screenshot 2026-06-24 at 11 15 04 AM

KMchaudhary (approved):

PR #1979, opened 2026-06-23
fix(gallery-v2): hover play, first-frame thumbs, play button toggle, sequential autoplay, load-more parity

rudrakshi-gupta:
# Video Gallery V2 — Feature & Bug Fix Batch

Fixes rtCamp/godam-plugin-wp#4

## What changed and why

**1. Fix hover/preview play for transcoded videos**
`rtgodam_transcoded_url` stores an MPEG-DASH `.mpd` manifest that a plain `<video>` element cannot decode. Switched `video_url` in `rtgodam_gallery_v2_get_video_data()` to always use `wp_get_attachment_url()` (the original MP4), which is universally browser-playable and sufficient for the muted hover preview.

**2. First-frame video thumbnail (no-thumbnail fallback)**
When a video has no manually set thumbnail, the frontend now shows the video's first frame — matching the shoppable video block's behaviour. This is done by:
- Emitting a `` in the PHP template when no thumbnail URL exists
- In `view.js`, `initFirstFrameThumbnails()` detects this placeholder, hides it, sets `preload="metadata"` + `video.load()` on the preview video, and adds a `--no-thumb` CSS class that keeps the video element permanently visible
- A CSS rule for `--no-thumb` items sets `opacity: 1` on the preview video so the browser-rendered first frame is visible as the thumbnail

**3. Fix thumbnail CSS opacity bug**
The global `opacity: 0` rule on `.godam-gallery-v2__thumbnail` was preventing thumbnails from rendering for items that don't go through the blur-up placeholder flow. Removed the global rule; the blur-up system manages its own opacity transition independently.

**4. Editor: default placeholder for videos without a thumbnail**
In the block editor, a tile with a video selected but no thumbnail was incorrectly showing the "Select Video" empty state. Added a distinct `--no-thumb` placeholder state that shows the video icon with Replace/Remove actions, making it clear a video is attached.

**5. Show Play Button toggle**
Added a `showPlayButton` boolean attribute (default `true`) and a `ToggleControl` in the Interaction panel of the block inspector. When off:
- PHP template passes `data-show-play-button="false"` on the gallery wrapper
- CSS `[data-show-play-button="false"]` hides `.godam-gallery-v2__play-icon` and `.godam-gallery-v2-item__play-icon` via `display: none`
- Editor query-tab preview was also fixed — the play icon SVG was missing from the React JSX for query preview tiles, so it's now rendered conditionally based on the `showPlayButton` attribute

**6. Sequential autoplay (one video at a time, full playback)**
The `<video>` elements had the `loop` HTML attribute, which prevents the `ended` event from ever firing. The existing `advanceAutoplaySequence` handler was listening for `ended` but never received it, so autoplay never advanced. Fixed by:
- Removing `loop` from both `<video>` elements in the PHP template
- Setting `video.loop = false` in `playAutoplayItem()` before play, ensuring `ended` always fires
- Setting `video.loop = true` in the hover-play path (`handleItemHoverStart`) so hovered videos still loop continuously

**7. Load-more / infinite scroll parity**
New tiles injected by Load More or infinite scroll were missing all interactive behaviour because two separate issues compounded:
- `class-dynamic-gallery.php` (the REST endpoint for load-more HTML) had its own stripped-down tile template that omitted the `<video>` element, play icon SVG, and `--pending` thumbnail — so the required DOM nodes were simply absent
- `initVideoPreview()` only ran once at construction and never re-ran for new items

Both are fixed:
- REST endpoint now renders the full tile structure (video element, play icon, `--pending` img) matching `godam-video-gallery.php`
- Introduced `initializedPreviewItems` WeakSet to track wired-up tiles
- Extracted `initVideoPreviewItem(item)` — idempotent per-item setup (event listeners + `initItemVideoState`)
- `initViewportPreload()` and `initAutoplayObserver()` now create their observers once and accept an optional item list to observe, so new items can be registered without re-creating the observers
- New `initNewItemsBehavior()` method called at the end of `loadMoreItems()` — finds uninitialised items, wires them up, and registers them with existing observers

## Files changed
- `inc/templates/godam-video-gallery.php` — video_url fix, `--pending` thumbnail markup, `loop` removal, `data-show-play-button` attribute
- `inc/classes/rest-api/class-dynamic-gallery.php` — full tile HTML parity (video, play icon, pending thumbnail)
- `assets/src/blocks/godam-gallery-v2/view.js` — first-frame thumbnails, sequential autoplay, load-more initialisation, viewport preload refactor
- `assets/src/blocks/godam-gallery-v2/edit.js` — show play button toggle, play icon in query preview JSX, `data-show-play-button` on canvas wrappers
- `assets/src/blocks/godam-gallery-v2/block.json` — `showPlayButton` attribute
- `assets/src/blocks/godam-gallery-v2/style.scss` — `--no-thumb` opacity, play button hide rule, thumbnail opacity fix
- `assets/src/blocks/godam-gallery-v2-item/edit.js` — no-thumbnail placeholder state in editor

## Demo

https://github.com/user-attachments/assets/50532b42-95ba-4881-a759-f3658bebb4d3

### DIFF
--- a/assets/src/blocks/godam-gallery-v2-item/edit.js
+++ b/assets/src/blocks/godam-gallery-v2-item/edit.js
@@ -220,6 +220,50 @@ export default function Edit( { attributes, setAttributes, context, clientId } )
 				</div>
 			</>
 		);
+	} else if ( videoId ) {
+		// Video selected but no thumbnail yet — show a generic video placeholder
+		// with the title so the editor tile looks populated, not empty.
+		previewContent = (
+			<>
+				<div className="godam-gallery-v2-item__placeholder godam-gallery-v2-item__placeholder--no-thumb">
+					{ videoIcon }
+				</div>
+				<div className="godam-gallery-v2-item__preview-overlay">
+					<MediaUploadCheck>
+						<MediaUpload
+							onSelect={ onSelectVideo }
+							allowedTypes={ [ 'video' ] }
+							value={ videoId }
+							render={ ( { open: openMediaModal } ) => (
+								<Button
+									variant="secondary"
+									icon={ pencil }
+									className="godam-gallery-v2-item__overlay-action"
+									onClick={ ( event ) => {
+										event.stopPropagation();
+										openMediaModal();
+									} }
+								>
+									{ __( 'Replace', 'godam' ) }
+								</Button>
+							) }
+						/>
+					</MediaUploadCheck>
+					<Button
+						variant="secondary"
+						icon={ closeSmall }
+						isDestructive
+						className="godam-gallery-v2-item__overlay-action"
+						onClick={ ( event ) => {
+							event.stopPropagation();
+							removeBlock( clientId );
+						} }
+					>
+						{ __( 'Remove', 'godam' ) }
+					</Button>
+				</div>
+			</>
+		);
 	} else {
 		previewContent = (
 			<div className="godam-gallery-v2-item__placeholder">
--- a/assets/src/blocks/godam-gallery-v2/block.json
+++ b/assets/src/blocks/godam-gallery-v2/block.json
@@ -79,6 +79,18 @@
 		"engagements": {
 			"type": "boolean",
 			"default": true
+		},
+		"autoplay": {
+			"type": "boolean",
+			"default": false
+		},
+		"playOnHover": {
+			"type": "boolean",
+			"default": true
+		},
+		"showPlayButton": {
+			"type": "boolean",
+			"default": true
 		}
 	},
 	"providesContext": {
--- a/assets/src/blocks/godam-gallery-v2/edit.js
+++ b/assets/src/blocks/godam-gallery-v2/edit.js
@@ -13,6 +13,7 @@ import {
 } from '@wordpress/block-editor';
 import {
 	PanelBody,
+	RadioControl,
 	RangeControl,
 	SelectControl,
 	ToggleControl,
@@ -21,16 +22,34 @@ import {
 	Popover,
 	Notice,
 	Spinner,
+	Button,
 	__experimentalToggleGroupControl as ToggleGroupControl,
 	__experimentalToggleGroupControlOption as ToggleGroupControlOption,
 	__experimentalToggleGroupControlOptionIcon as ToggleGroupControlOptionIcon,
-	Button,
 } from '@wordpress/components';
 import { useDispatch, useSelect, select as dataSelect } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as noticesStore } from '@wordpress/notices';
 import { useCallback, useEffect, useMemo, useRef, useState } from '@wordpress/element';
-import { columns, grid, listView, plus } from '@wordpress/icons';
+import { columns, grid, listView, plus, trash } from '@wordpress/icons';
+
+/**
+ * External dependencies
+ */
+import {
+	DndContext,
+	closestCenter,
+	PointerSensor,
+	useSensor,
+	useSensors,
+	DragOverlay,
+} from '@dnd-kit/core';
+import {
+	SortableContext,
+	verticalListSortingStrategy,
+	useSortable,
+} from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
 
 /**
  * Internal dependencies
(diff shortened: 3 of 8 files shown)


KMchaudhary (commented):
## Code Review: fix(gallery-v2): hover play, first-frame thumbs, play button toggle, sequential autoplay, load-more parity

### Summary
A well-structured batch of targeted fixes. The root-cause analysis in the PR description is accurate (DASH URL unusable in `<video>`, `loop` blocking `ended`, REST endpoint missing DOM nodes), and the load-more parity approach (WeakSet guard + idempotent per-item init) is solid. A few correctness and accessibility gaps are worth addressing before merge.

---

### Critical Issues

_None._

---

### Suggestions

| # | File | Area | Issue | Severity |
|---|------|------|-------|----------|
| 1 | `view.js` | `initAutoplayObserver` / `syncAutoplaySequence` | **Accessibility — `prefers-reduced-motion` not respected.** Autoplaying video for users who set this preference violates WCAG 2.2.2 (Pause, Stop, Hide). Add a guard at the top of `initVideoPreview()` (and skip setting up the autoplay observer): `if (window.matchMedia?.('(prefers-reduced-motion: reduce)').matches) return;` | 🟠 High |
| 2 | `view.js` | `syncPreviewVideo` | **Stale listener accumulation.** When `syncPreviewVideo(item, true)` is called multiple times before the video loads (rapid hover in/out), each call registers a new `{ once: true }` listener for both `loadeddata` and `canplay`. The state guard inside `attemptPlay` prevents ghost plays after the hover ends, but in autoplay mode the item can become active again later and all accumulated closures fire simultaneously. Fix by using an `AbortController` per item (or a `_pendingPlay` flag on the video) to cancel stale callbacks. | 🟡 Medium |
| 3 | `block.json` + `edit.js` | `playOnHover` attribute | **Redundant stored state.** The UI always sets `playOnHover = !autoplay`, so `playOnHover` is fully derivable. If a migration ever sets one without the other (e.g., importing pre-existing block JSON), the two attributes can silently disagree and `view.js` reads both independently. Either remove `playOnHover` from `block.json` and derive it in `view.js`, or add a `deprecated` migration to enforce the invariant. | 🟡 Medium |
| 4 | `edit.js` | Interaction panel radio group | **Accessibility — custom radio wrapper.** The `eslint-disable-next-line jsx-a11y/label-has-associated-control` suppresses a valid warning. The `<label>` wraps the `<input>` so association is correct, but the `<input>` is flex-sibling to `<span>` elements (label/hint text), not a child. Screen readers announce the full text of the wrapping label, so it works, but replacing the custom `<div class="godam-gallery-v2__radio-group">` pattern with WP's `RadioControl` (already imported) would remove the a11y suppression and be consistent with the rest of the inspector. | 🟡 Medium |
| 5 | `editor.scss` | `.godam-gallery-v2__canvas` block (line ~247) | **Mixed indentation.** The new rule uses spaces for `margin`, `border`, and `padding` declarations while the rest of the file (and this block itself) uses tabs. Run through Prettier/EditorConfig before merge. | 🔵 Low |
| 6 | `editor.scss` | Color values throughout new rules | **Hardcoded hex colors.** New rules use `#007cba`, `#1e1e1e`, `#757575`, `#F2F2FF`, `#5A13D51A`, etc. WP admin colour scheme can change these (e.g. high-contrast mode). Prefer `var(--wp-admin-theme-color)`, `var(--wp-components-color-foreground)`, `var(--wp-components-color-foreground-muted)` where equivalents exist. | 🔵 Low |
| 7 | `view.js` | `initFirstFrameThumbnails` | **Scans all items on every `loadMoreItems` call.** The method is idempotent so it's safe, but it iterates the full `this.items` array after each load-more append. Scope it to `uninitialised` items the same way `initNewItemsBehavior` does, or merge the logic there. | 🔵 Low |

---

### What Looks Good

- **Root-cause fix for `loop` / `ended`**: Correctly identified that `loop` attribute suppresses the `ended` event. Setting `video.loop = false` before play in `playAutoplayItem` and restoring `video.loop = true` for hover is clean and minimal.
- **WeakSet idempotency guard**: `initializedPreviewItems` WeakSet is the right data structure — entries are GC'd if a tile is removed from DOM, so re-inserted tiles are correctly re-initialised.
- **`wp_get_attachment_url()` for preview video**: Good call. The DASH `.mpd` URL would silently fail in any browser without video.js. The original MP4 is universally playable for the muted preview use-case.
- **REST endpoint HTML parity**: The load-more endpoint now emits the full tile DOM (video, play icon, `--pending` img), fixing a structural mismatch that was the root cause of load-more items missing all interactivity.
- **`data-show-play-button` attribute + CSS approach**: Attribute + descendant CSS selector is the right pattern for toggling block-level state in both the editor canvas and the frontend.
- **`initViewportPreload` / `initAutoplayObserver` observer reuse**: Creating observers once and passing an optional item list to observe is much better than tearing down and recreating observers on each load-more call.

---

### Verdict

**Request Changes** — items 1 (reduced-motion) and 2 (listener accumulation) are worth fixing before merge; the rest can be follow-ups if preferred.

KMchaudhary:
@rudrakshi-gupta tested and changes LGTM. Please take a look at the provided Claude review. Address the comments the comments that you feels necessary.

KMchaudhary (approved):

PR #1987, opened 2026-06-26
Feat/video editor transcript

KMchaudhary:
This pull request introduces a new AI-powered transcription feature to the video editor, adding both backend REST API support and a new frontend "Transcription" tab. Users can now generate, upload, and manage video transcriptions directly within the editor interface. The main changes are grouped below:

**Backend: REST API for Transcription**
- Added a new `Transcription` REST API class (`class-transcription.php`) that supports generating AI transcriptions, uploading caption files, fetching current transcription status, and deleting transcriptions for video attachments. This ensures secure handling of SaaS credentials and transcript metadata.
- Registered the new `Transcription` REST API in the plugin's main loader to ensure endpoints are available. [[1]](diffhunk://#diff-0beca18a469ce0ef11130f3a94b2f7f6997eecfa5f4f414442e163a0230e131cR41) [[2]](diffhunk://#diff-0beca18a469ce0ef11130f3a94b2f7f6997eecfa5f4f414442e163a0230e131cR186)

**Frontend: Video Editor Integration**
- Introduced a new "Transcription" tab in the video editor UI, including the tab icon and navigation logic, so users can access transcription features alongside layers, settings, and chapters. [[1]](diffhunk://#diff-2d227a735881884c63071a3e5d58d3056564026d18f0a0a4379c0ca4c70d33e8L10-R17) [[2]](diffhunk://#diff-2d227a735881884c63071a3e5d58d3056564026d18f0a0a4379c0ca4c70d33e8R28)
- Integrated the `Transcription` React component into the editor, passing necessary props such as `attachmentID`, `duration`, and formatted file size. [[1]](diffhunk://#diff-0559daddab966567cf8d9f589b35d6c35f4a44b62c60586606523dd0cbcb3494R44-R45) [[2]](diffhunk://#diff-0559daddab966567cf8d9f589b35d6c35f4a44b62c60586606523dd0cbcb3494R429-R435)

**Styling**
- Added a dedicated SCSS file (`_transcription.scss`) to style the new transcription tab, including layouts for empty, busy, and ready states, as well as cue list presentation.

## Demo

https://github.com/user-attachments/assets/b8e1456e-ffb6-4570-9f66-741648e9aa0b

### DIFF
--- a/inc/classes/class-plugin.php
+++ b/inc/classes/class-plugin.php
@@ -38,6 +38,7 @@ use RTGODAM\Inc\REST_API\Media_Library;
 use RTGODAM\Inc\REST_API\Video_Editor;
 use RTGODAM\Inc\REST_API\Ads;
 use RTGODAM\Inc\REST_API\Transcoding;
+use RTGODAM\Inc\REST_API\Transcription;
 use RTGODAM\Inc\REST_API\Analytics;
 use RTGODAM\Inc\REST_API\Polls;
 use RTGODAM\Inc\REST_API\Dynamic_Shortcode;
@@ -182,6 +183,7 @@ class Plugin {
 		Video_Editor::get_instance();
 		Ads::get_instance();
 		Transcoding::get_instance();
+		Transcription::get_instance();
 		Analytics::get_instance();
 		Deactivation::get_instance();
 		Polls::get_instance();
--- a/inc/classes/rest-api/class-transcription.php
+++ b/inc/classes/rest-api/class-transcription.php
@@ -34,6 +34,16 @@ class Transcription extends Base {
 	const STATUS_META = 'rtgodam_transcript_status';
 
 	/**
+	 * Meta flag set when the user deletes the transcript. While set, the GET
+	 * endpoint will NOT auto-resurrect the transcript from the SaaS (otherwise
+	 * `godam_get_transcript_path()` re-caches it and the delete is undone). It
+	 * is cleared on generate / upload so a fresh transcript can replace it.
+	 *
+	 * @var string
+	 */
+	const DELETED_META = 'rtgodam_transcript_deleted';
+
+	/**
 	 * Register custom REST API.
 	 *
 	 * @return array Array of registered REST API routes.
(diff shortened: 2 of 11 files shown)


subodhr258:
## Code Review

Solid, well-documented PR. The same-origin proxy pattern (keeping the SaaS key server-side) and the delete-marker design are both correctly thought through, and the comments explain the *why* well. A few things worth addressing before merge.

### 🟠 Should fix

**1. Busy state isn't restored on remount (real UX gap)**
`isGenerating` is local component state set only inside `handleGenerate`. The server *does* persist the in-flight status (`STATUS_META = 'Transcribing'`), but nothing reads it on mount:
- User clicks Generate → busy UI + 8s polling.
- User switches tab / reloads while the 10–15 min job runs → `isGenerating` resets to `false`, `transcriptPath` is still empty, so the panel falls back to the **empty "Generate / Upload" state and stops polling**.

Recommend seeding the busy state from the server status, e.g. derive it instead of relying solely on local state:
```js
const serverBusy = isTranscribingStatus( status );
const busy = isGenerating || serverBusy;
```
and drive `pollingInterval` + the progress UI off `busy`. Otherwise a returning user sees a "Generate" button for a job that's already running, and clicking it re-triggers the SaaS.

**2. Dead `file_size` fallback**
`Transcription.js` uses `formatBytes( transcription?.file_size )` as a fallback, but `shape()` in `class-transcription.php` never returns `file_size` — so the fallback is always `''`. Either add `file_size` to the PHP `shape()` response or drop the fallback to avoid implying data that never arrives.

### 🟡 Consider

**3. 15s blocking remote request**
`generate_transcription()` uses `'timeout' => 15` with a VIP phpcs:ignore. The existing `godam_get_transcript_path()` uses 3s. 15s holds the PHP/REST worker for the full duration on a slow SaaS. Since generation is async anyway (you poll for completion), the kickoff call shouldn't need to block 15s — a shorter timeout (≤5s) would be safer and closer to the VIP guideline. If 15s is genuinely required, the ignore comment should say why.

**4. Request encoding inconsistency**
`generate_transcription()` sends a form-encoded array `body`, while the sibling `godam_get_transcript_path()` POSTs JSON with an explicit `Content-Type` header to the same Frappe API base. Frappe usually accepts both, but worth confirming `process_transcription` does, and aligning the two for consistency.

**5. No unit tests for the pure helpers**
`parseCaptions`, `parseTimestamp`, `formatClock`, `formatBytes` are pure and trivially testable (VTT vs SRT comma variant, `MM:SS` vs `HH:MM:SS`, cue settings after the end timestamp, malformed blocks). These are exactly the kind of parsing logic that regresses silently — a small jest suite would lock in the behavior.

### 🟢 Looks good
- `edit_permission` correctly gates on `edit_post` for the attachment (falls back to `upload_files` only when no id). Nonce via `window.videoData.nonce` matches `attachment.js`. ✅
- Delete-marker (`DELETED_META`) preventing SaaS auto-resurrection, cleared on generate/upload — correct and well-commented. ✅
- Upload validates extension on both client and server (`.vtt`/`.srt`), `esc_url_raw` applied. ✅
- SCSS scoped under `.godam-video-editor`, accent pinned via `--wp-components-color-accent` / `--wp-admin-theme-color` per project theming convention. ✅
- `setCurrentTab` allow-list refactor is cleaner than the old `||` chain. ✅
- WP components (`Button`, `Modal`, `Spinner`, `Notice`) used throughout; strings translated with the `godam` text domain. ✅

### Minor
- `getTranscription` polls every 8s, and each poll re-hits the SaaS (3s) whenever the path is empty — ~110 SaaS calls per open tab over a 15-min job. Acceptable for completion detection, but worth being aware of if multiple editors poll at once.
- The empty-state "OR" divider renders even while generating (the upload button is replaced by the progress block but the divider stays) — minor visual nit, verify against the Figma.

KMchaudhary:
Good Recommendations
Todo: We will improve the transcription status in the future release.

subodhr258 (approved):

PR #1988, opened 2026-06-26
feat(video-editor): add onboarding product guide

KMchaudhary:
Add a driver.js-based guided tour that walks first-time users through making a video interactive (open video → pick timeline spot → add CTA layer → configure → save → copy), with a welcome/skip flow and a See how it works relaunch. Guide state is persisted per-user via a new onboarding REST endpoint.

## **Extra changes added**

Screenshot 2026-06-26 at 5 30 06 PM
Screenshot 2026-06-26 at 5 32 21 PM
Added an extra step of selecting the time where you want to add a new layer.

## Demo

https://github.com/user-attachments/assets/8b79e46d-d6d6-4865-b262-d67cf58e2c94


subodhr258:
## Code Review

**Summary:** While the product guide is running, opening the wp.media image picker in `CardCTA` would put the driver.js overlay/popover (z-index 1e9) on top of the modal and swallow its clicks. This suspends the guide visuals when the picker opens and resumes the current step when it closes. A small, well-scoped fix.

### What works well
- Reuses the established `teardownVisuals()` → `resume()` pattern already used by the "end guide?" confirm flow (`onCloseClick`), rather than inventing new lifecycle handling. Logical state (`active`/`index`) is untouched, so resume re-highlights the same step.
- I verified the picker sits inside the `config-panel` step, which advances via its **Next** button (`showNext`), not via image selection — so resuming the *same* step after the picker closes is the correct behavior, not a missed `notify()`.
- `fileFrame` is created fresh per `openImageCTAUploader()` call, so the `close` listener does not accumulate across invocations — no leak.
- Binding on `close` (rather than `select`) correctly handles both the select and cancel paths.
- Good inline comments explaining the *why* (z-index conflict).

### Minor notes (non-blocking)
- **No automated test.** Per the team convention of adding tests for logic-bearing changes, a small jest test (guide active → `openImageCTAUploader` calls `suspend`, and a synthesized `close` calls `resume`) would lock in the behavior. The wp.media frame would need mocking. Reasonable to defer given it's UI glue, but worth flagging.
- **Redundant-looking guard:** `isGuideActive()` is checked before `suspend()`, and `suspend()`/`resume()` re-check `active` internally. This is fine and actually necessary here — the outer guard gates whether the `close` listener gets registered at all — just calling it out so it's not "simplified" away later.
- **Scope:** This is the only media picker on the guide's happy path (Card CTA config), so covering just `CardCTA` is correct. If a future guide step ever opens wp.media from another layer type, it'd need the same suspend/resume wrapping.

### Risk
Low. No state-machine changes, no persistence changes, mirrors an existing working pattern. LGTM.

subodhr258 (approved):

PR #1998, opened 2026-07-01
fix: install plugins via core Plugin_Upgrader to satisfy Plugin Check

elifvish:
The Frappe Dispatch installer hand-rolled the download/extract/install of a plugin using unzip_file() and copy_dir() into WP_PLUGIN_DIR, which Plugin Check flags ("Plugin folders are deleted when upgraded. Do not save data to the plugin folder").

Replace the manual flow with WordPress core's Plugin_Upgrader::install() (quiet Automatic_Upgrader_Skin). Core now performs the download, extraction and install through its sanctioned upgrader, so the plugin no longer writes into the plugins folder from its own code. An upgrader_source_selection filter renames the extracted folder to the resolved slug (inside the upgrader's working dir) so plugin_exists()/activate_plugin() still locate it.

Removes now-dead helpers (extract_plugin, get_request, DOWNLOAD_TIMEOUT, load_wordpress_filesystem_dependencies, directory listing helpers).

### DIFF
--- a/lib/class-godam-frappe-dispatch-installer.php
+++ b/lib/class-godam-frappe-dispatch-installer.php
@@ -45,13 +45,6 @@ class Godam_Frappe_Dispatch_Installer {
 	const REQUEST_TIMEOUT = 30;
 
 	/**
-	 * HTTP timeout for downloads.
-	 *
-	 * @var int
-	 */
-	const DOWNLOAD_TIMEOUT = 60;
-
-	/**
 	 * Constructor.
 	 *
 	 * @param string $api_url Base URL for Frappe Dispatch.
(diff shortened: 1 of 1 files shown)


lib/class-godam-frappe-dispatch-installer.php
```diff
@@ -203,175 +196,85 @@ private function download_and_install_plugin( $plugin_data, $plugin_slug ) {
...
-			);
-		}
-
-		// Create a temporary extraction directory.
-		$temp_extract_dir = $plugins_dir . DIRECTORY_SEPARATOR . $plugin_slug . '_temp_' . wp_generate_password( 8, false );
-
-		// Extract zip file to temporary directory.
-		$unzip_result = unzip_file( $zip_file, $temp_extract_dir );
+	private function get_source_rename_callback( $desired_slug ) {
+		return function ( $source, $remote_source, $upgrader, $hook_extra = array() ) use ( $desired_slug ) {
```
rtBot, line 257: :warning: **Warning**: The method parameter $upgrader is never used (*Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed*).

rtBot, line 257: :warning: **Warning**: The method parameter $hook_extra is never used (*Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed*).

rtBot (commented):

**phpcs** scanning turned up:

:warning: 2 warnings 


***


subodhr258:
## Review

Solid, well-scoped change: it swaps the hand-rolled `download_url` + `unzip_file` + `copy_dir` flow for core's `Plugin_Upgrader::install()` with a quiet `Automatic_Upgrader_Skin`, and uses an `upgrader_source_selection` filter to force the extracted folder to the resolved slug so `plugin_exists()`/`activate_plugin()` still find it. The success detection (`is_wp_error()` then `true !== $result`), dependency loading via `class-wp-upgrader.php`, and filter add/remove all look correct. A couple of behavioral edges worth a look:

**1. `lib/class-godam-frappe-dispatch-installer.php:220` — SSL-verify opt-out silently dropped for the download.**
The old download used `sslverify => apply_filters( 'godam_frappe_dispatch_verify_ssl', true )`. Core's `Plugin_Upgrader::install()` → `download_url()` always verifies SSL, so a site that filtered `godam_frappe_dispatch_verify_ssl` to `false` (e.g. a self-hosted Dispatch server with a self-signed cert) will now fail the download. The filter still governs the metadata `post_request()`, so the two paths behave inconsistently. If self-signed Dispatch backends are a supported deployment, this is a regression; otherwise worth a note in the PR body.

**2. `lib/class-godam-frappe-dispatch-installer.php:255` — when the rename `move()` fails, install still reports success under the wrong path.**
In the source-rename callback, if `$wp_filesystem->move()` returns `false` the callback returns the original `$source`, so core installs the plugin under the archive's native top-level folder name. But `download_and_install_plugin()` still returns `'success' => true` with `path => WP_PLUGIN_DIR . '/' . $plugin_slug` (a directory that does not exist), and a subsequent `activate_plugin( $install_slug )` finds no PHP files and silently returns `false`. The caller can't distinguish "installed + activated" from "installed under a different name + activation no-op." Consider verifying the destination folder exists before returning success, or surfacing the activation failure.

**3. `lib/class-godam-frappe-dispatch-installer.php:257` (minor) — self-move on a flat archive.**
If the archive has no single top-level folder, core leaves `$source === $remote_source`; the callback then attempts `move( $remote_source, $remote_source . '/' . $desired_slug )` (moving a directory into its own child), which `rename()` rejects. Low likelihood since plugin zips ship foldered, but the callback could guard against `$source` being the working root.

None of these block merge; #2 is the one most likely to bite in practice (a failed rename masquerading as a successful install).

KMchaudhary (approved):

PR #2003, opened 2026-07-03
Fix the demo video MPD source mapping

KMchaudhary:
This pull request makes a small but important correction to how demo video asset URLs are handled in the `class-demo-assets.php` file. The change ensures that the correct MPD (MPEG-DASH) URL is used for demo video assets, improving the accuracy of asset metadata.

- Corrected the assignment of the `mpd_url` field in the asset data array to use the `transcoded_file_path` from the payload, rather than incorrectly using the MP4 URL. [[1]](diffhunk://#diff-4a06d33c38ee5fbe292d182a8f02ef33a3c25d4fd910b49f5a56e32738f084d3R194) [[2]](diffhunk://#diff-4a06d33c38ee5fbe292d182a8f02ef33a3c25d4fd910b49f5a56e32738f084d3L202-R203)

## Demo

https://github.com/user-attachments/assets/535ab6b1-496a-46f1-8a35-77e6ea484cdb

### DIFF
--- a/inc/classes/class-demo-assets.php
+++ b/inc/classes/class-demo-assets.php
@@ -191,6 +191,7 @@ class Demo_Assets {
 	 */
 	private static function create_from_payload( $key, $file_name, $payload ) {
 		$mp4 = ! empty( $payload['transcoded_mp4_url'] ) ? $payload['transcoded_mp4_url'] : '';
+		$mpd = ! empty( $payload['transcoded_file_path'] ) ? $payload['transcoded_file_path'] : '';
 		$hls = ! empty( $payload['transcoded_hls_path'] ) ? $payload['transcoded_hls_path'] : '';
 
 		$data = array(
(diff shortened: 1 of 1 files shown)


subodhr258:
Reviewed. The fix is correct and matches the convention used elsewhere in the plugin.

- `mpd_url` should carry the DASH manifest, not the MP4 URL. The same mapping already exists at `inc/classes/class-media-library-ajax.php:126` (`'mpd_url' => $item['transcoded_file_path'] ?? ''`), and `class-video-migration.php:1093` stores `transcoded_file_path` into the same `rtgodam_transcoded_url` meta that `mpd_url` feeds into (`class-media-library.php:1692`). So `transcoded_file_path` is unambiguously the MPD path across the codebase.
- The old value was effectively dead anyway: the player validates the source with `isMpd()` (`assets/src/js/media-library/views/attachment-details.js:131,152`), so an MP4 URL in the MPD slot was silently rejected. DASH playback only works now that the real manifest URL is passed.
- No regression from the empty fallback: when `transcoded_file_path` is absent, `mpd_url` is an empty string, which the player already handles via the `isMpd()` guard.

LGTM.

subodhr258 (approved):

PR #2006, opened 2026-07-03
Fix media block Replace > Upload bugs (type guard, thumbnail, quality menu, description)

KMchaudhary:
- Fix false 'only X files allowed' error on block media upload.
- Fix stale video thumbnail persisting after replacing video.
- Fix qualityMenu race throwing on video block player init Fix React crash on audio/PDF upload from object description.

## Demo

https://github.com/user-attachments/assets/a915b18a-1b59-45d9-bf71-90534c3fc897

### DIFF
--- a/assets/src/blocks/godam-audio/edit.js
+++ b/assets/src/blocks/godam-audio/edit.js
@@ -91,8 +91,13 @@ function AudioEdit( {
 			return;
 		}
 
-		// Guard against non-audio selections.
-		const mediaType = media.type || ( media.mime || media.mime_type || '' ).split( '/' )[ 0 ];
+		// Guard against non-audio selections. Derive the type from the MIME
+		// string rather than `media.type`: the latter is the top-level type
+		// ("audio") for library selections but the REST post type
+		// ("attachment") for freshly uploaded files, which would reject every
+		// uploaded audio file. `mime` is set on library selections,
+		// `mime_type` on uploads.
+		const mediaType = ( media.mime || media.mime_type || '' ).split( '/' )[ 0 ];
 		if ( mediaType && mediaType !== 'audio' ) {
 			createErrorNotice(
 				__( 'Only audio files are allowed in the GoDAM Audio block.', 'godam' ),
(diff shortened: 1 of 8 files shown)


subodhr258:
### Review

Solid set of fixes. Dropping `media.type` from the guards is the right call (it is `"attachment"` for uploads, which was rejecting every uploaded file), the description normalization prevents the React "objects are not valid as a child" crash, and the static `qualityMenu` import + defensive guard is a clean fix for the init race. A few things worth a look before merge:

**1. (correctness) The new `else { setDefaultPoster( '' ) }` in the id-based meta effect can wipe a GoDAM thumbnail that was just applied synchronously.** `edit.js:312-313`

In `onSelectVideo`, GoDAM media applies its poster right away:
```js
if ( media?.thumbnail_url ) {
    setDefaultPoster( media.thumbnail_url ); // ~532
}
setAttributes( { id: media.id, ... } );      // ~566
```
If `media.id` is numeric (a GoDAM attachment that exists in WP, or once the `godam-virtual-attachment-created` event sets a numeric id), the effect at line 294 fires and fetches `/wp/v2/media/{id}`. When that attachment's `rtgodam_media_video_thumbnail` meta is empty (GoDAM stores the thumbnail on its side, not necessarily in WP meta), the new `else` branch clears `defaultPoster`, so the poster set from `thumbnail_url` flickers on and then disappears. Before this PR there was no `else`, so the synchronous thumbnail survived. Consider skipping the clear when `origin === 'godam'` (or when a `thumbnail_url` was already applied), so the else only clears for plain WP attachments where it is actually needed.

**2. (cleanup) The two normalization helpers are now copy-pasted across blocks.** The description normalization (`typeof media.description === 'string' ? ... : media.description?.raw ?? ''`) is identical in `godam-audio/edit.js` and `godam-pdf/edit.js`, and the MIME derivation (`( media.mime || media.mime_type || '' ).split('/')[0]` / `mediaItem.mime || mediaItem.mime_type`) now appears in five files. There is no shared helper today. Extracting `getMediaMimeType( media )` and `normalizeDescription( media )` into a shared blocks util would keep the fix consistent if the shape changes again.

**3. (scope) Unrelated `package-lock.json` churn.** This PR removes two transitive optional/peer entries (`@wordpress/ui/node_modules/date-fns`, `tailwindcss/node_modules/yaml`) with no matching `package.json` change. Looks like incidental `npm install` dedup rather than part of the fix; worth reverting to keep the diff focused, or confirming it is intentional.

Nothing else blocking. Items 2 and 3 are optional; item 1 is the one I would verify against a real GoDAM attachment.

subodhr258 (approved):

PR #2011, opened 2026-07-09
feat(video-editor): inline title editing + attachment metadata popup

KMchaudhary:
Add 'Edit metadata' kebab action opening the native wp.media attachment modal (enqueues core media-grid), make the title click-to-edit with popup↔title sync, always show the Analytics link, and fix the title-focus layout shift caused by wp-admin's input min-height.

- Add 'Edit metadata' action that opens the native wp.media attachment modal by enqueuing core media-grid
- Make the title click-to-edit with two-way sync between the inline editor and attachment metadata popup
- Always display the Analytics link
- Fix the title focus layout shift caused by wp-admin's default input min-height

https://github.com/user-attachments/assets/5a580994-1223-419b-9fc5-c6b4ceb151ce

### DIFF
--- a/pages/video-editor/components/editor-shell/EditorTopBar.js
+++ b/pages/video-editor/components/editor-shell/EditorTopBar.js
@@ -136,6 +136,7 @@ const EditorTopBar = ( {
 	onSaveTitle,
 } ) => {
 	const homeUrl = window?.godamRestRoute?.homeUrl || '';
+	const hasValidApiKey = Boolean( window?.userData?.validApiKey );
 	const previewUrl = `${ homeUrl }?godam_page=video-preview&id=${ attachmentID }`;
 	const analyticsUrl = `${ homeUrl }/wp-admin/admin.php?page=rtgodam_analytics&id=${ attachmentID }`;
 
@@ -249,14 +250,16 @@ const EditorTopBar = ( {
 					>
 						{ ( { onClose } ) => (
 							<MenuGroup>
-								<MenuItem
-									icon={ chartBar }
-									href={ analyticsUrl }
-									target="_blank"
-									data-test-id="godam-video-editor-button-analytics"
-								>
-									{ __( 'Analytics', 'godam' ) }
-								</MenuItem>
+								{ hasValidApiKey && (
+									<MenuItem
+										icon={ chartBar }
+										href={ analyticsUrl }
+										target="_blank"
+										data-test-id="godam-video-editor-button-analytics"
+									>
+										{ __( 'Analytics', 'godam' ) }
+									</MenuItem>
+								) }
 								<MenuItem
 									icon={ pencil }
 									onClick={ () => {
--- a/pages/video-editor/components/editor-shell/EditorTopBar.js
+++ b/pages/video-editor/components/editor-shell/EditorTopBar.js
@@ -28,6 +28,12 @@ const EditableTitle = ( { title, onSave } ) => {
 	const [ isEditing, setIsEditing ] = useState( false );
 	const [ draft, setDraft ] = useState( title );
 	const inputRef = useRef( null );
+	const buttonRef = useRef( null );
+	// `editingRef` guards commit/cancel against a double-invocation (the input's
+	// blur can fire as it unmounts right after Enter). `refocusRef` records
+	// whether focus should return to the trigger button once edit mode ends.
+	const editingRef = useRef( false );
+	const refocusRef = useRef( false );
 
 	// Keep the draft aligned with upstream title changes while not editing
 	// (e.g. after a successful save or an external refresh).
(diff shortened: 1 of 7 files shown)


subodhr258:
## Code review

Thorough pass over the diff (7 files, +383/-41). Nice work: the wp.media/Backbone bridging and the SCSS-specificity fix are both non-obvious, and the inline comments explaining *why* make them reviewable. Flagging a few things below, ordered by importance.

### Verified as safe (checked, no action needed)
- **Reused-frame path is correct.** I was worried the globally-cached `godamAttachmentEdit` frame would show a stale attachment when you switch videos (VideoEditor remounts via `key={attachmentID}`, but the frame lives on `wp.media.frames`). Traced core `media-grid.js` — `rerender(model)` (`EditAttachments`, ~L589) does `this.model = model` and re-renders the content region, and you set `skipHistory` on each open, so the `?item=` URL rewrite is suppressed. Switching videos + reopening resolves correctly.
- **`decodeHtmlEntities` via detached `<textarea>`** is XSS-safe — textarea is a raw-text element, so entity strings decode without executing or creating nodes.

### Worth addressing

**1. Two unrelated changes bundled into this PR (scope).**
- `EntryScreen.jsx`: the Woo onboarding headline copy (`"You've Unlocked Woo-Specific GoDAM"` → `"You have unlocked GoDAM for Woo"`) has nothing to do with title editing / the metadata popup. User-facing onboarding copy ideally lands in its own PR so it can be reviewed/reverted independently.
- `video-dataview.scss`: `object-fit: contain` → `contain !important` is also unrelated to this feature.

Recommend splitting both out, or at least calling them out explicitly in the description.

**2. Analytics link is now shown unconditionally.** The `hasValidApiKey` gate was removed, so the Analytics menu item renders even without a valid API key. If that's intended (upsell / analytics works regardless), great — just confirm the target page is functional/appropriate for a user with no key, otherwise this surfaces a dead-end action.

**3. No tests for the new logic.** `EditableTitle` (commit/cancel/trim/empty no-op), `decodeHtmlEntities`, and the `openAttachmentDetailsModal` guard clause are all testable units. Per repo convention, logic-bearing changes should ship with coverage — green CI here only means nothing *broke*, not that the new branches are exercised.

### Minor / polish

**4. `change`-listener lifecycle on the cached wp.media model** (`openAttachmentDetails.js`):
- `model.set('skipHistory', true)` runs *after* the listener is attached, so `onChange` fires once immediately with the current attributes. Harmless (parent guards on `typeof === 'string'`), but unintended.
- `onChange` fires on *every* attribute change (alt/caption/description saves), not just title. Consider gating on `model.changed?.title`.
- The listener is never unbound when the modal closes, and the model is cached in `wp.media`. After VideoEditor unmounts, the closure still references the old component's `onChange`. Cleaner to unbind on the frame's `close` event.

**5. `!important` without a comment.** If it's genuinely needed to beat a global rule, a one-line note on *what* it overrides saves the next person a trace (this codebase already has a lot of global `!important` to untangle).

**6. Focus is not returned to the title button after commit/cancel.** On blur-commit the button re-renders and focus drops to `<body>`. Returning focus to `.godam-video-editor__title-button` keeps keyboard flow intact.

**7. Inline title save doesn't invalidate the RTK `getAttachmentMeta` cache** (nor update the Backbone model). Display stays correct because it's driven by local `videoTitle` state, but the two data layers can drift — worth a mention/`invalidatesTags` if anything downstream reads the cached title.

**8. PR description is missing the "Automation impact" line** the team's e2e convention expects for UI PRs. The `data-test-id`s themselves follow the `godam-<area>-<type>-<name>` convention — good.

pages/video-editor/components/editor-shell/EditorTopBar.js
```diff
@@ -37,9 +131,11 @@ const EditorTopBar = ( {
 	onBack,
 	onSave,
 	onCopy,
+	onEditMetadata,
+	editMetadataLabel,
+	onSaveTitle,
 } ) => {
 	const homeUrl = window?.godamRestRoute?.homeUrl || '';
-	const hasValidApiKey = Boolean( window?.userData?.validApiKey );
```
subodhr258, line 42: For what purpose are we removing this Gate from Analytics feature?

KMchaudhary: Resolved in commit [7ff3be7](https://github.com/rtCamp/godam/pull/2011/commits/7ff3be78d9b2f1c5de72010ea97d2992198613f4)

rudrakshi-gupta:
1. handleSaveTitle writes to two data sources with no single source of truth (VideoEditor.js).
It updates local videoTitle state, then separately pokes window.wp.media.attachment(attachmentID) to patch the Backbone model. That's two places tracking the same fact, kept in sync by hand. The comment explains why RTK refetch is avoided, but doesn't explain why the Backbone model write is safe from a similar race — e.g., what happens if handleEditMetadata's modal is open at the same moment a title save resolves? I'd want a one-line confirmation this was actually tested, not just reasoned through.
2. The onChange listener lifecycle in openAttachmentDetails.js is a real gap, not a nitpick.
Per the review thread: the listener fires on every attribute change, not just title, and nothing unbinds it when the modal closes. Since wp.media caches the frame globally rather than per-component, a stale closure over an old attachmentID/setVideoTitle referencing an unmounted component is a plausible bug path, not just style. I'd upgrade this from "polish" to "should fix before merge."
3. Scope creep — two unrelated changes riding in this PR.
EntryScreen.jsx's Woo onboarding copy and the video-dataview.scss !important have nothing to do with title editing or the metadata popup. This makes the PR harder to review and, if either needs reverting later, harder to isolate. I'd ask for these to be split out rather than just "called out in the description" — a description note doesn't give you an independent git-revert point.

@KMchaudhary Please check this review. And implement what's necessary
Other than things LGTM.

rudrakshi-gupta (approved):

pages/video-editor/utils/openAttachmentDetails.js
```diff
@@ -0,0 +1,87 @@
...
+			baseUrl: ( url ) => url,
+		} );
+
+		wp.media.frames.godamAttachmentEdit = new wp.media.view.MediaFrame.EditAttachments( {
+			controller: { gridRouter },
+			// A single-item library hides the prev/next navigation arrows.
+			library: new wp.media.model.Attachments( [ model ] ),
+			model,
+		} );
+	} );
```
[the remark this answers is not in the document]
KMchaudhary: False positive

KMchaudhary:
@rudrakshi-gupta Tested the reported feedback and fixed the required one.

> 1. handleSaveTitle writes to two data sources with no single source of truth

Not an actual issue

> 2. The onChange listener lifecycle in openAttachmentDetails.js is a real gap, not a nitpick.

Resolved

> 3. Scope creep — two unrelated changes riding in this PR.

Intentional

PR #2016, opened 2026-07-14
Fix video preview not resizing to fit the editor stage on tab & viewport changes

KMchaudhary:
This pull request improves the video preview resizing logic in the `VideoEditor` to ensure the video player always fits optimally within the available stage area, especially when the UI layout changes (e.g., timeline dock, panel switches, or window resizing). It introduces a more robust and responsive sizing algorithm, leverages `ResizeObserver` for dynamic adjustments, and updates refs and styles to support these changes.

**Responsive Video Preview Sizing:**

- Introduced a new `resizeVideoPlayer` function using `useCallback` that dynamically calculates the optimal video player width based on available stage height, aspect ratio, and container width, with sensible minimums and maximums to prevent layout issues.
- Replaced the old static sizing logic in the player initialization with a call to the new `resizeVideoPlayer` function, ensuring consistent sizing logic is used throughout.
- Added a `ResizeObserver` to the stage canvas via a `useEffect` hook, so the preview resizes automatically when the stage area changes size (e.g., due to tab switches, timeline visibility, or window resizes).

**Ref and DOM Structure Updates:**

- Added `canvasWrapperRef` and `stageCanvasRef` to directly reference the relevant DOM nodes for sizing calculations and observation, and attached these refs to the appropriate elements in the JSX. [[1]](diffhunk://#diff-0559daddab966567cf8d9f589b35d6c35f4a44b62c60586606523dd0cbcb3494R93-R97) [[2]](diffhunk://#diff-0559daddab966567cf8d9f589b35d6c35f4a44b62c60586606523dd0cbcb3494L603-R682)

**Styling Improvements:**

- Updated `.godam-video-editor__stage-canvas` styles to use `safe center` alignment and added vertical padding, improving layout stability and appearance when resizing.

## Demo

https://github.com/user-attachments/assets/1d19fcac-d0d3-4a3b-9d39-610268461c58

### DIFF
--- a/pages/video-editor/VideoEditor.js
+++ b/pages/video-editor/VideoEditor.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useCallback, useEffect, useState, useRef } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
 /**
@@ -90,6 +90,11 @@ const VideoEditor = ( { attachmentID, onBackToAttachmentPicker } ) => {
 	}, [] );
 
 	const playerRef = useRef( null );
+	// Wraps the video preview; its width drives the player's computed width.
+	const canvasWrapperRef = useRef( null );
+	// The stage area that holds the preview; its width AND height bound the
+	// preview (the layers/chapters tabs dock a timeline that eats vertical room).
+	const stageCanvasRef = useRef( null );
 	// Teardown for the "Edit metadata" popup's title-sync listener; called on unmount.
 	const detachTitleSyncRef = useRef( null );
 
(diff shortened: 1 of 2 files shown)


subodhr258 (approved):

subodhr258:
### Review

**What it does:** Extracts preview sizing into a `resizeVideoPlayer` `useCallback` (height-driven, clamped by container width and a min width), wires a `ResizeObserver` on the stage-canvas so the preview re-fits on tab/timeline/viewport changes, adds `canvasWrapperRef`/`stageCanvasRef`, and switches the stage to `align-items/justify-content: safe center` with vertical padding. Well-commented and defensive — nice.

**Correctness:**
- **Padding vs. buffer double-count.** The SCSS adds `padding: 1rem 0` (32px total), but `availableHeight = clientHeight - VERTICAL_BUFFER` only subtracts 16px. `clientHeight` *includes* padding, so the real usable content height is `clientHeight - 32`. The target height can end up ~16px taller than actually fits, producing a small unnecessary scrollbar. Either subtract the padding as well, or drop `VERTICAL_BUFFER` now that the padding provides the breathing room.
- **Possible ResizeObserver jitter on classic scrollbars.** The observer watches the stage-canvas content-box. On Windows/Linux (non-overlay scrollbars), a vertical scrollbar appearing shrinks the content-box width → re-fires the observer → width recomputes → scrollbar may vanish → oscillation. The `requestAnimationFrame` defer reduces but doesn't fully prevent this; macOS overlay scrollbars are unaffected (likely why it wasn't caught in self-test). Worth a quick check on Windows/Chrome; if it flickers, observe the wrapper or debounce.

**Style / consistency (non-blocking):**
- `player?.el_?.querySelector('video')` uses videojs's private `el_`. The public `el()` exists and is the more common form elsewhere in the player code — prefer it.
- Still `document.querySelector('#easydam-video-player')` while everything else moved to refs. Scoping via a ref (or `stageCanvasRef.current.querySelector`) would be more consistent and safe against duplicate IDs.
- `safe` alignment keyword is ignored by Safari <15.4 (falls back to plain `center`) — acceptable for the admin, just noting.

**Testing:** The width math is pure — extract it to a helper like `computePreviewWidth({ ratio, stageHeight, containerWidth })` and unit-test the clamps (portrait floor never exceeds natural width, `MIN_PREVIEW_WIDTH` floor, container cap). As written it reads the DOM directly and can't be unit-tested.

Logic is sound overall; the padding/buffer math and a Windows scrollbar check are the two things I'd confirm before merge.

PR #2017, opened 2026-07-14
fix: allow native browser shortcuts to pass through in keyboard event handling

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam-plugin-wp/issues/16

This pull request updates the keyboard event handling logic in the `PlayerManager` class to improve user experience and prevent interference with native browser shortcuts.

**Keyboard event handling improvements:**

* Updated the `shouldSkipKeyboardEvent` method in `playerManager.js` to allow native browser shortcuts (such as Cmd+F, Ctrl+F, Ctrl+arrows) to pass through by skipping events with modifier keys (`metaKey`, `ctrlKey`, or `altKey`). This ensures the player only handles unmodified keystrokes and avoids capturing or preventing default behavior for modified keys.

Self-tested on Safari and Chrome Browsers ✅

### DIFF
--- a/assets/src/js/godam-player/managers/playerManager.js
+++ b/assets/src/js/godam-player/managers/playerManager.js
@@ -315,6 +315,13 @@ export default class PlayerManager {
 	 * @return {boolean} True if event should be skipped
 	 */
 	shouldSkipKeyboardEvent( event ) {
+		// Let native browser shortcuts (Cmd+F, Ctrl+F, Ctrl+arrows, etc.) pass
+		// through untouched. The player only handles unmodified keystrokes, so a
+		// modified key must never be captured or preventDefault()-ed.
+		if ( event.metaKey || event.ctrlKey || event.altKey ) {
+			return true;
+		}
+
 		const skipTags = [ 'INPUT', 'TEXTAREA' ];
 		return skipTags.includes( event.target.tagName ) || event.target.isContentEditable;
 	}


subodhr258 (approved):

subodhr258:
### Review

**What it does:** In `shouldSkipKeyboardEvent`, bails out (skips the player's handler) whenever `metaKey`/`ctrlKey`/`altKey` is held, so native browser shortcuts (Cmd/Ctrl+F, Ctrl+←/→, etc.) are no longer captured or `preventDefault()`-ed.

**Correctness:** Verified against the player's keymap in `utils/constants.js` — every shortcut is an *unmodified* key (`f`, `arrowleft`, `arrowright`, ` `/`spacebar`). So skipping modified keystrokes can't break any intended shortcut. Clean, minimal, and correct. The root cause (e.g. `f` → fullscreen swallowing Cmd+F) is addressed directly. 👍

**Notes (non-blocking):**
- `shiftKey` is intentionally not skipped — fine, since no shortcut uses Shift and nothing native at the body level is being blocked. Worth a one-line comment noting Shift is deliberately still handled, so a future reader doesn't "complete the set."
- Consider a small unit test for `shouldSkipKeyboardEvent` — it's nearly pure and cheap to cover (modifier held → true, INPUT/TEXTAREA/contentEditable → true, plain key → false). Guards against regressions if the keymap grows.

LGTM once the above are considered.

PR #2024, opened 2026-07-17
Feat: Register Document Block in WP Bakery

rudrakshi-gupta:
<html><head></head><body><h1>Register Document (PDF) block in WPBakery</h1>

Issue - rtCamp/godam-plugin-wp#19

<h2>Summary</h2>
<p>The <code>godam/pdf</code> <strong>Document</strong> block was the only GoDAM media block not available in the WPBakery (Visual Composer) page builder — Video, Video Gallery, and Audio were all already mapped. This PR registers a <strong>Document</strong> element under the GoDAM category in WPBakery, wired the same way as the existing elements, and exposes the block's redesigned (2.0) attributes so the WPBakery element reaches feature parity with the Gutenberg block, including <strong>Default (embedded viewer)</strong> and <strong>Card</strong> views with an auto‑generated cover.</p>
<h2>Background</h2>
<p>WPBakery elements render through shortcodes (<code>base</code> → shortcode name), and each existing GoDAM element reuses the corresponding block's <code>render.php</code>. The Document block had no WPBakery element and no shortcode, so it couldn't be added from the builder. It also gained new attributes in the 2.0 redesign (title, description, preview mode, cover, etc.) that needed to be surfaced.</p>
<h2>What changed</h2>
<h3>New files</h3>
<ul>
<li><code>inc/classes/wpbakery-elements/class-wpb-godam-document.php</code> — maps the <code>godam_document</code> element via <code>vc_map()</code> under the <strong>GoDAM</strong> category. Follows the exact pattern of the other elements (WPBakery-active guard, <code>vc_before_init</code> hook, <code>Singleton</code>).</li>
<li><code>inc/classes/shortcodes/class-godam-document.php</code> — registers the <code>[godam_document]</code> shortcode, which reuses the Document block's <code>render.php</code> (same approach as <code>[godam_audio]</code>). Maps the flat shortcode attributes to the block's attribute shape.</li>
<li><code>assets/src/js/wpbakery/wpbakery-document-selector-param.js</code> — custom <code>document_selector</code> param: opens the media library filtered to PDFs, stores the attachment ID + source URL, and <strong>auto-populates Doc Title and Description</strong> from the attachment (matching the block's <code>onSelectPdf</code>).</li>
<li><code>assets/src/js/wpbakery/wpbakery-document-cover-selector-param.js</code> — custom <code>document_cover_selector</code> param: renders the <strong>auto-generated cover</strong> (fetched from the attachment's REST meta) as the default tile, with a "Select image" override, mirroring the block's cover picker.</li>
</ul>
<h3>Modified files</h3>
<ul>
<li><code>inc/classes/wpbakery-elements/class-wpb-godam-params.php</code> — registers the two new custom param types (<code>document_selector</code>, <code>document_cover_selector</code>) and their field renderers.</li>
<li><code>inc/classes/shortcodes/class-godam-document.php</code> <em>(new, listed above)</em> and <code>assets/src/blocks/godam-pdf/render.php</code> — <code>render.php</code> now supports being called outside block context: when <code>$godam_is_shortcode</code> is set it skips <code>get_block_wrapper_attributes()</code> (which warns without an active block) and applies any WPBakery <strong>Design Options</strong> CSS class instead.</li>
<li><code>inc/classes/class-plugin.php</code> — registers <code>GoDAM_Document</code> (shortcode) and <code>WPB_GoDAM_Document</code> (element) alongside the existing ones.</li>
<li><code>webpack.config.js</code> — adds build entries for the two new param scripts.</li>
</ul>
<h2>Attributes exposed in the WPBakery element</h2>

WPBakery param (base: godam_document) | Block attribute | Notes
-- | -- | --
id | id | PDF picker (media library, PDF-only)
src | src | Hidden, auto-filled by the picker
doc_title | docTitle | Auto-populated from attachment; editable
description | description | Auto-populated from attachment; editable
preview_mode | previewMode | Default View / Card View
height | height | Default view only
show_cover | showCover | Card view only
custom_cover | customCover | Card + Show Cover; empty = auto-generated cover
caption | caption |  
css | — | WPBakery Design Options tab

<h2>Key implementation notes</h2>
<ul>
<li><strong>Shortcode attribute casing.</strong> WordPress' <code>shortcode_parse_atts()</code> <strong>lowercases every attribute name</strong>, so WPBakery params must use lowercase/snake_case names (as the existing Video/Audio elements do). Params use <code>doc_title</code>, <code>preview_mode</code>, <code>show_cover</code>, <code>custom_cover</code>, and the shortcode maps them back to the block's camelCase attributes (<code>docTitle</code>, <code>previewMode</code>, …). Using camelCase param names silently breaks Card view because <code>previewmode</code> never matches <code>previewMode</code>.</li>
<li><strong>Auto-generated cover.</strong> The cover meta (<code>rtgodam_media_video_thumbnail</code>, then <code>rtgodam_media_pdf_thumbnail</code>) is registered <code>show_in_rest</code>, so it is readable in the public "view" context without a nonce. The cover param fetches <code>/wp/v2/media/{id}</code> client-side to preview the auto cover; an empty <code>custom_cover</code> value means "use the auto cover", which <code>render.php</code> already resolves on the front end.</li>
<li><strong>Single source of truth for rendering.</strong> No markup was duplicated — the shortcode reuses the block's <code>render.php</code>, so block and WPBakery output stay in sync.</li>
</ul>
<h2>How to test</h2>
<ol>
<li>Ensure WPBakery (<code>js_composer</code>) is active. Run <code>npm run build:js</code> (the two new param scripts are generated into <code>assets/build/js/</code>, which is gitignored).</li>
<li>Edit a page with WPBakery → add the <strong>GoDAM → Document</strong> element.</li>
<li>Select a PDF → confirm <strong>Doc Title</strong> and <strong>Description</strong> auto-populate.</li>
<li><strong>Default View:</strong> adjust Height → confirm the embedded PDF renders on the front end.</li>
<li><strong>Card View:</strong> set View = Card, Show Cover = Yes → the <strong>auto-generated cover</strong> appears as the first tile; optionally override via "Select image". Confirm the card (cover, title, description) renders on the front end.</li>
<li>Verify the <strong>Design Options</strong> tab styling applies on the front end.</li>
</ol>
<h2>Notes for reviewers</h2>
<ul>
<li><code>assets/build/js/*.min.js</code> are build artifacts (gitignored) — rebuild with <code>npm run build:js</code>.</li>
<li>No PHPCS error found.</li>
</ul></body></html>

## Demo

https://github.com/user-attachments/assets/0a8019b4-17b6-4ce6-b1bd-2eaa27d151ee

### DIFF
--- a/assets/src/blocks/godam-pdf/render.php
+++ b/assets/src/blocks/godam-pdf/render.php
@@ -53,8 +53,20 @@ if ( empty( $godam_sources ) ) {
 
 $godam_file_name = basename( $godam_sources[0] );
 
+// Root wrapper: in block context merge WordPress' block-support attributes
+// (align/spacing/etc.). The [godam_document] shortcode (used by the WPBakery
+// element) sets $godam_is_shortcode and runs outside a block, where
+// get_block_wrapper_attributes() would raise a warning, so it instead applies
+// any WPBakery Design Options CSS class passed in via $godam_css_class.
+if ( empty( $godam_is_shortcode ) ) {
+	$godam_wrapper_attributes = get_block_wrapper_attributes();
+} else {
+	$godam_shortcode_class    = trim( 'wp-block-godam-pdf ' . ( isset( $godam_css_class ) ? $godam_css_class : '' ) );
+	$godam_wrapper_attributes = 'class="' . esc_attr( $godam_shortcode_class ) . '"';
+}
+
 ?>
-<figure <?php echo wp_kses_data( get_block_wrapper_attributes() ); ?>>
+<figure <?php echo wp_kses_data( $godam_wrapper_attributes ); ?>>
 
 	<?php if ( 'card' === $godam_preview_mode ) : ?>
 
--- a/assets/src/blocks/godam-pdf/render.php
+++ b/assets/src/blocks/godam-pdf/render.php
@@ -163,7 +163,7 @@ if ( empty( $godam_is_shortcode ) ) {
 			style="height: <?php echo esc_attr( $godam_height ); ?>px;"
 		>
 			<object
-				id="pdfObject"
+				id="<?php echo esc_attr( wp_unique_id( 'godam-pdf-object-' ) ); ?>"
 				type="application/pdf"
 				width="100%"
 				height="100%"
--- a/assets/src/js/wpbakery/wpbakery-document-cover-selector-param.js
+++ b/assets/src/js/wpbakery/wpbakery-document-cover-selector-param.js
@@ -1,0 +1,168 @@
+( function( $ ) {
+	'use strict';
+
+	// Initialize on document ready and when WPBakery reloads the params.
+	$( document ).ready( initDocumentCoverSelector );
+	$( document ).on( 'vc.reload', initDocumentCoverSelector );
+
+	/**
+	 * Resolve the WP REST API root. The document cover meta
+	 * (rtgodam_media_video_thumbnail / rtgodam_media_pdf_thumbnail) is registered
+	 * show_in_rest, so it is readable in the public "view" context — no nonce
+	 * required.
+	 *
+	 * @return {string} REST root URL with trailing slash.
+	 */
+	function restRoot() {
+		if ( window.wpApiSettings && window.wpApiSettings.root ) {
+			return window.wpApiSettings.root;
+		}
+		return window.location.origin + '/wp-json/';
+	}
+
+	function initDocumentCoverSelector() {
+		$( '.document_cover_selector_block' ).each( function() {
+			setupBlock( $( this ) );
+		} );
+	}
+
+	/**
+	 * Read the selected document (attachment) ID from the sibling document
+	 * selector param in the same edit form.
+	 *
+	 * @param {Object} $block The cover selector block element.
+	 * @return {string} The attachment ID, or an empty string.
+	 */
+	function getDocumentId( $block ) {
+		const $formContainer = $block.closest( '.wpb_el_type_document_cover_selector' ).parent();
+		const $idInput = $formContainer.find( '.document_selector_field' );
+		return $idInput.length ? $idInput.val() : '';
+	}
+
+	/**
+	 * Fetch the attachment's auto-generated cover URL from its REST meta.
+	 *
+	 * @param {string}   id The attachment ID.
+	 * @param {Function} cb Callback receiving the cover URL (string, may be empty).
+	 */
+	function fetchAutoCover( id, cb ) {
+		if ( ! id ) {
+			cb( '' );
+			return;
+		}
+		fetch( restRoot() + 'wp/v2/media/' + id )
+			.then( ( res ) => ( res.ok ? res.json() : null ) )
+			.then( ( media ) => {
+				let cover = '';
+				if ( media && media.meta ) {
+					// Mirror render.php: video thumbnail first, then pdf thumbnail.
+					cover = media.meta.rtgodam_media_video_thumbnail ||
+						media.meta.rtgodam_media_pdf_thumbnail || '';
+				}
+				cb( cover );
+			} )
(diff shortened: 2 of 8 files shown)


KMchaudhary:
## Review notes

Reviewed the full diff plus the shared `render.php`, the block's `view.js`, the cover-meta REST registration, and the two new custom param scripts. This is a clean PR — it mirrors the existing Video/Audio/Gallery elements, handles the shortcode-lowercasing constraint correctly (snake_case params → camelCase mapping), reuses the block's `render.php` as the single source of truth, and the follow-up commit already hardened the important things (XSS in the doc-selector preview, i18n, `rest_url()`-derived REST root, `[name="src"]`-scoped hidden field). Confirmed the enqueue handles match the `godam/pdf` block name and that both cover meta keys are registered `show_in_rest`. A few smaller items worth a look:

### Worth fixing

1. **Switching View from Card → Default wipes the custom cover** — `wpbakery-document-cover-selector-param.js`
   The `change.godamCoverMode` handler clears `custom_cover` in addition to resetting `show_cover=0`:
   ```js
   $formContainer.find( '[name="show_cover"]' ).val( '0' ).trigger( 'change' );
   $valueInput.val( '' ).trigger( 'change' ); // <- discards the user's cover
   ```
   Resetting `show_cover` to `0` already hides `custom_cover` via its dependency, so the extra `$valueInput.val('')` isn't needed and silently destroys a selected cover URL. Repro: Card view → Show Cover = Yes → pick a custom image → click the View dropdown to peek at Default → back to Card = the image is gone and Show Cover is back to No.

### Minor

2. **Replace overwrites Doc Title / Description** — `wpbakery-document-selector-param.js`
   Selecting/replacing a document unconditionally re-populates both fields from the attachment, discarding any custom text the user typed. It matches the block's `onSelectPdf`, but consider only populating when the field is empty.

3. **Auto-cover REST fetch has no nonce** — `wpbakery-document-cover-selector-param.js`
   `fetch(restRoot + 'wp/v2/media/{id}')` is authenticated as logged-out, so the cover preview works for published attachments (the common case) but silently falls back to the placeholder for non-public ones (draft/private). `render.php` still resolves the cover server-side, so only the editor preview is affected.

4. **Cover tiles still build markup by string concatenation** — `wpbakery-document-cover-selector-param.js` (`tileHtml`)
   `title="' + label + '"` / `src="' + encodeURI(imgUrl) + '"` are interpolated into `innerHTML`. `label` is an i18n string (low risk today), but this is exactly the pattern the follow-up commit removed from the doc-selector preview by switching to `$('<span>').text(...)` — worth keeping consistent.

5. **Duplicate `id="pdfObject"`** — `render.php`
   Now that the shortcode enables multiple Document embeds per page, the hardcoded `id="pdfObject"` can appear more than once (invalid HTML). `view.js` scopes work per `.godam-pdf-wrapper`, so nothing breaks functionally, but the id is worth making unique.

6. **`view.js` only inits on `DOMContentLoaded`** — the Safari/iOS source-fallback + error-retry won't attach in the WPBakery inline-editor preview (markup injected after that event). The native `<object>` still loads `sources[0]`, so Default view mostly looks fine. (PR #2029 added a `readyState`/re-init fix to `frontend.js`; there's no equivalent here — likely fine given `<object>` renders natively, just noting it.)

7. **`page_count` is never supplied** — the element exposes no `page_count` param, so `pageCount` is always `0` and the Card "%d pages" meta never renders for WPBakery documents (parity gap vs the block, which computes it client-side). Fine to leave as a known limitation.

8. **Small redundancy** — `class-godam-document.php`'s empty-`id`/`src` early return and the `intval(height)` both duplicate guards already in `render.php`. Harmless.

Nice work overall — the mapping, dependency chaining, and cover-picker logic are solid.

KMchaudhary (approved):

PR #2028, opened 2026-07-20
feat(analytics): date-range picker + range plumbing (W1/W5)

subodhr258:
## What

Plugin-side date-range analytics (GoDAM 2.0 — rtCamp/godam-plugin-wp#2, W1/W5), building on the microservice range work (rtCamp/godam-analytics#231, #233).

Adds a reusable date-range picker, the full REST/RTK plumbing, a **DRY cross-plugin shared-component** export, and wires the picker into the dashboard **Top Videos** table. Companion PR: rtCamp/godam-for-woo#179 (Reel Pops) reuses the same component.

## Included in this PR

**Reusable `<DateRangePicker>`** (`pages/analytics/components/DateRangePicker/`) — matches Figma node 746-6186: preset chips (**7 Days · 15 Days · 1 Month · All Time**) + a **Date Range** custom calendar with start+end range selection. Emits ISO `{ startDate, endDate }`. `@wordpress/components`-based, themed via `--wp-admin-theme-color`, `data-test-id` hooks, unit-tested helpers.

**DRY cross-plugin sharing** — `pages/shared-components/index.js` re-exports the picker on `window.GoDAM.DateRangePicker`; registered as the `godam-shared-components` script (priority 5) so godam-for-woo reuses the exact same component (no duplication). GoDAM's own pages import it directly.

**REST proxy** (`class-analytics.php`) — optional `start_date`/`end_date` (strict `YYYY-MM-DD`) forwarded on `fetch`, `history`, `dashboard-metrics`, `dashboard-history`, `layer-analytics`, `top-videos`. Additive.

**RTK Query slices** — thread `start_date`/`end_date` (only when set; range beats `days`).

**Live integration** — dashboard **Top Videos** table (range-scoped + CSV export).

## Verified

`wp-scripts build` clean (incl. `shared-components` bundle); `wp-scripts test-unit-js` 8/8 new tests; ESLint clean; PHPCS clean.

## Remaining (core dashboard/per-video — live-QA pass)

These reuse the same picker + plumbing but touch the existing **imperative-DOM** KPI rendering (`charts.js` / `ChartsDashboard.js` set values via `getElementById` + `window.*` globals) and the **d3** Playback Performance chart, so they land as a focused, dev-site-verified follow-up:

- [ ] **Insights KPIs** — per-card picker + real `%` delta badges from the microservice `*_change` (the `SingleMetrics` `changeKey` hook already exists, currently unused).
- [ ] **Viewers gauge** — picker + range-scoped total-plays arc; `unique_viewers` shown unavailable in range mode (returns `null` until the `uniqExactState` rollup).
- [ ] **Playback Performance chart** — swap the preset dropdown for the picker + generalise the d3 x-axis span to an arbitrary range.
- [ ] **Per-video page (W5)** — one page-level picker threaded into the KPI/history/layer hooks.

## Recording

https://github.com/user-attachments/assets/bfa38a71-74f5-4552-ac84-0bd7b927be23


KMchaudhary:
## Code Review: date-range picker + range plumbing (W1/W5)

### Summary
Solid, carefully-scoped change. The reusable `<DateRangePicker>`, the additive REST proxy, and the RTK plumbing are all clean, well-commented, and the tricky parts (RTK cache-key dedup, DST-safe calendar math, d3 re-render idempotency) are handled deliberately rather than by accident. No blocking issues found — the notes below are polish/edge-cases.

### What looks good
- **Additive REST design** — `start_date`/`end_date` are optional, strictly validated (`DateTime::createFromFormat` round-trip), and `append_range_params()` is a no-op when absent, so all-time requests are byte-for-byte unchanged.
- **RTK dedup reasoning is correct** — omitting the date args at All Time (rather than sending `startDate: null`) keeps the same serialized cache key, so the ranged and primary queries collapse into one request until a range is actually picked. Nicely documented at every call site.
- **d3 re-render is genuinely idempotent** — the new `generatePostViewsChart` clear (donut + `#legend` + leaked `body div.tooltip`) plus the pre-existing container clears in `generateCountryHeatmap` mean re-running `main()` on range change replaces charts instead of stacking them. Verified.
- **Removing the imperative KPI writes from `ChartsDashboard.js`** is the right call — the comment about `innerText` detaching React's text nodes and freezing range updates is exactly right.
- **DST-safe span math** — `spanDays` uses calendar-field arithmetic (not fixed-ms subtraction) and there's a targeted unit test asserting exact calendar endpoints rather than an elapsed-ms count that would mask the bug.
- Good accessibility (folded selection into `aria-label`), `data-test-id` hooks throughout, and a clear "Remaining work" checklist.

### Suggestions (all low severity)
| 1 | `pages/analytics/components/DateRangePicker/index.js` | 1173 / 1189 | **Cross-midnight preset drift.** Presets are stored as concrete ISO dates, but `matchPreset`/`triggerLabelFor` recompute `today()` on every render. If the tab is left open past local midnight, a "Last 7 days" selection no longer matches `spanDays(7)`, so the trigger silently relabels to a literal `Nov 7 - Nov 13` custom span and the preset row loses its checkmark. Data stays correct; only the label/selection UI drifts. Consider re-resolving the active preset (or keying it) rather than comparing frozen dates. | Correctness (edge) |
| 2 | `pages/dashboard/Dashboard.js` / `ViewersGauge.js` | 2001–2002 | **"—" flicker during load.** `gaugeMetrics?.unique_viewers ?? null` (and the analytics `?? null`) resolves to `null` while the query is in flight, so the gauge shows "—" before the real all-time count appears (previously `?? 0`). Cosmetic — consider gating the "unavailable" display on `!isLoading` so it only shows for genuine range-mode nulls. | UX |
| 3 | `inc/classes/rest-api/class-analytics.php` | 113 | **No `start_date ≤ end_date` ordering check.** `validate_iso_date` validates format only; an inverted or absurd range from a direct authenticated REST caller is forwarded to the microservice unchecked. The picker swaps endpoints so it's unreachable from the UI, but a cheap ordering guard on the proxy would harden the public surface. | Robustness |
| 4 | `inc/classes/class-pages.php` | 41 | **Shared-components dep array is hand-maintained across a plugin boundary.** Hardcoding `array('wp-element','wp-components','wp-i18n')` matches the sibling bundles, but this handle is now an API surface that godam-for-woo depends on. Worth confirming the generated `shared-components.min.asset.php` externals match (in particular that `@wordpress/icons` is bundled, not externalized to a handle not listed here) so the Woo consumer can't load with a missing dependency. | Maintainability |
| 5 | `pages/analytics/SingleMetrics.js` | 29 | `chartConfigMap` still defines `total_videos_change` while the header comment says total-videos has no server delta. Harmless (badge relies on the server nulling the field), but the mismatch reads as a contradiction — drop the key or adjust the comment. | Clarity |

### Verdict
**Approve** — no correctness blockers. Items 1 and 2 are the most user-visible (both edge/cosmetic); the rest are hardening/clarity nits that can be follow-ups.

KMchaudhary (approved):

PR #2031, opened 2026-07-22
Add render-time video SEO schema for block templates, parts, and patterns

KMchaudhary:
Fixes: https://github.com/rtCamp/godam/issues/2032

This pull request introduces a new mechanism to collect and output VideoObject schema for videos embedded in block-theme templates, template parts, and synced patterns, ensuring that all videos on a page are properly represented in the page's JSON-LD without duplication. The changes add a render-time collection and output process in addition to the existing cached `wp_head` schema output, with careful de-duplication between the two paths.

**Video SEO schema improvements:**

* Added private properties to `Seo` for tracking which video content URLs and attachment IDs have already had schema emitted, and for collecting render-time video schema entries (`$emitted_content_urls`, `$emitted_attachment_ids`, `$render_collected_schemas`).
* Added a render-time collector using the `render_block` filter and a new `collect_render_time_video_seo` method to capture video blocks rendered outside the queried post's content (e.g., in templates or patterns). [[1]](diffhunk://#diff-23ed40a36cd121deb111228e8a2fc7d9b08cdbf3ddd59e9f87a4927a0c92befaR71-R75) [[2]](diffhunk://#diff-23ed40a36cd121deb111228e8a2fc7d9b08cdbf3ddd59e9f87a4927a0c92befaL463-R713)
* Added a new `output_render_time_video_seo_schema` method, hooked to `wp_footer`, to output JSON-LD for videos collected at render time, with de-duplication against videos already emitted in `wp_head`. [[1]](diffhunk://#diff-23ed40a36cd121deb111228e8a2fc7d9b08cdbf3ddd59e9f87a4927a0c92befaR71-R75) [[2]](diffhunk://#diff-23ed40a36cd121deb111228e8a2fc7d9b08cdbf3ddd59e9f87a4927a0c92befaL463-R713)
* Refactored schema construction into a new `build_video_object_schema` method, ensuring both cached and render-time paths produce identical schema and fire the same filters.
* Introduced an `is_render_time_seo_context` method to precisely determine when render-time schema collection/output should run (i.e., only on front-end HTML page renders, not in admin, REST, or feeds).

## Screenshots (Self tested using https://validator.schema.org/ tool)
Frontpage template
Screenshot 2026-07-22 at 7 51 59 PM

Normal page with video and sync patterns
Screenshot 2026-07-22 at 7 54 32 PM

### DIFF
--- a/inc/classes/class-seo.php
+++ b/inc/classes/class-seo.php
@@ -573,6 +573,10 @@ class Seo {
 	 * excludes admin screens, REST requests (e.g. editor block previews) and
 	 * feeds, none of which output a `wp_footer` document head we can attach to.
 	 *
+	 * The `godam_video_seo_render_context` filter lets integrators narrow this
+	 * scope further (e.g. suppress on archives to avoid repeating the same
+	 * schema across listing pages).
+	 *
 	 * @return bool True on a front-end HTML page render, false otherwise.
 	 */
 	private function is_render_time_seo_context() {
@@ -588,7 +592,33 @@ class Seo {
 			return false;
 		}
 
-		return true;
+		/**
+		 * Filters whether render-time video SEO schema should be collected and
+		 * emitted for the current front-end view.
+		 *
+		 * This runs only for genuine front-end page renders — admin, REST and
+		 * feed requests are already excluded and never reach the filter. Return
+		 * false to suppress the render-time VideoObject output on specific
+		 * views. For example, to keep it on singular content and the front page
+		 * only, and off other archives/search (avoiding the same schema being
+		 * repeated across listing pages):
+		 *
+		 *     add_filter(
+		 *         'godam_video_seo_render_context',
+		 *         function ( $emit ) {
+		 *             return is_singular() || is_front_page();
+		 *         }
+		 *     );
+		 *
+		 * All conditional tags (is_singular(), is_archive(), is_front_page(),
+		 * is_search(), …) are available inside the callback.
+		 *
+		 * @since 2.0.0
+		 *
+		 * @param bool $emit Whether to emit render-time video SEO on this view.
+		 *                    Defaults to true for every front-end view.
+		 */
+		return (bool) apply_filters( 'godam_video_seo_render_context', true );
 	}
 
 	/**
--- a/inc/classes/class-seo.php
+++ b/inc/classes/class-seo.php
@@ -24,6 +24,33 @@ class Seo {
 	const POST_ATTACHMENTS_META_KEY         = '_godam_seo_attachments';
 
 	/**
+	 * Content URLs already emitted by the cached wp_head schema output.
+	 *
+	 * Lets the render-time collector skip videos that were already output from
+	 * the queried post's content, avoiding duplicate JSON-LD.
+	 *
+	 * @var array<string,bool>
+	 */
+	private $emitted_content_urls = array();
+
+	/**
+	 * Attachment IDs already emitted by the cached wp_head schema output.
+	 *
+	 * @var array<int,bool>
+	 */
+	private $emitted_attachment_ids = array();
+
+	/**
+	 * Video SEO entries collected at render time from block-theme templates,
+	 * template parts and synced patterns (content that is not part of the
+	 * queried post's own post_content). Keyed by contentUrl (or headline) for
+	 * de-duplication.
+	 *
+	 * @var array<string,array>
+	 */
+	private $render_collected_schemas = array();
+
+	/**
 	 * Construct method.
 	 */
 	protected function __construct() {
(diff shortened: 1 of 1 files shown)


subodhr258:
## Code review

### What this PR does
Extends video `VideoObject` JSON-LD so it also covers `godam/video` blocks that live **outside** the queried post's `post_content` (block-theme templates, template parts, synced patterns). The cached `wp_head` path only ever saw the post's own content; those composed-at-render videos were silently missing from the page schema.

The approach:
- `render_block` collector (`collect_render_time_video_seo`) gathers `godam/video` blocks as they render.
- `wp_footer` output (`output_render_time_video_seo_schema`) emits a single JSON-LD `<script>` for what was collected, de-duplicated against whatever `wp_head` already emitted.
- Schema construction is factored into `build_video_object_schema()`, shared by both paths so they stay identical and fire the same filters.
- `is_render_time_seo_context()` gates the whole thing to real front-end renders (excludes admin/REST/feed) and exposes a `godam_video_seo_render_context` filter.

Overall this is a clean, well-documented refactor. The shared builder, the two-phase collect/flush split, and deferring cross-path de-dup to `wp_footer` (where both state sets are guaranteed populated) are the right calls. Comments are genuinely helpful.

### Correctness

- **De-dup identity gap (edge case).** Flush-time de-dup vs `wp_head` (lines ~713–718) matches only on `contentUrl` and `attachment_id`, but collect-time de-dup falls back to a `headline:` key when both are absent. A `godam/video` with a headline but **no** `contentUrl` and **no** `attachment_id` (e.g. a `seoOverride` entry without a media id) that appears both in the queried post content *and* is render-collected would emit twice, because `wp_head` also can't have recorded it (nothing to key on). Low likelihood in practice (blocks almost always carry an `id`), but worth either handling (also record/compare a normalized headline) or noting as a known limitation.

- **VPG / `godam/video-product-gallery` blocks aren't collected at render time.** The collector hard-filters to `'godam/video'` only, whereas the cached path picks up Shoppable Video via `godam_video_seo_extra_block_schemas`. A VPG block placed in a template/pattern gets no render-time schema. This is probably acceptable scope for this PR, but it should be called out in the linked issue so it isn't assumed covered.

- **Order-independence is fine, but the inline comment overstates it.** The comments (lines ~666–670, ~706–709) assert "the template — and therefore `render_block` — runs *before* `wp_head*`". In the block-theme canvas, `wp_head()` runs in `<head>` and the body's `do_blocks()` runs after, so that ordering is situational. It doesn't matter for correctness — `wp_footer` runs after both, which is exactly why deferring de-dup there is robust — but consider softening the comment so a future reader doesn't rely on the stated ordering.

- **Behavior change in the `wp_head` path (likely an improvement).** Previously a `godam_video_seo_schema` filter returning empty still pushed the (empty) entry; now `build_video_object_schema()` returns it and the caller does `if ( empty( $schema ) ) continue;`. Skipping empties is better, just flagging it as an intentional-looking change.

### Conventions
- **`@since 2.0.0` on the new `godam_video_seo_render_context` filter (line ~616).** Every other filter in this file uses a released number, but the project convention is to write `@since n.e.x.t` for unreleased additions and resolve it at release time. `develop` is already bumped to `2.0.0`, so this is ambiguous — please confirm with the release owner whether new hooks should still use the `n.e.x.t` placeholder here.

### Tests
- No tests added. `tests/php/` already exists (`VideoEditorTest`, `OnboardingResponseTest`), and this PR adds real branching logic (context gating, two-phase de-dup). A PHPUnit test covering: (a) render-time collection for a `godam/video` in a template/pattern, (b) no duplicate when the same video is in the queried post content, and (c) the `godam_video_seo_render_context` filter suppressing output — would lock in the behavior and guard the de-dup edge above.

### Performance
- `render_block` fires for every block on every front-end request. The `blockName !== 'godam/video'` early return is the correct cheapest-first guard, so overhead is negligible. Minor: `is_render_time_seo_context()` (which runs `is_admin()`/`is_feed()` + an `apply_filters`) is evaluated per `godam/video` block and again at flush; caching the result in a property would trim redundant work, though it's not material.

### Security
- The JSON-LD `echo` reuses the existing `wp_json_encode(..., JSON_UNESCAPED_SLASHES | ...)` pattern. Values are sanitized upstream (`sanitize_text_field`, `wp_strip_all_tags`, `esc_url_raw`) in `build_video_object_schema()`, so no new injection surface is introduced. No concern; just noting it matches the established pattern rather than adding one.

### Verdict
Solid, low-risk change that fills a genuine SEO gap for block themes. Nothing blocking. Recommend: (1) confirm the `@since` convention, (2) add PHPUnit coverage for the collect/de-dup logic, and (3) note the headline-only de-dup edge and the VPG-block exclusion in the issue so they're tracked rather than assumed handled.

subodhr258 (approved):

PR #2033, opened 2026-07-22
feat(analytics): Viewer Retention Curve (convert the per-video heatmap)

subodhr258:
## What

Converts the per-video **"Views across the video"** heatmap into a standalone **Viewer Retention Curve**, matching the Figma "Video Detail" frame, and adds a per-card date-range filter over it.

The old card drew a bare line *on top of the video player* (per-second view density). This replaces it with a proper retention chart over the **video timeline** (x-axis = `mm:ss`), plus a **video hero** (thumbnail + title + layer count) so the user still has a visual reference now that the player is gone.

## Retention metric

The curve plots **absolute audience retention** = `views[t] / views[0] x 100` — the percentage of the starting audience still watching at each second, anchored at 100% at `0:00`. It is deliberately **not** forced monotonic: dips are skip-ahead / drop-off, bumps are rewatch / skip-to (which can briefly exceed 100%). This is what makes it a retention *metric* rather than the old raw-count overlay, which plotted the same `all_time_heatmap` as bare counts. The tooltip shows the retention % plus the underlying viewer count and timestamp.

## Date-range filter

A `DateRangePicker` sits in the retention card head (`testIdPrefix="godam-retention-daterange"`), following the same **per-card** picker pattern #2028 uses for the Insights card. All-time by default; when a range is set, the curve reads a range-scoped copy of the payload. The range-scoped `all_time_heatmap` is summed across the in-range daily heatmaps **server-side** in godam-analytics **#235** (the daily `heatmap` column is per-day and summable; the all-time snapshot is not).

## Changes

- **`pages/analytics/helper.js`** — `generateRetentionCurve()`: self-contained d3 area+line chart (axes, integer-second `mm:ss` x-ticks, gridlines, hover tooltip). Normalizes counts to % of the starting audience; y-domain keeps headroom above 100% for rewatch spikes. Idempotent + viewBox-scaled (responsive, no resize handler). `generateLineChart()` is left as-is for the A/B overlays.
- **`pages/analytics/Analytics.js`** — card is now the Viewer Retention Curve with an empty state; a video hero (thumbnail / title / layer-count badge); a retention date picker scoping a range copy of the analytics payload. A hidden `#analytics-video` element is retained solely to preserve `charts.js`'s per-video render trigger (KPIs / geography / Views-by-Source key off that id). Subheading is now "Single Video Analytics" (removes the duplicated video-title line).
- **`pages/analytics/index.scss`** — neutral-grey curve + area per Figma; WP admin accent only for the hover focus dot; a scoped card-head layout for the picker.
- Plus **5 files vendored from #2028** (`DateRangePicker/{index,index.test,style}`, `analyticsApi.js`, `class-analytics.php`) so this PR builds standalone — see **Dependencies**.

## Dependencies (merge order)

> [!IMPORTANT]
> The date-range filter reuses date-range infrastructure that originates on **#2028**, not on `develop`:
> - the shared `DateRangePicker` component,
> - `fetchAnalyticsData` range threading (`analyticsApi.js`),
> - REST proxy range forwarding (`class-analytics.php`).
>
> To keep this PR buildable and testable on its own, those five files are **vendored verbatim** from `#2028` in commit `dc8340c1` (the dashboard range wiring from the same #2028 commit is intentionally excluded — it's only half-wired without the rest of #2028). **#2028 is the source of truth.**
>
> **Merge #2028 into `develop` first**, then rebase this branch: on the rebase these five files de-duplicate against `develop` (drop this vendored copy, keep develop's), and the `Analytics.js` conflict (both PRs edit the per-video page) resolves at the same point. Do not merge this ahead of #2028.
>
> Also requires godam-analytics **#235** deployed for range mode to return the summed heatmap (all-time mode is unaffected).

## Verification

`wp-scripts build` clean; ESLint clean on the changed JS; the vendored `DateRangePicker` unit test passes (8/8); CI on this branch is green (JS unit tests + PHPCS). Rendered live on godam-dev.local — the all-time curve, video hero, and picker all render correctly (see **Screenshots**). Full range-mode E2E (a range returning the summed heatmap) is covered by #235's unit tests and lands once #235 is deployed.

> [!NOTE]
> If a screenshot looks sparse ("1 viewer" / spiky), that's a **dev-data artifact**, not a curve defect: the dev site's API key scopes analytics to a test account where most videos have a single viewer. Verified against ClickHouse — the curve faithfully plots the `all_time_heatmap` it receives.

## Automation impact

New E2E hook: the retention date picker exposes `data-test-id` under the `godam-retention-daterange` prefix (open trigger, presets, apply/clear), consistent with #2028's `godam-video-insights-daterange`.

## Screenshots

Screenshot 2026-07-23 at 4 08 43 PM
Screenshot 2026-07-23 at 1 52 52 PM


subodhr258:
Thanks for the review — addressed in `b559be1e`, plus two items from manual testing:

**Retention curve was not monotonic (manual testing).** `all_time_heatmap` is per-second *concurrent* view counts (spiky as viewers seek/rewatch), so plotting it raw produced up-and-down spikes. A retention curve must be non-increasing ("viewers who reached at least second *t*"), so it's now reconstructed as the **suffix maximum** — the tightest monotonic curve consistent with the data. Line, area, and tooltip all use it.

**Video hero added.** Since the inline player was removed, added the Figma "Video Detail" head: thumbnail (`meta.rtgodam_media_video_thumbnail`, default fallback) + title + a real layer-count badge (`meta.rtgodam_meta.layers`). Placements badge omitted — no data source yet (that's the deferred Placements feature).

**Copilot comments:**
- **x-axis duplicate labels** — now uses integer-second tick values at a fixed step (always including the last second) instead of d3's fractional ticks rounded to `mm:ss`.
- **`getBoundingClientRect()` in `mousemove`** — the rect/scale is cached on `mouseenter` now, no per-move layout read.
- **double JSON parse** — `all_time_heatmap` is parsed once via `useMemo`, reused by the effect and the empty-state flag.
- **hidden `#analytics-video` trigger** — changed the placeholder from a fake `<video>` to a neutral hidden `<div>` (charts.js only reads its id + `data-id`). The proper fix (explicit exported init keyed on `attachmentID`) lands with the charts.js refactor in the date-range PR (#2028).

subodhr258:
**Correction on the curve semantics** (`7f236694`) — the earlier "make it monotonic" fix was wrong.

After researching how audience-retention curves actually work: they are **not** monotonic. Retention = the % of the video's starting viewers still watching at each second (`views[t] / views[0]`), anchored at 100% at 0:00. Dips mark skip-ahead / drop-off, bumps mark rewatch / skip-to (a rewatched second can exceed 100%). My suffix-maximum transform both destroyed that dip/spike information and mis-handled skip-forward (it back-filled a skipped-past section to fully-retained).

So this now:
- plots **% of starting viewers** (the retention metric) instead of raw counts — that's the real difference from the old "Views across the video" count heatmap;
- stays non-monotonic (correct);
- keeps the underlying viewer count in the hover tooltip.

Verified on the dev site (video 163): 100% → ~36% smooth decline, y-axis in %, tooltip shows % + viewers.

KMchaudhary (approved):

PR #2036, opened 2026-07-23
Feat: Add Image Block in WP Bakery

rudrakshi-gupta:
# Register the GoDAM Image block in WPBakery

Issue - https://github.com/rtCamp/godam-plugin-wp/issues/19

## Summary

The `godam/image` **GoDAM Image** block (hotspot / WooCommerce product layers over an image) was available in Gutenberg but not in the WPBakery (Visual Composer) page builder. This PR registers a **GoDAM Image** element under the GoDAM category in WPBakery — wired the same way as the existing Video, Video Gallery, Audio, and Document elements — and makes the hotspot / product layers render in the WPBakery inline-editor preview.

Branch: `feat/image-block-wp-bakery` → `develop`

## Background

WPBakery elements render through shortcodes (`base` → shortcode name), and each existing GoDAM element reuses the corresponding block's `render.php` as the single source of truth. The Image block had no WPBakery element and no shortcode, so it couldn't be added from the builder.

The GoDAM Image block keys its authored hotspot / product-hotspot layers off the **attachment ID** (they live in the attachment's `rtgodam_meta`), so the WPBakery element captures the image by ID (not just a URL).

### New files
- `inc/classes/wpbakery-elements/class-wpb-godam-image.php` — maps the `godam_image` element via `vc_map()` under the **GoDAM** category (WPBakery-active guard, `vc_before_init` hook, `Singleton`), mirroring the other GoDAM elements.
- `inc/classes/shortcodes/class-godam-image.php` — registers the `[godam_image]` shortcode, which reuses the Image block's `render.php` (same approach as `[godam_audio]`/`[godam_document]`) and maps the flat shortcode attributes to the block's attribute shape. It also enqueues the image + hotspot styles the block normally attaches via `wp_enqueue_block_style()` (which only fires in block context).
- `assets/src/js/wpbakery/wpbakery-image-selector-param.js` — the custom `image_selector` param UI: opens the media library (images only), stores the **attachment ID**, and auto-fills the hidden `url` / `alt` params from the chosen attachment (the render falls back to the attachment for anything empty).

### Modified files
- `inc/classes/wpbakery-elements/class-wpb-godam-params.php` — registers the new `image_selector` custom param type and its server-rendered field (ID-based, distinct from the existing URL-based `image_src_selector`).
- `assets/src/blocks/godam-image/render.php` — supports being called outside block context: when the shortcode sets `$godam_is_shortcode` it skips `get_block_wrapper_attributes()` (which warns without an active block) and applies the stable `godam-image` hook class + any WPBakery **Design Options** CSS class instead.
- `inc/classes/class-plugin.php` — registers `GoDAM_Image` (shortcode) and `WPB_GoDAM_Image` (element) alongside the existing ones.
- `webpack.config.js` — adds the build entry for the new `image_selector` param script.
- `assets/src/js/godam-image-layers/frontend.js` — makes the layers render in the **WPBakery editor preview** (see below).

## Element params (`base: godam_image`)

| WPBakery param | Block attribute | Notes |
|---|---|---|
| `id` | `id` | Image picker (media library, images only) |
| `url` | `url` | Hidden, auto-filled by the picker |
| `alt` | `alt` | Hidden, auto-filled by the picker |
| `show_image_layers` | `showImageLayers` | Toggle (default on) — overlays the authored hotspot / product layers |
| `css` | — | WPBakery Design Options tab |

## Layers in the WPBakery editor preview

**Problem:** The block's hotspot / product layers are drawn by the front-end script `godam-image-layers/frontend.js`, which initialized only on `DOMContentLoaded`. WPBakery's inline editor renders the shortcode markup into its preview iframe **after** that event, so the script never saw the injected image and the layers were invisible in the WPBakery editor (they render fine on the published page).

**Fix (in `frontend.js`):**
- Run init immediately when the DOM is already parsed (`readyState` fallback), not only on `DOMContentLoaded`.
- In a WPBakery editor preview only (detected via the `#vc_inline-frame` iframe / `vc_editor` body class), re-run init on a few deferred passes plus a `MutationObserver` for ongoing edits. `initFrame()` is idempotent (guarded by `data-godam-layers-rendered`), so re-running is safe and never double-draws. Nothing extra is scheduled on the published front end.

This is the same editor-init pattern used for the Video player preview.

## Key implementation notes

- **ID-based selection.** Layers resolve from the attachment's `rtgodam_meta`, so the element stores the attachment ID (via `image_selector`) rather than only a URL; `url`/`alt` are auto-filled for convenience and the render falls back to the attachment when they're empty.
- **Single source of truth.** No markup is duplicated — the shortcode reuses the block's `render.php`, so block and WPBakery output stay in sync.
- **Shortcode attribute casing.** As with the other elements, param names are lowercase/snake_case (`show_image_layers`) because WordPress lowercases shortcode attribute names; the shortcode maps them to the block's camelCase attributes.

## How to test

1. Ensure WPBakery (`js_composer`) is active. Run `npm run build:js` and `npm run build:blocks` (build artifacts under `assets/build/` are gitignored).
2. Edit a page with WPBakery → add the **GoDAM → GoDAM Image** element.
3. Pick an image (ideally one with authored hotspot / product layers) → confirm the selected-image preview and the **Show Image Layers** toggle.
4. In the WPBakery inline-editor preview, confirm the image renders and — for an image with layers and the toggle on — the hotspot / product layers are drawn.
5. Publish and confirm the front end renders the image + layers (parity with the Gutenberg-inserted block).
6. Verify the **Design Options** tab styling applies on the front end.

## Notes for reviewers

- `assets/build/**` are build artifacts (gitignored) — rebuild with `npm run build:js` / `npm run build:blocks`.
- Drawing the layers inside the **Gutenberg** block-editor canvas (and the companion `godam-for-woo` editor enqueue for product hotspots) is a separate follow-up and not part of this PR.

## Demo

https://github.com/user-attachments/assets/56df847e-636e-4483-90fa-cdddf422c2e9

### DIFF
--- a/inc/classes/shortcodes/class-godam-image.php
+++ b/inc/classes/shortcodes/class-godam-image.php
@@ -25,6 +25,64 @@ class GoDAM_Image {
 	 */
 	final protected function __construct() {
 		add_shortcode( 'godam_image', array( $this, 'render' ) );
+		add_action( 'wp_enqueue_scripts', array( $this, 'preload_wpbakery_editor_assets' ) );
+	}
+
+	/**
+	 * Register the shared image-layers front-end script (idempotent).
+	 *
+	 * Mirrors the lazy registration in the block's render.php so the script can
+	 * also be enqueued up front (e.g. in the WPBakery inline editor) with the
+	 * same Woo dependencies applied via the filter.
+	 *
+	 * @return void
+	 */
+	private function register_image_layers_script() {
+		if ( wp_script_is( 'godam-image-layers-frontend', 'registered' ) ) {
+			return;
+		}
+
+		$asset_path = RTGODAM_PATH . 'assets/build/js/godam-image-layers-frontend.min.asset.php';
+		$asset      = file_exists( $asset_path )
+			? include $asset_path
+			: array(
+				'dependencies' => array(),
+				'version'      => RTGODAM_VERSION,
+			);
+
+		$deps = apply_filters( 'godam_image_layers_frontend_dependencies', $asset['dependencies'] );
+
+		wp_register_script(
+			'godam-image-layers-frontend',
+			RTGODAM_URL . 'assets/build/js/godam-image-layers-frontend.min.js',
+			$deps,
+			$asset['version'],
+			true
+		);
+	}
+
+	/**
+	 * Preload the image-layers assets in the WPBakery inline editor.
+	 *
+	 * WPBakery adds/updates an element via AJAX and does NOT inject newly
+	 * enqueued scripts into the running editor page. So when a GoDAM Image is
+	 * added for the first time and saved, the layers script wouldn't be present
+	 * yet — the hotspot / product layers only appeared after a full reload. Load
+	 * the script (and hotspot styles) up front in the inline editor so the
+	 * editor-scoped re-init in frontend.js can draw a just-added image's layers
+	 * immediately, without a refresh. Editor-only; nothing changes on the front end.
+	 *
+	 * @return void
+	 */
+	public function preload_wpbakery_editor_assets() {
+		if ( ! function_exists( 'vc_is_inline' ) || ! vc_is_inline() ) {
+			return;
+		}
+
+		$this->register_image_layers_script();
+		wp_enqueue_script( 'godam-image-layers-frontend' );
+		wp_enqueue_style( 'godam-image-style' );
+		wp_enqueue_style( 'godam-player-style' );
 	}
 
 	/**
@@ -52,8 +110,10 @@ class GoDAM_Image {
 		}
 
 		// Map the flat shortcode atts to the block attribute shape render.php reads.
+		// Normalize the attachment ID to an int (shortcode atts arrive as strings)
+		// so it matches how core attachment APIs and the block attribute expect it.
 		$attributes = array(
-			'id'              => $atts['id'],
+			'id'              => absint( $atts['id'] ),
 			'url'             => $atts['url'],
 			'alt'             => $atts['alt'],
 			'showImageLayers' => filter_var( $atts['show_image_layers'], FILTER_VALIDATE_BOOLEAN ),
--- a/inc/classes/shortcodes/class-godam-image.php
+++ b/inc/classes/shortcodes/class-godam-image.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * GoDAM Image Shortcode Class.
+ *
+ * @package GoDAM
+ */
+
+namespace RTGODAM\Inc\Shortcodes;
+
+defined( 'ABSPATH' ) || exit;
+
+use RTGODAM\Inc\Traits\Singleton;
+
+/**
+ * Class GoDAM_Image.
+ *
+ * Handles the [godam_image] shortcode, which powers the WPBakery "GoDAM Image"
+ * element by reusing the Image block's render template.
+ */
+class GoDAM_Image {
+	use Singleton;
+
+	/**
+	 * Constructor.
+	 */
+	final protected function __construct() {
+		add_shortcode( 'godam_image', array( $this, 'render' ) );
+	}
+
+	/**
+	 * Render the GoDAM image shortcode.
+	 *
+	 * @param array $atts Shortcode attributes.
+	 * @return string HTML output of the image (with hotspot/product layers).
+	 */
+	public function render( $atts ) {
+		$atts = shortcode_atts(
+			array(
+				'id'                => '',
+				'url'               => '',
+				'alt'               => '',
+				'show_image_layers' => true,
+				'css'               => '',
+			),
+			$atts,
+			'godam_image'
+		);
+
+		// A GoDAM Image needs either an attachment ID or a source URL.
+		if ( empty( $atts['id'] ) && empty( $atts['url'] ) ) {
+			return '';
+		}
+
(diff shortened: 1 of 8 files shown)


KMchaudhary:
## Code Review: Feat: Add Image Block in WP Bakery

### Summary
Registers a **GoDAM Image** WPBakery element + `[godam_image]` shortcode that reuses the block's `render.php` as the single source of truth, and makes the hotspot/product layers render in the WPBakery inline preview. The implementation closely (and correctly) mirrors the existing Audio/Document elements, escaping is thorough, and the front-end code paths are well-guarded so there's **no production regression**. Solid PR — findings below are all minor.

### What looks good
- **ID-based selection with server-side fallback** — `render.php` falls back to the attachment's own URL/alt/dimensions when the `url`/`alt` params are empty, so even if the selector JS fails to populate the hidden fields the front end still renders correctly. Nicely defensive.
- **Shortcode is a thin adapter** — no markup duplication; `$godam_is_shortcode` branch matches the Document/Audio precedent exactly (`get_block_wrapper_attributes()` skipped, hook class + Design Options class applied instead).
- **Front-end guarding is correct** — verified that `isBuilderPreview()` returns false on the published front end (no `vc_editor` body class / vc frame), the extra timers + `MutationObserver` only run in the editor, and the `readyState` fallback still hits the `DOMContentLoaded` branch when the footer script runs (parser still `loading`). The PR body's "nothing extra scheduled on the front end" claim holds up.
- **Checkbox → boolean mapping** (`show_image_layers` → `filter_var(..., FILTER_VALIDATE_BOOLEAN)`) and the `checkbox`/`std => 'true'` param shape are identical to the Audio element. Correct.
- **Selector JS is actually an improvement over the sibling** — uses `.attr('src', …)` instead of HTML string concatenation, and adds `type="button"` to the injected buttons (avoids accidental form submits). Escaping in the PHP field (`esc_attr`/`esc_url` + `is_numeric($value)` guard) is right.

### Suggestions
| 1 | `assets/src/js/wpbakery/wpbakery-image-selector-param.js` | 20, 22, 41, 51, 84, 100 | **Localization inconsistency.** The PHP field renders translated labels (`esc_html__( 'Select image' / 'Replace' / 'Remove', 'godam' )`), but the JS hardcodes English (`$button.text('Replace')`, remove button `'Remove'`, reset `'Select image'`, media-frame `'Select or Upload Image'` / `'Select Image'`). On a non-English site the labels render translated, then flip to English after the first select/remove. Use `const { __ } = wp.i18n;` (add `wp-i18n` as a dep + `wp_set_script_translations`). Matches existing debt in `wpbakery-image-src-selector-param.js`, but worth fixing in the new file. | Maintainability / i18n |
| 2 | `inc/classes/wpbakery-elements/class-wpb-godam-image.php` | 55 | **Icon path convention.** Uses `assets/src/images/godam-image-filled.svg` (file exists, so it works) — but Video/Video Gallery/Document elements use `assets/images/…`. Only Audio uses `assets/src/…`. Two conventions now coexist; consider standardizing. | Consistency |
| 3 | `assets/src/js/wpbakery/wpbakery-image-selector-param.js` | 22 vs 100 | **Label casing.** Media-frame button is `'Select Image'` (capital I) while the field's reset text is `'Select image'`. Cosmetic. | Style |
| 4 | `inc/classes/shortcodes/class-godam-image.php` | 78 | **Build reminder (not a defect).** The shortcode `require`s the *build* `render.php`; the `src` change only takes effect after `npm run build:blocks`. Just make sure the release/build pipeline picks it up — the checked-in build render.php still has the pre-PR single-line wrapper. | Process |

### Notes / non-issues (verified, no change needed)
- `wp_enqueue_style( 'godam-image-style' )` resolves — it's the auto-generated handle for `godam/image`'s `block.json` `style` field; `godam-player-style` is registered via `wp_enqueue_block_style()` in `class-blocks.php`. Both are registered by `init`, well before the shortcode runs.
- `[name="url"]` / `[name="alt"]` selectors look broad but are scoped to the param container's `.parent()` and are exactly the element's own `url`/`alt` params; also resilient via the server-side fallback.
- `window.frameElement` cross-origin access is wrapped in try/catch. Good.
- No automated coverage added, but the `data-test-id` hooks (`godam-wpb-*`) are in place for E2E; consistent with the sibling elements.

### Verdict
**Approve** (with minor suggestions) — nothing blocking. Recommend addressing the i18n point (#1) since it's a fresh file; the rest are cosmetic/process.

rtBot (commented):

**phpcs** scanning turned up:

:warning: 1 warning 


***


inc/classes/shortcodes/class-godam-image.php
```diff
@@ -0,0 +1,141 @@
...
+	 * @return void
+	 */
+	private function register_image_layers_script() {
+		if ( wp_script_is( 'godam-image-layers-frontend', 'registered' ) ) {
+			return;
+		}
+
+		$asset_path = RTGODAM_PATH . 'assets/build/js/godam-image-layers-frontend.min.asset.php';
+		$asset      = file_exists( $asset_path )
+			? include $asset_path
```
rtBot, line 48: :warning: **Warning**: File inclusion using variable (`$asset_path`). Probably needs manual inspection (*WordPressVIPMinimum.Files.IncludingFile.UsingVariable*).

KMchaudhary (approved):

PR #2037, opened 2026-07-23
Fix/media editor qa

KMchaudhary:
This pull request implements a comprehensive refactor and enhancement of the hotspot layer styling system to match the WooCommerce product hotspot behaviour, improve colour handling, and add robust support for audio previews. The main themes are: (1) improved and unified hotspot style resolution and UI, (2) frontend and editor consistency for hotspot appearance, and (3) extending the preview system to support audio files with appropriate UI changes.

**Hotspot Layer Styling Improvements**

* Refactored `resolveHotspotStyle` to unify style resolution for both new (layer-level) and legacy (per-hotspot) models, ensuring that icon, custom icon, background, and glyph colours are resolved consistently and match WooCommerce behaviour. Added new defaults for icon colour and custom icon backgrounds, and clarified the fallback logic.
* Updated the editor (`HotspotLayer.js`) to only seed `styleType` and `backgroundColor` during legacy migration, letting pickers and `resolveHotspotStyle` handle other defaults. Colour pickers for icon and background now mirror WooCommerce, and UI logic ensures custom icons and library glyphs get the correct default and user-selected colours. [[1]](diffhunk://#diff-cc8f25ee0ba0c63a5a0e4bc2d60dbd2a25b017461aa758d25ef353f1900bdf4cL39-R39) [[2]](diffhunk://#diff-cc8f25ee0ba0c63a5a0e4bc2d60dbd2a25b017461aa758d25ef353f1900bdf4cL100-R122) [[3]](diffhunk://#diff-cc8f25ee0ba0c63a5a0e4bc2d60dbd2a25b017461aa758d25ef353f1900bdf4cR554-R585)
* On the frontend (`hotspotLayerManager.js`), ensured that icon and background colours are applied according to the resolved style, passing icon colour to the icon element and using the correct background for custom images and glyphs. [[1]](diffhunk://#diff-018a48b7e80f60287b83da3683ef9de37f32ba8a6fd70dfc124feaab4986d681L630-R632) [[2]](diffhunk://#diff-018a48b7e80f60287b83da3683ef9de37f32ba8a6fd70dfc124feaab4986d681L658-R659) [[3]](diffhunk://#diff-018a48b7e80f60287b83da3683ef9de37f32ba8a6fd70dfc124feaab4986d681R678-R681) [[4]](diffhunk://#diff-018a48b7e80f60287b83da3683ef9de37f32ba8a6fd70dfc124feaab4986d681L689-R691) [[5]](diffhunk://#diff-018a48b7e80f60287b83da3683ef9de37f32ba8a6fd70dfc124feaab4986d681L646-R647)

**Preview and Media Type Support**

* Extended the preview page and media capabilities to support audio files, adapting labels and UI (like the Analytics and Edit buttons) to show appropriate options for images, audio, and video. Audio now uses the shared preview page with correct labelling and components. [[1]](diffhunk://#diff-8d5b3a697adbbde3a3cd7d21f701f6a6822f6cbb9a449806d1f78bb95cf89c8eL19-R31) [[2]](diffhunk://#diff-8d5b3a697adbbde3a3cd7d21f701f6a6822f6cbb9a449806d1f78bb95cf89c8eL62-R71) [[3]](diffhunk://#diff-8d5b3a697adbbde3a3cd7d21f701f6a6822f6cbb9a449806d1f78bb95cf89c8eL72-R88) [[4]](diffhunk://#diff-ca52aaad878bed81a6bc6e7f38c68bcf6423d289e5acd2d6508e7b9fc550cfd5L57-R62)

**Layer Initialization and Defaults**

* Changed the default hotspot layer initialization to avoid seeding colours that would override the new conditional defaults, ensuring that the UI and rendered output always reflect the intended default colours unless changed by the user.

**Editor and Frontend Consistency**

* Ensured that both the editor and frontend always use the same logic and default colours for hotspots, so the preview and published output match exactly, including for legacy layers. [[1]](diffhunk://#diff-949eb5cf1a64e3c11fb3ee09e7e494b72e2bfa9a809b50dd9b5a9ee8775091c9R132-R138) [[2]](diffhunk://#diff-cc8f25ee0ba0c63a5a0e4bc2d60dbd2a25b017461aa758d25ef353f1900bdf4cL726-R751) [[3]](diffhunk://#diff-cc8f25ee0ba0c63a5a0e4bc2d60dbd2a25b017461aa758d25ef353f1900bdf4cL738-R763)

These changes provide a more robust, predictable, and user-friendly experience for both editors and viewers, and set the foundation for future enhancements to hotspot styling and media support.


subodhr258:
## Review

Reviewed at `689f36ee`. Three logical groups here: the hotspot style unification, the audio deleted-attachment state, and audio joining the shared preview page.

The PHP side (`video-preview.php`) is clean: correct escaping, correct i18n, and the `previewPage: 'video-preview'` flip lines up with the audio branch that already exists in `godam_preview_page_content()`. The inline comments explaining *why* each fallback exists are genuinely useful, please keep that style.

Most of my notes are on `resolveHotspotStyle()`. The new fallback chain mixes layer-level and per-hotspot fields, and on pre-2.0 data that has some consequences I don't think are intended.

---

### 1. Legacy hotspot layers change appearance on the front end with no re-save

Pre-2.0 hotspots were always created with `backgroundColor: '#0c80dfa6'` seeded (old `HotspotLayer.js`, hotspot factory), and the old player rendered any hotspot that had an icon as **white circle + black glyph**:

```js
// develop, hotspotLayerManager.js
hotspotDiv.style.backgroundColor = ( style.icon || style.customIconUrl ) ? 'white' : ( style.color || '#0c80dfa6' );
iconEl.style.color = '#000';
```

After this PR, for the same untouched saved layer:

- `icon` resolves from `hotspot.icon`, `hasIcon` is true, `styleType` infers to `'icon'`
- `color` = `layer.backgroundColor` (unset) `|| hotspot.backgroundColor` (`#0c80dfa6`, always present on legacy data) `|| bgDefault`
- `iconColor` = `#ffffff`

So every already-published legacy hotspot layer flips from **white circle / black glyph** to **translucent blue circle / white glyph** the moment the plugin updates. No editor visit, no re-save.

The custom-image case worries me more: an uploaded logo PNG that was authored against a white circle now sits on translucent blue, and `DEFAULT_HOTSPOT_CUSTOM_ICON_BG = '#ffffff'` can never apply to legacy data because `hotspot.backgroundColor` is always populated and wins first. The "custom image sits on a neutral white circle by default" comment is only true for layers created after this PR.

If matching Woo is the deliberate call, that's fine, but it should be an explicit, called-out visual migration rather than a side effect: please add it to the PR description and the release notes, and QA it against a site with pre-2.0 hotspot layers. If it isn't intended, the narrow fix is to only consult `hotspot.backgroundColor` in icon mode when the layer has actually been migrated (`layer.backgroundColor` set), and let the conditional default win otherwise.

### 2. "Icon" style with no icon selected renders an invisible white circle, with no picker to fix it

`VeSegmented` sets `styleType: 'icon'` immediately, before any icon is chosen. At that point:

- `HotspotLayer.js:576` gates both pickers behind `hasLayerIcon`, so **no colour picker renders at all**
- `resolveHotspotStyle()` takes the icon branch, `icon` is null, so `bgDefault` is `DEFAULT_HOTSPOT_CUSTOM_ICON_BG` (`#ffffff`)
- editor preview (`HotspotLayer.js:751`) and player (`hotspotLayerManager.js:631`) both paint the circle `#ffffff`

Result: a white pulsing dot on (usually) light content, effectively invisible, and the UI offers no control to change it. On `develop` this same state rendered blue via `layer.iconColor || DEFAULT_HOTSPOT_COLOR`.

Suggestion: make `bgDefault` fall back to `DEFAULT_HOTSPOT_COLOR` when neither `icon` nor `customIconUrl` is set, and/or always render the Background picker in icon mode (drop the `hasLayerIcon` gate on that one, keep it on the glyph "Icon" picker).

### 3. The icon cannot be cleared on a migrated legacy layer

`FontAwesomeIconPicker`'s trash button fires `onChange({ icon: null, ... })`, which `HotspotLayer` stores as `updateField( 'icon', '' )`. Then:

```js
const icon = layer?.icon || hotspot.icon || null;
```

`''` is falsy, so it falls straight through to the stale per-hotspot `hotspot.icon` that migration never clears. On a migrated legacy layer, "Remove icon" appears to do nothing in both the editor preview and the player, and individual hotspots can each resurrect their own different old glyph on a layer that is supposed to have one shared style. New layers are unaffected (their hotspots carry no `icon`).

Either clear the per-hotspot style fields during migration, or distinguish "unset" from "explicitly cleared" (e.g. only fall back when `layer.icon === undefined`).

### 4. Migration seeds a field that pulse mode never reads, and leaves the one it does read unset

```js
dispatch( updateLayerField( { id: layer.id, field: 'backgroundColor', value: first.backgroundColor || DEFAULT_HOTSPOT_COLOR } ) );
```

For a legacy **pulse** layer this writes `backgroundColor`, which is only consulted in the icon branch, and leaves `pulseColor` unset. So for a legacy layer whose dots were, say, red:

- picker swatch shows `layer?.pulseColor || DEFAULT_HOTSPOT_COLOR` = default blue
- rendered dot resolves `layer.pulseColor` (unset) `|| hotspot.backgroundColor` = red

The control and the thing it controls disagree until the user touches it. The comment acknowledges this is deliberate, but a swatch that reports a colour the hotspot isn't using is a bug report waiting to happen. Seeding `pulseColor` (and `iconColor` when the first hotspot has one) from the same source would keep picker and preview honest.

### 5. Audio deleted-attachment detection false-positives on any failed lookup

```js
const isAudioDeleted = Boolean( id ) && isMediaResolved && ! audioMedia;
```

`hasFinishedResolution` is true whether the resolver succeeded, 404'd, 401'd, or hit a network error, so this also fires when the attachment merely isn't readable by the current user, or when the REST request failed transiently. The block then replaces the player, hides "Customize Audio", and states as fact that the file "was deleted from the Media Library", which is a scary and possibly wrong claim for a user who just has a flaky connection.

Worth either checking the resolution error (`getResolutionError` / the entity's last error) so only a genuine 404 triggers it, or softening the copy to something like "couldn't be loaded (it may have been deleted)". The recovery actions themselves are a nice touch.

### 6. Mobile audio card: `display: contents` drops the overflow guard

`display: contents` on `__body` promotes the children to grid items but also discards `&__body { flex: 1 1 auto; min-width: 0; }`. Grid items default to `min-width: auto`, and `__title` has no `overflow`/`text-overflow` of its own, so a long unbroken title (a filename with no spaces, for instance) can now blow out the `1fr` column and give the card horizontal overflow at =<480px. `__description` is safe because of its own `overflow: hidden`. Adding `min-width: 0` to `__title` (or to the promoted items) covers it.

Two smaller things in the same block:

- The `.godam-audio-player { grid-area: player; margin-top: 12px; }` rule sits at the top level, not scoped under `.godam-audio-card`, so it also retargets any player rendered outside a card at =<480px. Scoping it to `.godam-audio-card &` keeps it contained.
- `<Spinner />` is a sibling of the player inside `__body`, so during upload it becomes an unplaced grid item and auto-places into an implicit fourth row.

### 7. No tests, and the new pickers have no test ids

`hotspotStyle.js` is now the single decision point for four inputs across two style modes plus a legacy path, it is pure, and the repo already runs jest (`test:unit`, and there are existing specs like `layerActions.test.js`). This is close to the ideal unit-test target, and cases 1 to 4 above are all things a handful of assertions would have pinned down. Please add a spec covering: new pulse layer, new icon layer (glyph and custom image), icon mode with no icon, legacy pulse layer, legacy icon layer, and post-migration layer.

Also, the new "Icon" and "Background" `ColorPickerButton`s have no `data-test-id`, while the rest of the file consistently does (`godam-hotspot-control-*`, `godam-hotspot-button-*`). The audio error state does this correctly (`godam-audio-error-button-replace` / `-remove`), so it is just the two pickers.

### 8. CI is red, though not because of this PR

"Run JS unit tests" and "Run Plugin Check" both fail at `npm ci` in ~15s with `Missing: @opentelemetry/core@1.30.1 from lock file`. This PR does not touch `package.json` or `package-lock.json`; the drift came in with #2027 on `develop` (which added `overrides` entries without a matching lockfile regen) and this branch inherited it through the develop merge. #2042 fails identically.

Flagging it because the practical effect is that **no JS test or plugin check has actually run against this PR**. The lockfile regen belongs on `develop`, but this branch stays unverified until it lands and gets merged in.

---

### Summary

Items 1, 2 and 3 are behaviour changes on saved data that I'd want resolved (or consciously accepted and documented) before merge. Item 1 in particular is the one to make a decision on explicitly, since it repaints hotspots on live sites without anyone opening the editor. The rest are smaller.

For QA, the cases I'd exercise on a site with pre-2.0 content:
1. Legacy hotspot layer with a library glyph, viewed on the front end without opening the editor
2. Legacy hotspot layer with a custom uploaded image, same
3. Legacy pulse layer with a non-default colour: open the editor and compare the swatch to the dot
4. Migrated legacy icon layer: press "Remove icon" and confirm it actually clears
5. New layer: switch to "Icon" and pause before picking one
6. Audio block whose attachment was deleted, and separately one loaded as a user without permission on it
7. Audio card at 375px on all three surfaces (block editor, front end, customization preview), including a very long title

subodhr258 (approved):

PR #2038, opened 2026-07-24
Feat: Add layers in Image Block in Gutenberg

rudrakshi-gupta:
# Render GoDAM Image layers in the Gutenberg block editor

## Summary

The `godam/image` block's hotspot / WooCommerce product-hotspot layers were only drawn on the **published page** — the Gutenberg editor canvas previewed just the image (by the block's original design). This PR makes those layers render **inside the block editor canvas**, so authors see the actual hotspot / product overlays while editing, matching the published output.

It does this by extracting the front-end layer renderer into a shared module and reusing it from the block's `edit.js`, so the editor and front end draw layers with the exact same code.

Branch: `chore/add-layers-in-gutenberg` → `develop`

> **Companion:** WooCommerce **product** hotspots additionally need the `godam-for-woo` add-on to load its layer manager + product-hotspot CSS in the editor. That is a separate PR in the `godam-for-woo` repo (`enqueue_block_editor_assets` + a block-tied `wp_enqueue_block_style`). This PR renders core hotspot layers on its own; product hotspots light up once the companion is merged.

### New file
- `assets/src/js/godam-image-layers/render-image-frame.js` — the frame renderer (`initImageFrame` + the media-host adapter + hotspot/Woo manager wiring) extracted from `frontend.js` into a shared, side-effect-free module (no auto-run on import). Unchanged rendering logic; it just no longer owns the bootstrap.

### Modified files
- `assets/src/js/godam-image-layers/frontend.js` — now a thin bootstrap: imports `initImageFrame`, finds the server-rendered `.godam-image__frame[data-godam-image-layers]` frames and initializes them. Runs on `DOMContentLoaded` **or immediately if the DOM is already parsed**, and (in a page-builder preview only) re-runs on a few deferred passes + a `MutationObserver` so late-injected markup still hydrates.
- `assets/src/blocks/godam-image/edit.js` — the editor now:
  - fetches the selected attachment's authored layers from `rtgodam_meta` (exposed as a REST field on attachments) and keeps only drawable ones (hotspot layers with hotspots, Woo layers with product hotspots) — mirroring `render.php`;
  - when **Show image layers** is on and layers exist, renders the `.godam-image__frame` with `data-id` / `data-instance-id` / `data-godam-image-layers` + the overlay div via a React ref (the same markup `render.php` outputs);
  - draws them with `initImageFrame( frameRef )` in a `useEffect`, resetting the render guard + overlay so it redraws when the image, layers, or toggle change.

## How it works

The block editor canvas is an iframe; `edit.js`'s React output renders into it. The `useEffect` runs in the editor realm but draws onto the ref'd node inside the iframe (same-origin), using the shared renderer. Because it's the same renderer the front end uses, hotspot layers — and Woo product hotspots, when the Woo layer manager is registered (companion PR) — are drawn identically. Product data is fetched via `api-fetch`, which is available in the editor.

`initImageFrame()` is idempotent (guarded by `data-godam-layers-rendered`); the effect clears the overlay and resets the guard before each redraw so edits (new image, toggled layers) repaint cleanly.

## How to test

1. `npm run build:blocks` and `npm run build:js` (build artifacts under `assets/build/` are gitignored).
2. Gutenberg → add a **GoDAM Image** block → pick an image that has authored hotspot layers (from the GoDAM image editor) → ensure **Show image layers** is on → the hotspots should now render on the image in the editor canvas.
3. Toggle **Show image layers** off → overlays disappear. Replace the image → overlays update to the new attachment's layers.
4. Confirm the published page still renders layers (unchanged path).
5. With the `godam-for-woo` companion active, confirm **product** hotspots render in the editor too.

## Notes for reviewers

- Editor-only caveats: FontAwesome icon-style hotspots may appear unstyled in the canvas (FA injects into the parent `<head>`, not the iframe) — pulse/regular hotspots and product hotspots are unaffected; the front end is unaffected. Repeated redraws in the editor add a `resize` listener each time (small, editor-only).
- WooCommerce **product** hotspots require the `godam-for-woo` companion PR (editor enqueue of `godam-for-woo-layer-frontend` + block-tied product-hotspot CSS). Without it, core hotspot layers still render.
- **Sequencing:** this PR and `feat/image-block-wp-bakery` both modify `godam-image-layers/frontend.js`. Merge order matters — expect a small conflict in `frontend.js` (both add the editor re-init; this PR also relocates the renderer into `render-image-frame.js`). Recommend merging one, then rebasing the other.
- `assets/build/**` are gitignored build artifacts — rebuild rather than reviewing them. Run a full `npm run build` so split chunks regenerate with consistent names.

## Related PR (GoDAM for Woo)

rtcamp/godam-for-woo#183

## Reference

[Discussion with @KMchaudhary on Slack](https://rtcamp.slack.com/archives/D07CELU7Z6J/p1784807069874849)

## Demo Screenshots
Screenshot 2026-07-24 at 11 50 01 AM
Screenshot 2026-07-24 at 11 50 13 AM

### DIFF
--- a/assets/src/blocks/godam-image/edit.js
+++ b/assets/src/blocks/godam-image/edit.js
@@ -134,18 +134,16 @@ function ImageEdit( { attributes, setAttributes, isSelected } ) {
 	// image or layers change. The canvas is an iframe; the ref points at the real
 	// node inside it, and initImageFrame() draws directly onto that node — the
 	// same renderer the front end uses (Woo hotspots included, when the woo add-on
-	// registers its manager). We reset the render guard + overlay so edits redraw.
+	// registers its manager). initImageFrame is self-cleaning and returns a
+	// teardown that removes its resize listener / ResizeObserver / pending load
+	// handler and clears the overlay; returning it lets React tear down before
+	// each redraw and on unmount (e.g. toggling layers off), so nothing leaks.
 	useEffect( () => {
 		const frame = frameRef.current;
 		if ( ! frame || ! showLayers ) {
-			return;
+			return undefined;
 		}
-		frame.dataset.godamLayersRendered = '';
-		const overlay = frame.querySelector( '.godam-image-layer' );
-		if ( overlay ) {
-			overlay.innerHTML = '';
-		}
-		initImageFrame( frame );
+		return initImageFrame( frame );
 	}, [ showLayers, layersJson, url, instanceId ] );
 
 	// The inspector "Image Selection" panel mirrors the GoDAM Video block: an
--- a/assets/src/blocks/godam-image/edit.js
+++ b/assets/src/blocks/godam-image/edit.js
@@ -15,15 +15,17 @@ import {
 	useBlockProps,
 } from '@wordpress/block-editor';
 import { __ } from '@wordpress/i18n';
-import { useDispatch } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
 import { store as noticesStore } from '@wordpress/notices';
-import { useEffect } from '@wordpress/element';
+import { useEffect, useMemo, useRef } from '@wordpress/element';
 import { plus, trash } from '@wordpress/icons';
 
 /**
  * Internal dependencies
  */
 import { CustomizeVideoIcon } from '../godam-player/icons';
+import { initImageFrame } from '../../js/godam-image-layers/render-image-frame.js';
 import './editor.scss';
 
 const ALLOWED_MEDIA_TYPES = [ 'image' ];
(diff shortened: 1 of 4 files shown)


KMchaudhary:
## Combined code review — `godam#2038` + `godam-for-woo#183`

Reviewed both PRs together (editor-canvas layer preview + the Woo companion enqueues). The extraction of `render-image-frame.js` and the editor↔front-end parity are clean and well-commented, and the realm handling (parent-window managers drawing into the iframe node, `computeContentRect` diffing two same-realm rects) is correct. Findings below are ranked; most are cleanup/leak-class, with one narrow correctness race.

### 1. Redraw effect leaks listeners/observers/managers — `edit.js`
The `useEffect` that calls `initImageFrame()` returns **no cleanup**. Each redraw (image swap, layer edit, toggle) runs `window.addEventListener('resize', …)` + a fresh `ResizeObserver(...).observe(img)` inside `render()`, and nothing removes them. Toggling **Show image layers** off unmounts the frame but leaves the old manager's observer pinning the detached ``. Over an editing session this accumulates unboundedly. The PR notes acknowledge "adds a resize listener each time (small, editor-only)" — worth an actual cleanup (effect teardown, or have `initImageFrame` return a disposer) rather than a known leak.

### 2. Redraw racing image load can render a **stale** layer set — `edit.js` (narrow)
`initImageFrame` defers via `img.addEventListener('load', start, {once:true})` when the image isn't complete. If the effect re-runs before load fires, a second deferred handler is registered over new layers, but the shared `godamLayersRendered` guard means **the first pending handler wins** — so the older closure's layers render and the newer set is dropped. Resetting the guard to `''` doesn't cancel the earlier pending `load` handler. Conditions are narrow (rapid change while loading), but it's a latent correctness bug.

### 3. Builder-preview `MutationObserver` never disconnects — `frontend.js`
In a WPBakery preview the observer watches `document.body` `{subtree:true}` for the page's lifetime and, on every mutation batch, runs a whole-document `querySelectorAll` via `initAllFrames()`. Idempotent so not *wrong*, but it's continuous wasted work in an editor that mutates the DOM constantly. Consider disconnecting once frames are hydrated, or scoping the observer.

### 4. Woo layer script enqueued on **every** block-editor screen — `godam-for-woo/inc/class-bootstrap.php`
`enqueue_woo_layer_editor_assets` (on `enqueue_block_editor_assets`) enqueues `godam-for-woo-layer-frontend` + its `wp-data`/`wp-api-fetch` deps unconditionally — even editing a post with no GoDAM Image block. Mirror `render.php`'s lazy approach and gate on the block actually being present/used.

### 5. Front-end Woo CSS enqueue is redundant — `godam-for-woo/inc/class-bootstrap.php`
`wp_enqueue_block_style('godam/image', …'godam-for-woo-player-style'…)` also ties the CSS to the block on the **front end**, where `maybe_enqueue_woo_player_style()` already enqueues the same handle. WP dedupes by handle so it's harmless today, but it registers the same handle from two paths — if `src`/`deps`/`ver` ever diverge, first-registration wins silently. Only the editor-iframe injection is genuinely new.

---
*Nits / non-blocking:* the editor overlay omits the `id="layer-…-all"` that `render.php` emits (no consumer relies on it, so fine); FontAwesome-icon hotspots rendering unstyled in the canvas is already documented.

Overall: **no blockers.** #1 and #2 are the ones worth addressing before merge; #3–#5 are polish.

<sub>🤖 Combined review generated with Claude Code.</sub>

KMchaudhary (approved):

PR #2039, opened 2026-07-24
Feat: Upgate Audio Block in elementor

rudrakshi-gupta:
# PR Desc

Issue - rtCamp/godam-plugin-wp#21

The widget was already registered in Elementor, but it output a plain `<audio controls>` with a figcaption — none of the redesigned player (custom UI, cover, title/description, Chapters/Transcript panel). I aligned it with the `godam/audio` block the same way the video widget and WPBakery element were aligned:

- **Controls** now mirror the block inspector: audio file, optional Title / Description / Thumbnail overrides, Autoplay, Loop, Preload, and **Show Transcript** / **Show Chapters** toggles. (Dropped the old "Show Caption" switch — the redesigned block has no caption; it uses title/description.)
- **Render** reuses the block's `render.php` — the single source of truth shared by the block, the `[godam_audio]` shortcode, and the WPBakery element. Widget settings map into the block's camelCase attribute shape (`audioTitle`, `showTranscript`, etc.); title/cover/transcript/chapters fall back to the attachment's own data when left empty. It enqueues `godam-audio-view-script` + `godam-audio-style` (the block's handles) and sets `$godam_is_shortcode` so `render.php` skips `get_block_wrapper_attributes()` outside block context.
- The block's view script/style are also declared as `depended_script`/`depended_styles` so the editor preview picks them up.

To test: in Elementor, drop the **GoDAM Audio** widget, pick an audio attachment (ideally one with a transcript + chapters). You should see the redesigned player with cover/title/description and the Chapters/Transcript tabs, matching the Gutenberg block — plus the toggles hiding each panel, and title/cover falling back to the attachment when the override fields are blank.

## Demo

https://github.com/user-attachments/assets/c059b405-cbc6-4ed6-b76f-b3e789c8e5d2

### DIFF
--- a/assets/src/js/elementor/editor.js
+++ b/assets/src/js/elementor/editor.js
@@ -399,23 +399,62 @@ function refreshPanelControls( names ) {
  * @param {Object} values Map of setting keys to values.
  */
 function applyAudioSettings( values ) {
-	if ( activeAudioContainer && window.$e?.run ) {
-		window.$e.run( 'document/elements/settings', {
-			container: activeAudioContainer,
-			settings: values,
-		} );
-	} else if ( activeAudioSettings ) {
-		Object.keys( values ).forEach( ( key ) => activeAudioSettings.set( key, values[ key ] ) );
+	// The container ref is captured at hydrate; if Elementor rebuilt the element
+	// it can go stale and $e.run() may throw. Guard it (this runs from a fetch
+	// .then()/.catch(), so an unguarded throw would surface as an unhandled
+	// rejection) and fall back to a direct model update.
+	try {
+		if ( activeAudioContainer && window.$e?.run ) {
+			window.$e.run( 'document/elements/settings', {
+				container: activeAudioContainer,
+				settings: values,
+			} );
+		} else if ( activeAudioSettings ) {
+			Object.keys( values ).forEach( ( key ) => activeAudioSettings.set( key, values[ key ] ) );
+		}
+	} catch ( e ) {
+		try {
+			if ( activeAudioSettings ) {
+				Object.keys( values ).forEach( ( key ) => activeAudioSettings.set( key, values[ key ] ) );
+			}
+		} catch ( e2 ) {}
 	}
 
 	refreshPanelControls( Object.keys( values ) );
 }
 
 /**
+ * Apply only the fields the user has NOT edited since `snapshot` was captured,
+ * so a late REST populate never clobbers an in-progress Title / Description edit
+ * (the user may select an audio and immediately start typing an override).
+ *
+ * @param {Object} values   Proposed setting values.
+ * @param {Object} snapshot Field values captured when the populate started.
+ */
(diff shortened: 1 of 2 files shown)


KMchaudhary:
### Code review — correctness (medium effort)

Reviewed the diff for the Elementor GoDAM Audio widget upgrade. A few issues worth a look:

---

**1. `Browser default` preload option can never take effect (regression)** — `inc/classes/elementor-widgets/class-godam-audio.php:207`

The `preload` control offers `'' => Browser default`, but the value is passed through `render.php`, which coerces any empty preload to `metadata`:
```php
$godam_preload = ! empty( $attributes['preload'] ) ? esc_attr( $attributes['preload'] ) : 'metadata';
```
The old widget echoed `$preload` directly, so selecting *Browser default* actually omitted the attribute. Now that choice silently renders as `preload="metadata"`. Either drop the option or pass a sentinel render.php can distinguish from "unset".

---

**2. Auto-fill can clobber in-progress edits and defeats the "leave empty to use attachment title" fallback** — `assets/src/js/elementor/editor.js:445`

`populateAudioMeta()` fetches the attachment over REST and then `applyAudioSettings()` → `refreshPanelControls()` **re-renders** the Title/Description control inputs. If the fetch resolves after the user has started typing in *Audio Title* / *Description* (they select an audio, then immediately edit the field), the re-render discards the in-progress edit and resets the field to the fetched value.

Separately, because the JS always writes a concrete `audio_title`, the control is never actually empty — so `render.php`'s documented fallback (`if ( '' === $godam_audio_title ) $godam_audio_title = get_the_title(...)`, control help text: *"Leave empty to use the attachment title"*) is dead for widgets, and a later media-library rename won't propagate. Consider only filling empty fields, or not baking the title in at all and relying on render.php's fallback.

---

**3. External/remote audio loses its fallback `<source>` variants** — `inc/classes/elementor-widgets/class-godam-audio.php:202-213`

The old `render()` emitted multiple `<source>` entries from `$attachment['sources']` for non-numeric/remote attachments. The new mapping passes only a single `src`, and `render.php`'s virtual-attachment branch (`! $godam_attachment_id && ! empty( $godam_src )`) uses just that one URL. Remote attachments that supplied fallback source variants now render with no fallback.

---

**4. `applyAudioSettings` fallback path can throw an unhandled rejection** — `assets/src/js/elementor/editor.js:401-411`

`activeAudioContainer` is captured at hydrate time; if Elementor rebuilds the element the ref goes stale. `applyAudioSettings()` calls `window.$e.run(...)` without a guard, and it's invoked from the fetch `.catch(...)` as well — so if `$e.run` throws on a dead container, the failure escapes the promise chain (unhandled rejection) and the panel inputs aren't refreshed. Wrapping the `$e.run` call the way `refreshPanelControls()` already is would make it resilient.

---

*Nits:* `?? ''` / `?? 'metadata'` on `get_settings_for_display()` results are effectively dead — it returns `''`, never `null`, for unset text/select controls.

KMchaudhary (approved):

PR #2041, opened 2026-07-24
feat(analytics): Placements — player block_source emit + per-video section

subodhr258:
## What

godam side of **Placements** (rtCamp/godam-plugin-wp#22): the player emits `block_source` on analytics events, and the per-video Analytics page gains a **Placements** section. Consumes rtCamp/godam-analytics#237.

## Player emit

- `rtgodam_get_block_source_from_context()` maps the existing `godam_context` to the placement slug (`shoppable-video` / `wc-product-gallery` / `product-reels` / `reel-pop`; empty → `video-block`; unknown passes through). One mapper, no per-surface instrumentation.
- The player template stamps `data-block-source` (always) + `data-host-post-id` (embeds only) on the `<video>`; explicit `block_source`/`host_post_id` shortcode atts override (used by the embed page).
- `analytics.js`: type=1 batch entries are now `[videoId, jobId, blockSource]` triples; every type=2 path (track + keepalive teardown) carries `block_source`; type=3 flushes best-effort.
- **Gallery/iframe attribution:** gallery `view.js` appends `host_post_id` + `block_source=video-gallery` to the `video-embed` iframe URL; `video-embed.php` sanitizes + threads them; `buildAnalyticsRequestBody` overrides `post_id` with `hostPostId` (>0) so lightbox plays attribute to the hosting page.

## Placements section (per-video analytics)

- Block-source sections (auto-select first) → per-page rows: "Displayed on" (title + permalink), **Edit Page** (hidden without the edit cap; "Deleted page" fallback), **Views / Plays / Play Rate / Avg Watch Time** (derived client-side from raw primitives; full numbers).
- Own `DateRangePicker` (cache-key dedup via conditional spread), loading/error/empty states.
- Proxy enrichment: `fetch_analytics_data` adds `title`/`permalink`/`edit_url`/`is_deleted` per row (≤100 lookups); key stays absent on old microservices.
- Top Videos: new "Placements" column (`placements_count`).

## Testing

- Jest: Placements helpers (17) + picker suite (8) → **25 passed**. Lint clean; PHPCS (main-checkout vendor, worktree lacks vendor/) exit 0.
- Manual (needs #237 running): play a video via different surfaces (block / gallery lightbox / Woo gallery), run `/process-analytics`, open the video's analytics → sections group by surface, gallery plays attribute to the host page; check Top Videos "Placements" column.

## Automation impact
New test ids: `godam-placements-section`, `godam-placements-list-item-<block_source>`, `godam-placements-row`, `godam-placements-daterange-toggle`.

## Deploy order
godam-analytics#237 first (store) → this PR (emit + consume) → godam-for-woo#184 (Reel Pop tuple).

Rebased onto `develop` after #2028 merged, so the date-range plumbing is inherited (no vendored copy) and the card's picker is live end to end: `startDate`/`endDate` → `useFetchAnalyticsDataQuery` → `start_date`/`end_date` → microservice range-sums the placements. Verified against the dev site: a past window returns 0 placements, today returns 4.

## Review fixes

Two adversarial review rounds (8 dimensions, 3-lens refutation per finding).

**Round 1**
- `enrich_placements` gates on `is_post_publicly_viewable()` + `current_user_can('edit_post')`, so the public `/analytics/fetch` route no longer leaks private/draft/pending titles or permalinks to anonymous callers. Verified live: a private page returns "unavailable"/no permalink for anon, and the real title + Edit link for a capable admin.
- Rows past the 100-lookup cap get a constant-cost `Post #<id>` fallback (PHP) plus a `getPlacementTitle()` guard (JS), so no row renders nameless.
- The new Placements column was added to the Top Videos CSV export (header + row).

**Round 2**
- **Untrusted `block_source` as an object key (MED, security).** Values like `toString`, `valueOf` or `constructor` resolved to inherited `Object.prototype` members, making the label lookup truthy and the bucket a function — so `groupPlacementsByBlockSource` threw a `TypeError` inside the card's `useMemo` and **white-screened the whole per-video Analytics page**. Reachable by an anonymous caller (the embed page takes `block_source` from a query arg; the microservice normalizes but never rejects it). Buckets are now a null-prototype map and both the grouping and the label lookup use an own-property check.
- **Trashed pages (MED).** WordPress maps `edit_post` on a trashed post to its pre-trash status, so `current_user_can` passed and a trashed page rendered as a live row with an Edit link that `wp-admin/post.php` refuses to open (HTTP 409). Trashed posts now route to the unavailable state regardless of the edit capability.
- **Host-post attribution (MED, data-integrity).** `hostPostId` was read once per page and applied to every batched type=1 entry, so a single host-stamped player re-attributed every other video's page_load on that page. It is carried per queue entry now and the flush groups by it (`post_id` is one top-level field per request); the type=3 flush reads it from that video's own element.
- Unavailable rows keep their ID (`Post #12 (unavailable)`) so several of them stay distinguishable.
- `video-embed.php` uses `mb_substr` for the `block_source` cap — a byte-wise cut could split a multibyte character, after which `esc_attr()` dropped the value entirely.
- `enrich_placements` primes the post cache in one query instead of up to 100 uncached `get_post()` calls per request on a public endpoint.
- The empty state distinguishes "no placements in this date range" (with a **View all time** reset) from the plain no-data state, instead of showing the same message whenever the range is narrowed.

**Tests: 76 passed** (Placements helpers 33 incl. 13 hostile-key cases; emit-path helpers 19). ESLint clean; PHPCS exit 0; husky `lint:php` + `lint:js` pass on the staged diff.

**Emit path coverage:** `groupBatchByHostPostId` moved into `analytics-helpers.js` (analytics.js has top-level side effects, so it is not importable in jest) and covered by a new 19-case spec — `block_source` defaulting/passthrough, `post_id` overridden only when `hostPostId` parses > 0 (untouched for 0/null/undefined/negative/non-numeric), triple passthrough per event type, both bail conditions, and the grouping regression itself: entries with different host attributions land in separate requests.

## Automation impact
Test ids: `godam-placements-section`, `godam-placements-list-item-<block_source>`, `godam-placements-row`, `godam-placements-daterange-toggle`, `godam-placements-reset-range`.

## Deploy order
godam-analytics#237 first (store; run its alembic migration before the app restarts) → this PR (emit + consume) → godam-for-woo#184 (Reel Pop tuple + WC gallery attribution).

## No catch-all section, no historical-data note

The card shows only real, labelled placements. Anything it cannot attribute to a surface — an empty `block_source` (unattributed) or a slug with no label — is **dropped** rather than collected into an "Other" section: those plays are already reported in the video's overall metrics above the card, so repeating them here would double-report them.

- `groupPlacementsByBlockSource` drops unlabelled rows; the `'other'` key is gone from `SECTION_ORDER`, and `getBlockSourceLabel` returns `''` for anything unknown instead of falling back to "Other".
- Removed the "Placement analytics starts collecting after this update…" note. The no-data state is now simply **"No placement data for this video yet"**; the range-scoped empty state (with its **View all time** reset) is unchanged.
- Hostile `block_source` values stay safe: the null-prototype bucket map and own-property label check remain, and such rows are now dropped instead of bucketed (tests still assert no throw for `Object.prototype` key names).

The microservice excludes unattributed rows at the read layer too, so the per-video sections and the dashboard's Placements count agree.

## Demo Recording
https://app.godam.io/web/video/6821bb3iu9

### Added a note at the bottom:

Screenshot 2026-07-27 at 7 33 30 PM


KMchaudhary:
## Code review — Placements (#2041 + godam-for-woo#184, reviewed together)

Reviewed both halves of the feature end-to-end (emit → embed threading → proxy enrichment → render). Overall it's carefully built and well-tested — the null-prototype bucket map, per-entry `hostPostId` grouping, multibyte `block_source` cap, cache-key dedup, and the TopVideos column/CSV integrity all check out, and no page-breaking bug survived. Findings below, ranked.

### Security / info-disclosure
- **[MED · new] `enrich_placements` leaks metrics of non-public pages.** Title/permalink are gated for anonymous callers, but the row's `post_id` + `views/plays/page_load/play_time` still pass through the public `/analytics/fetch` route (`permission_callback => __return_true`). An anonymous caller can enumerate the private/draft page IDs a video is placed on, plus play counts. The "unavailable" branch rewrites the title but never strips the primitives. — `inc/classes/rest-api/class-analytics.php` (~L516)
- **[MED · pre-existing, same function] `post_details` is worse and ungated.** The sibling `post_views → post_details` enrichment returns real `get_the_title()` + `get_permalink()` for *every* post — private/draft/pending included — with no viewability/capability check, on the same public route. This PR adds exactly the protection to placements that this adjacent block still lacks; worth fixing them together. — `inc/classes/rest-api/class-analytics.php` (~L642)
- **[LOW] Attacker-controlled `block_source` rendered as an admin label.** `block_source` is accepted from the public embed page (`?block_source=`) and only length-capped (100), never allowlisted. An unmapped value is rendered verbatim as a section label in the admin Placements card. React escapes it (no XSS), but it's stored text / UI spoofing in the admin surface. — `pages/analytics/components/Placements/index.js` (`getBlockSourceLabel`, ~L65)

### Correctness
- **[LOW–MED · godam-for-woo#184] `get_the_ID()` for `host_post_id` in block-render filters.** `class-wc-featured-video-gallery-compat.php::get_video_embed_url()` (~L111) reads `get_the_ID()` inside `render_block_woocommerce/*` callbacks instead of the block context `postId` (`$block->context['postId']`). In FSE / query-loop contexts `get_the_ID()` may not be the product (or may be `false`), so lightbox plays misattribute — the same class of bug this PR fixes for the classic gallery. (The classic `class-wc-featured-video-gallery.php` path is safe because it runs inside the single-product loop.)
- **[LOW] Hardcoded version string.** The disclaimer reads `reflects activity since v2.1.0`, but the plugin is at `2.0.0` (`godam.php` / `RTGODAM_VERSION`). Nothing ties the string to the actual release version. — `pages/analytics/components/Placements/index.js` (~L474)

### Cleanup / altitude
- **[LOW] `getPlayRate(plays, pageLoad)` duplicates `calculatePlayRate(pageLoad, plays)`** from `pages/analytics/helper.js` with the arguments in **reversed** order — a real footgun for the next caller. Consider reusing the existing helper. — `Placements/index.js` (~L134)
- **[LOW] Host-post-id parsing is copy-pasted 4×** with two "none" encodings (`id > 0 ? id : null` in `elementHostPostId`/`getPageLoadVideoInfo` vs `parseInt(...) || 0` in the two `type=2` heatmap paths). Only harmless because `buildAnalyticsRequestBody` treats both as "no override". One shared helper would remove the divergence. — `assets/src/js/godam-player/analytics.js` (~L102)
- **[LOW] Shared `godam_context` patched with an override.** The featured-gallery lightbox reuses `godam_context='godam-video-product-gallery'` (core maps it to `shoppable-video`) and then compensates with an explicit `block_source='wc-product-gallery'`. Two placements sharing one context means every new emit site must remember the override or silently fold into `shoppable-video`; a 1:1 context→placement mapping would be more robust. — godam-for-woo `class-wc-featured-video-gallery.php` (~L357)

### Deploy-order note (not a code bug)
Both PRs convert **all** `type=1` tuples to 3-element triples, which older microservices reject with HTTP 400 — so shipping either plugin PR before **godam-analytics#237** breaks *all* `page_load` analytics, not just placements. The documented order (analytics#237 → #2041 → godam-for-woo#184) is mandatory; worth enforcing in release automation.

KMchaudhary (approved):

PR #2042, opened 2026-07-27
Feat: Add Image block in elemnetor

rudrakshi-gupta:
# Feat: Register GoDAM Image block as an Elementor widget

Issue - rtcamp/godam-plugin-wp#21

## Summary

Adds the **GoDAM Image** block to the Elementor page builder, bringing it to parity with the Gutenberg block, the `[godam_image]` shortcode, and the WPBakery element. The widget lets an author pick an image attachment and toggle whether the hotspot / product layers authored in the GoDAM image editor are overlaid, and it renders through the block's own `render.php` — so every surface (block, shortcode, WPBakery, Elementor) emits identical markup from a single source of truth.

It also makes those layers **draw live in the Elementor editor canvas** (not just on the published page), and gives the widget a proper icon and a friendly empty state.

## What changed

**New widget — `inc/classes/elementor-widgets/class-godam-image.php`**
- `Godam_Image` extends the shared `Base` widget: name `godam-image`, GoDAM category, `godam-eicon-image` panel icon, keywords `image / hotspot / shoppable`.
- **Controls** mirror the block's "Image Selection" inspector panel:
  - `image-file` — the `godam-media` picker (restricted to images).
  - `show_image_layers` — a switcher (default **on**, explicit `return_value => 'yes'`) that maps to the block's `showImageLayers`.
- **Render** maps the widget settings into the block attribute shape (`id`, `url`, `alt`, `showImageLayers`), enqueues the block's `godam-image-style` + `godam-player-style`, sets `$godam_is_shortcode`, and `require`s `assets/build/blocks/godam-image/render.php`. `alt` is read from the attachment's stored alt text (`_wp_attachment_image_alt`), matching how the block seeds `alt` on selection — there is intentionally no alt control (it isn't author-editable in the block inspector either).
- **Editor-only empty state**: instead of a bare, confusing box, an image-less widget shows a dummy-image placeholder ("Add Image Here" + prompt). Styling is a small self-contained, scoped `<style>` block (printed once per request) because the widget preview renders inside Elementor's preview iframe, where the plugin's editor stylesheet isn't loaded. Nothing is emitted on the front end.

**Registration + icon — `inc/classes/class-elementor-widgets.php`**
- Registered `Godam_Image` alongside the existing Video / Gallery / Audio widgets.
- Added `.godam-eicon-image` to the panel-icon CSS, pointing at `assets/src/images/godam-image-filled.svg` (the image SVG ships only under `assets/src/images/`, the same path the WPBakery element uses).
- Added `preload_image_layers_in_editor()` on `wp_enqueue_scripts`: in Elementor **preview mode** it registers/enqueues the layers front-end script + hotspot styles up front. Elementor re-renders a widget on edit via AJAX and doesn't inject newly-enqueued scripts into the running preview, so preloading is what lets a just-added GoDAM Image's layers draw without a full editor reload. Editor-only — the published front end still enqueues these lazily via `render.php`.

**Editor-canvas layer drawing — `assets/src/js/godam-image-layers/frontend.js`**
- Extended `isBuilderPreview()` to also detect the Elementor editor (the `elementor-preview-iframe` iframe id and the `elementor-editor-active` body class), in addition to the existing WPBakery detection. This enables the deferred re-init passes + `MutationObserver` inside the Elementor preview so layers are drawn on the widget Elementor injects/re-renders into its iframe (a one-shot init on load would miss it). The published front end is unaffected — no extra work is scheduled there.

## Why render through the block's `render.php`

Reusing the block template keeps a single rendering contract across the block, `[godam_image]` shortcode, WPBakery element, and now the Elementor widget. Any future change to the image markup, hotspot wiring, or lazy asset loading happens in one place and all four surfaces stay in sync.

## Design notes / scope

- **`alt` field**: omitted as a control (parity with the block/WPBakery), derived from the attachment.
- **Empty state**: intentionally re-implemented as scoped inline PHP rather than the block's React empty state, because it renders in the preview iframe where the editor stylesheet isn't available. Neutral light-theme palette.
- **Layers in editor** use the same editor re-init mechanism already proven for the WPBakery inline editor.

## Build

The `frontend.js` change requires a JS rebuild:

```
npm run build:js
```

The PHP changes (widget, registration, icon, preload) need no build. `assets/build/**` is gitignored, so reviewers must rebuild locally.

## Testing

1. Edit a page with Elementor → search "image" in the Elements panel → the **GoDAM Image** widget appears with its GoDAM icon.
2. Drop the widget → the empty state shows a dummy-image placeholder above "Add Image Here".
3. Pick an image (ideally one with hotspot and/or product layers authored in the GoDAM image editor):
   - The image renders in the canvas and the **layers draw live** in the editor.
   - Toggle **Show image layers** off/on → the overlay hides/shows.
4. Publish / view the page → image + layers render identically to the Gutenberg block.
5. Regression: confirm the Video / Gallery / Audio GoDAM widgets and their icons are unaffected, and that the published front end (non-editor) is unchanged.

## Related

Completes the page-builder rollout of GoDAM blocks (Video, Video Gallery, Audio, Document, Image) across WPBakery and Elementor.

## Demo

https://github.com/user-attachments/assets/526f09d5-37d3-4122-8f17-9a0eec92c217


KMchaudhary:
## Code review — `feat/image-block-elementor`

Reviewed the three source changes (`class-godam-image.php`, `class-elementor-widgets.php`, `frontend.js`) against the shared `render.php` contract, the WPBakery/shortcode path, and the sibling widgets. The widget correctly mirrors the block's attribute shape and the `showImageLayers` semantics, and the `isBuilderPreview()` extension won't false-positive on published pages. **No blocking correctness bugs found.** The items below are hygiene / DRY / cleanup, roughly in priority order.

### 1. `package-lock.json` carries large unrelated churn (incl. downgrades) — Medium
The lockfile diff has nothing to do with an Elementor widget: OpenTelemetry/Sentry/Babel entries gaining `dev`/`peer` flags and losing `license` fields, plus **dependency downgrades** — `markdown-it` 14.2.0 → 12.3.2 and `linkify-it` 5.0.2 → 3.0.3 (moved to nest under `markdownlint`). This looks like a regeneration artifact from a different npm/node version rather than an intentional change. It risks non-reproducible installs and unrelated dependency drift landing under a feature PR. Please revert `package-lock.json` to `develop` and re-run `npm install` with the repo's pinned Node (see `.nvmrc`) so only intended changes remain — ideally none.

### 2. Third copy of the `godam-image-layers-frontend` registration block — DRY / altitude
The exact "read `.asset.php` → `apply_filters( 'godam_image_layers_frontend_dependencies' )` → `wp_register_script(...)`" block now exists in **three** places: [`render.php`](assets/src/blocks/godam-image/render.php), `GoDAM_Image::register_image_layers_script()` in [class-godam-image.php](inc/classes/shortcodes/class-godam-image.php), and the new `preload_image_layers_in_editor()`. A future change to deps/version/handle has to be made in all three. Extract one shared registrar (e.g. a static helper) and call it from all callers.

### 3. Widget re-implements the shortcode render path instead of delegating — altitude
`Godam_Document::render()` delegates to `\RTGODAM\Inc\Shortcodes\GoDAM_Document::get_instance()->render( $atts )` — a single mapping/enqueue entry point. `Godam_Image::render()` instead duplicates the mapping + `wp_enqueue_style()` calls + `$godam_is_shortcode` + `require render.php` that `GoDAM_Image::get_instance()->render()` already does (`FILTER_VALIDATE_BOOLEAN` accepts `'yes'`, so the switcher value maps cleanly). Delegating here matches the Document widget and removes a second thing to keep in sync.

### 4. Redundant style enqueue in `render()` — simplification
`render()` calls `wp_enqueue_style( 'godam-image-style' )` / `'godam-player-style'`, but both are already declared in `set_default_config()`'s `depended_styles`, which Elementor enqueues via `get_style_depends()` whenever the widget is on the page. The explicit calls duplicate that (and the editor preview additionally preloads them). Drop the in-`render()` enqueues, or drop them from `depended_styles` — pick one source.

### 5. `preload_image_layers_in_editor()` loads assets on every preview render — efficiency (minor)
In Elementor preview mode it enqueues the layers JS + two stylesheets unconditionally, even on pages with no GoDAM Image widget. This mirrors the accepted WPBakery `preload_wpbakery_editor_assets()` pattern and is editor-only, so it's low priority — noting it for awareness.

---
*Verified as non-issues:* the `assets/src/images/godam-image-filled.svg` icon path (confirmed `assets/src/images` is **not** in `.distignore`, so it ships); `godam-image-style` / `godam-player-style` handles are registered at `init` (block registration) before the `wp_enqueue_scripts` preload runs; the empty-state `static $style_printed` once-per-request guard; and the `showImageLayers` boolean mapping vs. `render.php`'s `! empty()` default-on logic.

KMchaudhary (approved):

PR #2051, opened 2026-07-29
GoDAM - VideoJet Support

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam-plugin-wp/issues/25, and other bugs mentioned in this [sheet](https://docs.google.com/spreadsheets/d/1o_cRlGipRk1iHw-Fwwdx1CDZvs7AHE4X/edit?gid=150154066#gid=150154066)

This pull request introduces a new "Play on modal" feature for the GoDAM video player, allowing videos to open in a lightbox modal when clicked, and improves the media library's transcoding UI logic for better compatibility with other plugins. It also adds guide message support to the thumbnail panel and refines related UI styling.

**GoDAM Player: "Play on modal" feature**
- Added a new `playOnModal` block attribute and toggle control in the editor, enabling videos to open in a centered modal/lightbox overlay when clicked (`block.json`, `edit-common-settings.js`, `godam-player.scss`, `playerManager.js`, `modalManager.js`). [[1]](diffhunk://#diff-0452960eaa8ba8724cc49af281678ab3c7d21abd9ed841f2362d9aa0e7b14d3dR130-R133) [[2]](diffhunk://#diff-da8e9c5db56f46907c50188af390d920ed08982cb6174c610e6b097e9eed2b8cL25-R25) [[3]](diffhunk://#diff-da8e9c5db56f46907c50188af390d920ed08982cb6174c610e6b097e9eed2b8cR85) [[4]](diffhunk://#diff-da8e9c5db56f46907c50188af390d920ed08982cb6174c610e6b097e9eed2b8cR129-R137) [[5]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79R2508-R2603) [[6]](diffhunk://#diff-158210cc4e3e8543c29c89cd5005c40e3e01ebcc4955b68c0a21f4b97737f63bR21) [[7]](diffhunk://#diff-158210cc4e3e8543c29c89cd5005c40e3e01ebcc4955b68c0a21f4b97737f63bR41-R58) [[8]](diffhunk://#diff-158210cc4e3e8543c29c89cd5005c40e3e01ebcc4955b68c0a21f4b97737f63bR80-R84) [[9]](diffhunk://#diff-158210cc4e3e8543c29c89cd5005c40e3e01ebcc4955b68c0a21f4b97737f63bR166-R168) [[10]](diffhunk://#diff-7d93261fd8f31f07893378082ffdd8f7b53ff984faf302bbb99769492c097de0R1-R184)
- Implemented modal overlay, close button, and keyboard accessibility; the inline player is moved into the modal and restored on close. [[1]](diffhunk://#diff-7d93261fd8f31f07893378082ffdd8f7b53ff984faf302bbb99769492c097de0R1-R184) [[2]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79R2508-R2603)

**UI Improvements: Thumbnail panel guide message**
- Added support for an optional, sanitized guide message (with markup) in the video thumbnail panel, styled for clarity in both the block editor and media library. [[1]](diffhunk://#diff-8ef389641b0e223fd7a8a6153e70bf26c93a292604bb3a587651b14142921ab7R1-R5) [[2]](diffhunk://#diff-8ef389641b0e223fd7a8a6153e70bf26c93a292604bb3a587651b14142921ab7R137-R155) [[3]](diffhunk://#diff-2da561cc60d267b68baaad3869d1f5fc8b60db4bbc15584d15a1ae71f1e8e7fdR130-R144) [[4]](diffhunk://#diff-44882211d1c439639de320a8abf16a2c5f71960f24f7c39c5658220ec0f8da4dR559-R574)

**Media Library: Transcoding UI logic**
- Refactored logic to ensure GoDAM's transcoding status badges and retranscode bulk action remain available on the media grid page (`upload.php`), even when folder organization is disabled (additive mode), while keeping other custom UI minimal for compatibility with other DAM plugins. [[1]](diffhunk://#diff-c96ea6334c321bd29a6e93f99ac0b39070d7d6995c8acd4497cfff409f37bfb0L28-R28) [[2]](diffhunk://#diff-c96ea6334c321bd29a6e93f99ac0b39070d7d6995c8acd4497cfff409f37bfb0L246-R251) [[3]](diffhunk://#diff-9ee098078438429f0c7dcb498e01c5a558ff80439d9e11e7ede6b8b71fe38474R62-R77) [[4]](diffhunk://#diff-9ee098078438429f0c7dcb498e01c5a558ff80439d9e11e7ede6b8b71fe38474L176-R192) [[5]](diffhunk://#diff-fb7c950b71c9461563537f8c0937e81f6ff688496066c97a24ae6d3d6e0fb1deL40-R44) [[6]](diffhunk://#diff-fb7c950b71c9461563537f8c0937e81f6ff688496066c97a24ae6d3d6e0fb1deR66)

**Styling and Accessibility**
- Added and updated CSS for the modal overlay, close button, and guide messages to ensure a consistent and accessible user experience. [[1]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79R2508-R2603) [[2]](diffhunk://#diff-2da561cc60d267b68baaad3869d1f5fc8b60db4bbc15584d15a1ae71f1e8e7fdR130-R144) [[3]](diffhunk://#diff-44882211d1c439639de320a8abf16a2c5f71960f24f7c39c5658220ec0f8da4dR559-R574)

**Other minor improvements**
- Added stylelint comments to clarify and suppress linter warnings for intentional CSS selector usage. [[1]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79R1146) [[2]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79R2323)

**1. Custom thumbnail guide message**
Screenshot 2026-07-29 at 4 07 02 PM

Screenshot 2026-07-29 at 4 11 38 PM

**2. Re-add the support transcoding status and re-transcode media action**

https://github.com/user-attachments/assets/863d36d0-73e0-463c-a2df-416c65b497cc

**3. Add Lightbox support for GoDAM Video block, shortcode, as well as elementor and wpbakery blocks**

https://github.com/user-attachments/assets/fbb1b90b-69b6-4509-b915-d63607eb5d81


assets/src/css/godam-player.scss
```diff
@@ -2503,3 +2505,99 @@ body.godam-share-modal-open {
...
+}
+
+.godam-player-modal-video {
+	width: 100%;
+}
+
+/* The moved player root carries its own inline max-width (from Player Height);
+ * neutralise it so the player fills the modal width instead. The class is added
+ * to the player root by modalManager.js while it lives inside the modal. */
+.godam-player-modal-item {
```
subodhr258, line 2667: **Blocking:** `max-width: 100% !important` here discards the height-derived `max-width` the template computes, so the player's height comes only from `aspect-ratio` on `.easydam-player.video-js`. The `max-height: 90dvh` on the wrapper cannot rein it in, because the wrapper is `overflow: visible` and the child is never height-constrained.

Measured in Chrome at 1470x836 with this CSS verbatim: 16:9 renders 1000x563 (fine), 4:3 renders 1000x750 (fine), 1:1 renders 1000x1000 and overflows the viewport by 164px, 9:16 renders 1000x1778 and overflows by 942px with the control bar at y ~1780, entirely below the fold.

Letting the height cap drive the width fixes it:

```scss
.godam-player-modal-item {
	width: auto !important;
	max-width: 100% !important;
	max-height: 90dvh;
	margin: 0 auto !important;
	aspect-ratio: var(--rtgodam-video-aspect-ratio, 16 / 9);
}
```

The custom property is already on the moved root, so tall videos size off `max-height` and wide ones off the wrapper width.

subodhr258 (commented):
## Review

Nice, well-commented PR. The "Play on modal" approach (move the already-initialised inline player into a shared lightbox, move it back on close) is a good call: it avoids a second player instance and keeps analytics/layers/engagement on one object. The additive-mode carve-out for transcoding UI is correctly scoped, and the guide-message filters are cleanly documented.

Two issues I would want resolved before merge, plus some accessibility and scoping points.

### Verification I ran

- `Run PHPCS inspection` passes. `Run JS unit tests` and `Run Plugin Check` are red, but both die at `npm ci` with lockfile drift, which is the known repo-wide breakage on `develop`, not this PR.
- ESLint and Stylelint are clean on every file this PR touches (the handful of errors I hit locally all sit on pre-existing lines).
- `DOMPurify` is already imported in `attachment-detail-two-column.js`, `isomorphic-dompurify` matches the repo convention, and `videojs.getPlayer( element )` matches existing usage in `videoPlayer.js` / `analytics.js` / `playerManager.js`. All good.
- I checked the additive-mode claim in the `isMediaTranscodingUIEnabled()` comment: `attachment.js:39` does gate drag-to-folder on `isFolderOrgDisabled()`, so opening up `wp.media.view.Attachments` on `upload.php` does not leak folder behaviour. Claim holds.

---

### 1. Modal overflows the viewport for anything taller than ~4:3 (blocking)

`.godam-player-modal-item` forces `max-width: 100% !important`, which discards the height-derived `max-width` the template computes for Player Height. The player's height then comes only from `aspect-ratio` on `.easydam-player.video-js`. `max-height: 90dvh` on `.godam-player-modal-wrapper` does nothing here, because the wrapper's `overflow` is `visible` and the child is never height-constrained, so the video just overflows past it.

Measured in Chrome at 1470x836, using the PR's CSS verbatim:

| Aspect ratio | Rendered in modal | Overflows viewport by |
|---|---|---|
| 16:9 | 1000 x 563 | 0 |
| 4:3 | 1000 x 750 | 0 |
| 1:1 | 1000 x 1000 | 164px |
| 9:16 | 1000 x 1778 | 942px |

At 9:16 the control bar lands at y ≈ 1780, entirely below the fold, so the video is unplayable and uncloseable except via Escape. Vertical/reel video is a first-class GoDAM use case, so this will get hit.

The fix is to let the height cap drive the width instead of pinning width to 100%:

```scss
.godam-player-modal-item {
	width: auto !important;
	max-width: 100% !important;
	max-height: 90dvh;
	margin: 0 auto !important;
	aspect-ratio: var(--rtgodam-video-aspect-ratio, 16 / 9);
}
```

The custom property is already on the moved root, so `aspect-ratio` plus `max-height` plus `width: auto` will size the box off the height cap for tall videos and off the wrapper width for wide ones. Worth a quick check at 9:16, 1:1 and 16:9 on both a tall and a short viewport.

### 2. Share modal renders behind the player modal (blocking, if global share is on)

`.share-modal-container` is `z-index: 9999` (`godam-player.scss:663`) and `shareManager.getModalParent()` appends it to `document.body` outside fullscreen. The new overlay is `z-index: 100000` and the wrapper `100001`, also on `document.body`, so they compare directly in the root stacking context.

Once the player is live inside the lightbox, clicks pass through to the Video.js controls, so the share button is reachable. Pressing it puts the share UI underneath the lightbox backdrop: the user sees the dark overlay and nothing else, with body scroll locked by both. Reproducible whenever `enableGlobalVideoShare` is on and the video has the share button.

Either raise `.share-modal-container` above the lightbox, or have `getModalParent()` return the modal wrapper when the player is inside it.

### 3. Close button is unreachable for screen readers

`wrapper` gets `role="dialog"` and `aria-modal="true"`, but `closeBtn` is appended to `document.body` as a sibling of the wrapper, not inside it (`modalManager.js:92`). `aria-modal="true"` hides everything outside the dialog from the accessibility tree, so AT users have no Close control at all; Escape is the only way out.

Move the button inside `wrapper` and position it with CSS. While you are there, the dialog has no accessible name: `aria-label={ __( 'Video player', 'godam' ) }` or `aria-labelledby` pointing at the video title would fix it.

### 4. No focus trap

Focus is moved into the wrapper on open and restored on close, which is the right shape, but Tab still walks straight out of the dialog into the page behind it. Given `aria-modal="true"` is already claiming the dialog is modal, a Tab/Shift+Tab cycle in `handleKeydown` would make that true.

---

### Worth discussing

**Overlay inner blocks get swallowed.** The capture-phase `preventDefault()` / `stopPropagation()` is bound to the whole player root, which contains `.godam-video-overlay-container` and its author-placed inner blocks. With Play on modal enabled, a "Buy now" button or link in the video overlay silently stops working and opens the lightbox instead. If that is intended, it is worth saying so in the toggle's `help` text. If not, skip the intercept when `event.target.closest( '.godam-video-overlay-container' )` matches.

**Autoplay + Play on modal is unresolved.** With both on, the video autoplays inline (muted) and a click then re-plays it in the lightbox. The template already resolves the analogous conflict one line up (`if ( $godam_autoplay ) { $godam_hover_select = 'none'; }`), so there is a precedent for picking a winner here rather than leaving both active.

**No tests.** `modalManager.js` is 184 lines of DOM move-and-restore with an obvious failure mode: the player not landing back in its original position. The repo already runs JS unit tests next door (`assets/src/js/godam-player/analytics-helpers.test.js`, `utils/layerActions.test.js`), so an open then close then "DOM is byte-identical to before" assertion in jsdom would be cheap insurance.

**Missing E2E hooks.** The toggle correctly got `data-test-id="godam-video-control-play-on-modal"`, but the modal DOM it drives (overlay, wrapper, close button) has none, so there is nothing stable to assert against.

**iOS QA.** Re-parenting a live `<video>` can reset playback state in Safari/iOS. The open path orders the move before `play()` and the close path pauses before moving, which is the right order, but this is worth an explicit check on a real device.

### Nits

- `isMediaTranscodingUIEnabled()` also gates `wp.media.view.Attachments` (column-width logic and the custom `AttachmentView`), which is wider than "transcoding UI". The behaviour is right, the name just undersells the scope.
- `rtgodam_video_block_video_thumbnails_guide_message` and `rtgodam_media_library_video_thumbnails_guide_message` are near-identical. One filter with a `$context` argument would be easier to document, though two is defensible if you expect the copy to diverge.
- The guide message's only sanitisation is client-side. A `wp_kses_post()` on the filter output would mean the trust boundary does not rest entirely on JS. Low risk since only code can set it, but it is one line.
- The two `stylelint-disable-next-line` comments are reasonable, but `no-duplicate-selectors` firing on `.godam-video-wrapper` (three separate blocks now, at 1747, 2324 and 2515) reads more like the file wanting a consolidation pass than a rule wanting a silence.

assets/src/css/godam-player.scss
```diff
@@ -2503,3 +2505,99 @@ body.godam-share-modal-open {
...
+/* Prevent the page from scrolling behind an open modal. */
+body.godam-player-modal-open {
+	overflow: hidden;
+}
+
+/* Dark backdrop. */
+.godam-player-modal-overlay {
+	position: fixed;
+	inset: 0;
+	z-index: 100000;
```
subodhr258, line 2535: **Blocking (when global share is on):** `.share-modal-container` is `z-index: 9999` (line 663) and `shareManager.getModalParent()` appends it to `document.body` outside fullscreen, same as this overlay. At `100000` the lightbox backdrop sits on top of it.

Once the player is live in the modal, clicks pass through to the Video.js controls, so the share button is reachable. Pressing it shows the user a dark overlay and nothing else, with body scroll locked twice over.

Either raise `.share-modal-container` above the lightbox, or have `getModalParent()` return the modal wrapper when the player is inside it.

assets/src/js/godam-player/managers/modalManager.js
```diff
@@ -0,0 +1,184 @@
...
+		wrapper.setAttribute( 'role', 'dialog' );
+		wrapper.setAttribute( 'aria-modal', 'true' );
+		wrapper.setAttribute( 'tabindex', '-1' );
+
+		const content = document.createElement( 'div' );
+		content.className = 'godam-player-modal-video';
+		wrapper.appendChild( content );
+
+		document.body.appendChild( overlay );
+		document.body.appendChild( closeBtn );
```
subodhr258: The close button is appended to `document.body` as a sibling of `wrapper`, but `wrapper` carries `aria-modal="true"` (line 84). That attribute hides everything outside the dialog from the accessibility tree, so screen reader users have no Close control at all and Escape becomes the only exit.

Moving `closeBtn` inside `wrapper` and positioning it with CSS keeps the same visual result and puts it back in the a11y tree. The dialog is also unnamed, so `aria-label={ __( 'Video player', 'godam' ) }` (or `aria-labelledby` pointing at the video title) would be worth adding on the same line.

assets/src/js/godam-player/managers/modalManager.js
```diff
@@ -0,0 +1,184 @@
...
+		// so an inline click opens the modal instead of playing inline.
+		playerRoot.addEventListener(
+			'click',
+			( event ) => {
+				// Once the player is live inside the modal, let clicks through to
+				// the Video.js controls.
+				if ( this.activeEntry && this.activeEntry.video === video ) {
+					return;
+				}
+				event.preventDefault();
```
subodhr258, line 217: This intercepts every click anywhere in the player root, which includes `.godam-video-overlay-container` and its author-placed inner blocks. With Play on modal enabled, a link or button an author dropped into the video overlay silently stops working and opens the lightbox instead.

If that is the intended behaviour it is worth stating in the toggle's `help` text. If not, an early bail on `event.target.closest( '.godam-video-overlay-container' )` would keep the overlay interactive.

assets/src/js/godam-player/managers/modalManager.js
```diff
@@ -0,0 +1,184 @@
...
+		}
+		this.lastFocused = null;
+	}
+
+	/**
+	 * Close on Escape.
+	 *
+	 * @param {KeyboardEvent} event - Keydown event.
+	 */
+	handleKeydown( event ) {
```
subodhr258, line 742: Escape closes and focus is restored on close, which is the right shape, but Tab still walks straight out of the dialog into the page behind it. Since `aria-modal="true"` is already asserting the dialog is modal, a Tab / Shift+Tab cycle here would make that assertion true.

inc/classes/class-assets.php
```diff
@@ -187,14 +187,30 @@ public function admin_enqueue_scripts() {
...
+		 * godam/video block Inspector (block editor).
+		 *
+		 * Empty by default. Add-ons/extensions can hook in to surface guidance about
+		 * video thumbnails (how they are generated, links to docs, upgrade prompts,
+		 * etc.). The returned HTML is sanitized client-side before it is rendered, so
+		 * only a safe subset of markup survives. Return an empty string to show nothing.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param string $message Guide message HTML. Default empty string.
```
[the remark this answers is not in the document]
subodhr258: Disregard this one: `@since n.e.x.t` is the intended state during development in this repo. It is a WordPress-standard placeholder that gets resolved to the real version at release time (step 7 of the release checklist), which is exactly what stops a wrong version number from leaking into the docs. Please leave both occurrences as `n.e.x.t`.

KMchaudhary:
**Elementor:**

https://github.com/user-attachments/assets/a9afab1e-6e12-44d6-9898-7eb4460a4d18

**Gutenberg block**

https://github.com/user-attachments/assets/722e41a2-4918-4e46-86c2-a98b04bb305f

subodhr258 (commented):
## Re-review (`181b4697` + `9ffbc48d`)

Both blockers from the last pass are genuinely fixed: `.godam-player-modal-item` now sizes from the height cap (`width: auto` + `max-height`), so 1:1 and 9:16 stay on screen, and `body.godam-player-modal-open .share-modal-container { z-index: 100003 }` puts share back above the lightbox. Close button is inside the dialog, focus trap and the overlay-container bail are in. The trigger/deep-link layer is a good addition and it is well documented.

Verified locally: 126 unit tests pass, ESLint and Stylelint clean on every file touched, PHPCS clean on all changed PHP.

Three things:

**1. `@since 2.0.1` should be `n.e.x.t`** ([class-godam-player.php#L252](https://github.com/rtCamp/godam/blob/9ffbc48d/inc/classes/shortcodes/class-godam-player.php#L252))

The two new filters in `class-assets.php` use `n.e.x.t`, which is the convention here (the placeholder is resolved at tag time, step 7 of the release checklist). Current version is 2.0.0, so 2.0.1 is guessing the next release.

**2. Cmd/Ctrl+click on an `<a href>` trigger is swallowed** ([lightboxTriggers.js#L111](https://github.com/rtCamp/godam/blob/9ffbc48d/assets/src/js/godam-player/lightboxTriggers.js#L111))

The delegated handler calls `preventDefault()` unconditionally, so the anchor pattern the README recommends cannot be opened in a new tab:

```html
<a href="/?godam_page=video-embed&id=1641" data-godam-lightbox="1641">Watch from 0:42</a>
```

Confirmed in jsdom: a `click` with `metaKey: true` gives `defaultPrevented = true` and opens the lightbox in place. An early bail keeps the browser's own behaviour for modified clicks (middle-click is already fine, it fires `auxclick`):

```js
if ( event.metaKey || event.ctrlKey || event.shiftKey || event.altKey ) {
	return;
}
```

**3. `role="button"` on the player root hides the Video.js controls from assistive tech** ([modalManager.js#L70-L78](https://github.com/rtCamp/godam/blob/9ffbc48d/assets/src/js/godam-player/managers/modalManager.js#L70-L78))

`button` is one of the roles with presentational children, so browsers prune its subtree from the accessibility tree. Once the player is moved into the lightbox, a screen reader user gets a single "Play video in a lightbox" button and no play/pause, volume, fullscreen or share. `trapFocus()` still tabs onto those controls, since `querySelectorAll` sees the DOM rather than the a11y tree, so focus lands on nodes that are never announced.

The button semantics are only wanted while the root is a click-to-open poster, so removing `role`/`tabindex`/`aria-label` in `openElement()` and restoring them in `close()` would fix it without changing the inline behaviour.

subodhr258:
All three points from the last review are fixed, and the poster-semantics one is handled better than I suggested: recording which attributes `applyPosterSemantics()` actually added means suspend/resume can never strip an author's own `role` or `aria-label`. The touch/iOS commit holds up too, `close()` releasing the custom-fullscreen scroll lock and Escape unwinding one layer at a time were both worth catching.

**Correcting my first version of this comment:** I called the red `Check the committed POT matches a fresh build` a blocker. It is not, please ignore that. This PR does not touch `languages/` at all, and the POT is only regenerated just before a release, so that check goes red on any PR that adds or moves a string (#2066 is red for the same reason). Nothing to do here.

Verified on `28966a89`: 221 JS tests pass, PHPCS clean on all changed PHP, and JS unit tests / PHP unit tests / PHPCS / Plugin Check are all green. The ESLint and Stylelint errors I hit locally are all on pre-existing lines, nothing this PR touched.

Two non-blocking notes:

- `godam-player-frontend-script` has no `wp_set_script_translations()` call, so the new `__()` strings in `modalManager.js` will not translate on the frontend even once the POT is regenerated at release. Pre-existing (`shareManager.js` has the same gap with `Check out this video!`), so not yours to fix here, just worth knowing the four new strings land on an unwired handle.
- The description still describes "Play on modal" / `playOnModal`. Worth refreshing before merge, mainly because the PR grew a public API that is not mentioned at all: `data-godam-lightbox` triggers, `#godam-video-{id}` deep links, `GoDAMAPI.openLightbox()` and the `rtgodam_enqueue_lightbox_runtime` filter.

Looks good to me otherwise.

subodhr258 (approved):

PR #2052, opened 2026-07-29
Fix GoDAM 2.1 QA: non-PDF documents, audio overflow, stray media icon, Image rename

subodhr258:
## Summary

Fixes five items from the GoDAM 2.1 QA tab plus one naming request from Slack. All were reproduced locally on Elementor 4.2.1 + WPBakery (js_composer) 9.0-beta.3 + WooCommerce before any code was written.

| QA | Issue | Status |
|----|-------|--------|
| 3, 6 | Document looks blank in editor and on frontend | Fixed |
| 5 | WPBakery page opens a download popup on refresh | Fixed (same root cause as 3/6) |
| 2 | Audio block overlap in 3-column layout | Fixed (was still reproducible, not already done) |
| 7 | Thumbnail displayed with separate icons | Fixed |
| Slack | "GoDAM Image" should read "Image" | Done (core, Elementor, WPBakery) |

## QA 3, 5, 6: non-PDF documents

`render.php` emitted `<object type="application/pdf">` pointing at whichever file was selected. Given a non-PDF the browser receives a type it cannot display, and the outcome splits by browser setting:

| Browser state | Result | Reported as |
|---|---|---|
| PDF viewer enabled | object collapses to 0px wide, inner content empty, and the `<object>` fallback is suppressed, so nothing at all is painted | QA 3, 6 "looking blank" |
| PDF viewer disabled ("Download PDFs") | the file is handed to the download manager | QA 5 "download popup" |

Evidence for QA 5: loading the page, with no click at all, wrote a file into `~/Downloads` that was byte-identical to the test `.docx` (945 bytes, same zip entries, same body text).

PDF is the only format this block supports, so `godam_is_supported_document()` now gates output. The attachment's stored MIME type is authoritative, with the URL extension as a fallback for GoDAM tab media and URL-only documents.

The Gutenberg block, the `[godam_document]` shortcode (WPBakery) and the Elementor widget all render through the same `render.php`, so one guard covers all three front ends. Because rendering nothing would otherwise leave an author staring at an empty element, each editor now shows an "unsupported format" notice instead.

Also closed a hole in the block's own selection guard: it read `if ( mime && mime !== 'application/pdf' )`, so a missing MIME type passed straight through. It now tests for a positive PDF match and falls back to the extension.

## QA 2: audio overlap in 3 columns

`.godam-audio-player` is a `nowrap` flex row with a 14px gap. `__scrubber` had `flex: 1 1 auto` but no `min-width: 0`, and `input[type=range]` carries a ~129px intrinsic min-content width, so the track refused to shrink:

```
56px play + 129px scrubber + 28px time + 28px gaps = 241px  in a 198px card body
```

The 43px excess pushed the non-shrinking `__time` label outside the card and 8.1px past the Elementor column. It only shows in 3-column layouts because wider columns have slack. Adding `min-width: 0` shrinks the track to 86px and the row fits; the same file already does exactly this on `__body` and `__play`.

## QA 7: stray icon over the thumbnail

Elementor absolutely centres `.eicon-video-camera` over the media area as its empty-state affordance and never hides it, because its own control only ever sets `src` on the preview `<video>`. Our control additionally sets `poster` from the GoDAM thumbnail, which left that icon stranded on top of a real image (measured 21x21, visible, dead centre over the 258x145 poster, with `e-media-empty` absent). The icon is now hidden while a poster is showing and kept for the empty state.

Worth noting for reviewers: the markup in `class-godam-media.php` is byte-identical to Elementor's native control and we add no CSS of our own, so this is only a bug for us because we are the ones adding a poster.

## Rename

"GoDAM Image" to "Image" in core, Elementor and WPBakery, matching Video / Audio / Document / Video Gallery / Shoppable Video, which already drop the prefix inside the GoDAM category.

One behaviour change to be aware of: typing "godam" into WPBakery's Add Element search no longer matches this element, because no GoDAM element now carries "GoDAM" in its name. The GoDAM category tab remains the discovery path and lists all six correctly. Say the word if you would rather keep it searchable.

## Testing

`tests/php/DocumentSupportTest.php` covers the new helper, which is the single gate protecting the embed: MIME beating a misleading `.pdf` URL, other media types, the URL fallback, query strings and fragments, a deleted attachment falling through to its URL, malformed input, and "pdf" appearing outside the extension.

```
OK (31 tests, 71 assertions)
```

Confirmed the tests can fail: mutating the helper to always return `true` fails all 7.

`tests/bootstrap.php` gains a `wp_parse_url` stub, WP time constants, and a require of `custom-functions.php`.

## Manual verification steps

Needs both Elementor and WPBakery active, plus a non-PDF (for example a `.docx`) and a PDF in the media library.

1. **Frontend, non-PDF**: put the `.docx` in a Document block, a `[godam_document]` shortcode and an Elementor Document widget. Each should render nothing at all, and no download should start on load or refresh. Check with the browser's PDF viewer both enabled and disabled (`chrome://settings/content/pdfDocuments`), since the old bug presented differently in each.
2. **Frontend, PDF (regression)**: same three surfaces with the PDF still embed and display normally.
3. **Gutenberg editor**: the `.docx` block shows "Unsupported file format" with the file name instead of a blank embed. Selecting a non-PDF from the library shows the snackbar warning and does not attach.
4. **Elementor editor**: the Document widget shows the "Unsupported file format" panel. A valid PDF is unaffected.
5. **WPBakery frontend editor**: the Document element shows the unsupported notice. The backend editor still shows its normal parameter summary.
6. **QA 2**: an Elementor 3-column section with a GoDAM Audio widget per column. The duration label stays inside each card at desktop, tablet and mobile widths, and in the editor canvas.
7. **QA 7**: select a transcoded video with a thumbnail in an Elementor GoDAM Video widget. The Video File preview shows the thumbnail with no icon on top. Clearing the media brings the camera icon back.
8. **Rename**: `/godam` in the Gutenberg inserter, the Elementor widget panel, and WPBakery's Add Element > GoDAM tab all read "Image".

## Automation impact

New hooks for E2E:

- `godam-pdf-content-unsupported` (Gutenberg canvas notice)
- `godam-document-elementor-unsupported` (Elementor editor notice)
- `godam-document-unsupported` (WPBakery front-end editor notice)

No existing `data-test-id` was renamed or removed. Existing document selectors are unchanged for PDFs; suites that assert on a rendered non-PDF document will now correctly find no output.


subodhr258:
Superseded by #2053. The branch was renamed to `fix/godam-2.1-qa-items` (the old name read as "audio icon", which is not a thing in this change); GitHub closed this PR instead of retargeting it. Same commit, same diff.

PR #2060, opened 2026-08-05
Fix/godam media library issues

rudrakshi-gupta:
<html><head></head><body><h1>Media Library — Improvements &amp; Bug Fixes (Epic #23)</h1>
<p>Closes / addresses <strong><a href="https://github.com/rtCamp/godam-plugin-wp/issues/23">rtCamp/godam-plugin-wp#23</a></strong> — <em>[Epic] Media Library – Improvement and Bug fixes</em>.</p>
<p>This PR works through all 9 points of the epic (3 bugs, 3 security, 2 performance, 1 correctness-cleanup batch), each in its own commit. Every point below documents <strong>what was happening earlier</strong>, <strong>the fix</strong>, and <strong>how to test</strong>.</p>
<blockquote>
<p><strong>Build note:</strong> JS source changes require a rebuild — run <code>npm run build:js</code> before testing. <code>assets/build/**</code> is gitignored, so the build runs in CI/release.</p>
</blockquote>
<hr>
<h2>1. Folder operations break when the Media Library is opened as a wp.media selector popup</h2>
<p><strong>What was happening earlier</strong>
Folder-tree operations worked on the standalone grid (<code>upload.php</code>) but broke inside a picker modal (Add Media, block image picker, featured image). Root causes: the React tree never re-mounted on reopen (roots were unmounted on close but only injected on <code>initialize()</code>, and reused frames call <code>open()</code>); three redundant close-detectors reset the active folder mid-use; folder clicks drove filtering through grid-only DOM (<code>#media-folder-filter</code> / <code>#post-query-submit</code>) that isn't present in the modal, so clicks silently no-op'd; and a duplicate <code>#rt-transcoder-media-library-root</code> id + a "last visible frame" heuristic targeted the wrong frame.</p>
<p><strong>Fix</strong>
Re-inject and re-mount the tree on frame <code>open</code> (not only <code>initialize</code>); stop resetting UI state on unrelated modal churn (reset only once the last modal closes); make <code>triggerFilterChange</code> set the frame's Backbone <code>media-folder</code> collection prop directly (works without the grid DOM), which the server-side <code>ajax_query_attachments_args</code> filter then applies; and scope each root to its own frame via a <code>_godamFrame</code> handle instead of a shared id.</p>
<p><strong>How to test</strong></p>
<ol>
<li>Edit a post/page → open the block image picker.</li>
<li>Select a folder → the grid filters to that folder's media.</li>
<li>Close and reopen the picker → the folder tree is still present and select/create/rename/delete/drag all work.</li>
<li>Repeat on first and subsequent opens without the active folder resetting.</li>
</ol>
<hr>
<h2>2. Folder delete reports success even when the server rejects it</h2>
<p><strong>What was happening earlier</strong>
<code>DeleteModal.jsx</code> awaited <code>deleteFolderMutation</code> / <code>bulkDeleteFoldersMutation</code> <strong>without <code>.unwrap()</code></strong>. RTK Query <em>resolves</em> (does not throw) on HTTP 4xx/5xx, so the <code>catch</code> never ran — the UI showed "Folder deleted successfully," removed the folder from the tree, and re-filtered, even though the folder still existed server-side.</p>
<p><strong>Fix</strong>
Added <code>.unwrap()</code> to both mutation calls so a server rejection re-throws into the <code>catch</code>; the success snackbar and tree removal now only run on a real success (matching how Rename/Create already worked).</p>
<p><strong>How to test</strong></p>
<ol>
<li>Force a delete to fail (e.g. temporarily return a 500 / a locked folder / insufficient caps).</li>
<li>The UI shows "Failed to delete folder" and the folder <strong>stays</strong> in the tree; a refresh confirms it still exists.</li>
<li>Happy path: a normal delete still succeeds and removes the folder.</li>
</ol>
<hr>
<h2>3. "New Folder" button creates a sub-folder under the last right-clicked folder</h2>
<p><strong>What was happening earlier</strong>
The top-level <strong>New Folder</strong> button opened the creation modal without clearing <code>currentContextMenuFolder</code>. <code>FolderCreationModal</code> derives the new folder's parent from that value, and it was only cleared when clicking All/Uncategorized. So after right-clicking any folder, the top button created the new folder <strong>nested under that folder</strong> instead of at the root.</p>
<p><strong>Fix</strong>
The top-level button now dispatches <code>setCurrentContextMenuFolder( null )</code> before opening the modal, so <code>FolderCreationModal</code> falls into its root-parent branch. The right-click → "New Folder" (sub-folder) path is untouched.</p>
<p><strong>How to test</strong></p>
<ol>
<li>Right-click any folder (e.g. "Folder A"), then dismiss the menu.</li>
<li>Click the top-level <strong>New Folder</strong> button, create a folder → it appears at the <strong>root</strong>, not under A.</li>
<li>Regression: right-click "Folder A" → <strong>New Folder</strong> from the context menu → created <strong>inside</strong> A.</li>
</ol>
<hr>
<h2>4. [Security] "Locked" folders were not enforced on the server</h2>
<p><strong>What was happening earlier</strong>
Folder locking was enforced <strong>only in React</strong>, and <code>window.MediaLibrary.roles</code> is client-controllable. No PHP handler checked the <code>locked</code> term-meta, so a direct REST call could rename/delete/re-parent/populate a locked folder — the disabled UI buttons were the only protection.</p>
<p><strong>Fix</strong>
Added <code>is_folder_locked()</code> — a folder counts as locked if its own <code>locked</code> meta is truthy <strong>or any ancestor's</strong> is (a locked folder protects its whole subtree). Enforced it at every mutation path:</p>
<ul>
<li><strong>Native <code>wp/v2/media-folder</code> endpoints</strong> (rename/delete/re-parent/create-under) — a <code>rest_pre_dispatch</code> guard returns <strong>403</strong> for <code>POST/PUT/PATCH/DELETE</code> on a locked term and for creating/moving into a locked parent. Reads are untouched.</li>
<li><strong><code>assign_images_to_folder</code></strong> — 403 when assigning into (or removing out of) a locked folder.</li>
<li><strong><code>bulk_delete_folders</code></strong> — 403 upfront if any target is locked (never partially deletes).</li>
</ul>
<p><strong>How to test</strong> (as a REST client / browser console with your nonce)</p>
<ol>
<li><code>POST wp/v2/media-folder/{lockedId}</code> with a new name → <strong>403</strong> <code>godam_folder_locked</code>.</li>
<li><code>DELETE wp/v2/media-folder/{lockedId}?force=true</code> → <strong>403</strong>.</li>
<li>Rename/delete a <strong>child</strong> of the locked folder → <strong>403</strong> (cascade).</li>
<li><code>POST wp/v2/media-folder</code> with <code>parent = lockedId</code> → <strong>403</strong>.</li>
<li><code>POST .../assign-folder</code> with <code>folder_term_id = lockedId</code> → <strong>403</strong>.</li>
<li><code>DELETE .../bulk-delete-folders</code> with a locked id in the list → <strong>403</strong> (unlocked ones not deleted).</li>
<li>Regression: the same operations on <strong>unlocked</strong> folders return 200.</li>
</ol>
<hr>
<h2>5. [Security] Folder ZIP export was world-readable at a predictable URL and Contributor-triggerable</h2>
<p><strong>What was happening earlier</strong>
The export wrote to <code>uploads/godam/media-folder-{slug}.zip</code> — a <strong>guessable, unauthenticated</strong> URL in a directory with no listing protection, persisting 24h — and the endpoint was gated only by <code>edit_posts</code>, so Contributors could generate a ZIP of any folder (information disclosure + resource exhaustion).</p>
<p><strong>Fix</strong></p>
<ul>
<li><strong>Unguessable filename</strong>: <code>media-folder-{slug}-{random-20-char-token}.zip</code>.</li>
<li><strong>Directory hardening</strong>: write <code>index.php</code> (blank) and <code>.htaccess</code> (<code>Options -Indexes</code>) into <code>uploads/godam/</code> to prevent enumeration.</li>
<li><strong>Capability</strong>: raised the download-folder endpoint from <code>edit_posts</code> to <code>upload_files</code> (Author+), excluding Contributors.</li>
</ul>
<p><strong>How to test</strong></p>
<ol>
<li>Download a folder → the returned <code>zip_url</code> ends in a 20-char random token.</li>
<li>Fetch the real URL → 200 (<code>application/zip</code>); fetch a guessed <code>media-folder-{slug}.zip</code> → <strong>404</strong>.</li>
<li>Visit <code>/wp-content/uploads/godam/</code> → <strong>403</strong> (no listing).</li>
<li>As a Contributor, <code>POST .../download-folder/{id}</code> → <strong>403</strong>; as admin/author → success.</li>
</ol>
<hr>
<h2>6. [Security] Missing per-object capability check on folder assignment (IDOR)</h2>
<p><strong>What was happening earlier</strong>
<code>assign_images_to_folder</code> was gated only by the broad <code>edit_posts</code> capability and never checked <code>current_user_can( 'edit_post', $attachment_id )</code>, so a lower-privileged user could move <strong>anyone's</strong> attachments between folders. The remove branch (<code>folder_term_id = 0</code>) called <code>remove_all_terms_from_id()</code> on raw IDs without verifying they were attachments, allowing <code>media-folder</code> terms to be stripped from arbitrary objects.</p>
<p><strong>Fix</strong>
Added one upfront guard (covering both the assign and remove paths): every ID must be a real <code>attachment</code> <strong>and</strong> pass <code>current_user_can( 'edit_post', $id )</code>. Any failure rejects the whole request — <strong>400</strong> for a non-attachment, <strong>403</strong> for a permission failure — so there are no partial moves.</p>
<p><strong>How to test</strong></p>
<ol>
<li>Assign a <strong>post/page</strong> id via <code>.../assign-folder</code> → <strong>400</strong> <code>invalid_attachment</code> (both assign and remove paths).</li>
<li>A mixed batch (valid attachment + non-attachment) → <strong>400</strong>, and the valid one is <strong>not</strong> moved.</li>
<li>As an <strong>Author</strong>, try to move an <strong>admin-owned</strong> attachment → <strong>403</strong> <code>rest_forbidden</code>; moving the Author's <strong>own</strong> upload → <strong>200</strong>.</li>
</ol>
<hr>
<h2>7. [Performance] Media Library JS bundles loaded on every admin page (+ video.js code-split)</h2>
<p><strong>What was happening earlier</strong>
Both media-library bundles — <code>media-library.min.js</code> (bundles video.js) and <code>pages/media-library.min.js</code> (React/Redux/RTK/PostHog) — were enqueued via a globally-hooked <code>admin_enqueue_scripts</code>, so they downloaded and parsed on <strong>every</strong> wp-admin screen, not just where the media library / <code>wp.media</code> modal is used.</p>
<p><strong>Fix</strong></p>
<ul>
<li>Added <code>godam_should_load_media_library_assets( $screen )</code> — loads the bundles only on the Media grid, the post/page/attachment editors (which also covers the Elementor and WPBakery editors, both <code>post.php</code>), and GoDAM's own admin pages; extensible via the <code>godam_should_load_media_library_assets</code> filter. Gated both enqueues.</li>
<li><strong>Follow-up (code-split):</strong> converted the static <code>import 'video.js'</code> to a dynamic <code>import()</code> (via a small <code>videojs-loader</code>), so video.js is now a separate on-demand chunk loaded only when a video attachment's detail panel is opened. The always-loaded <code>media-library.min.js</code> dropped from ~803 KB to ~125 KB.</li>
</ul>
<p><strong>How to test</strong></p>
<ol>
<li>DevTools → Network → filter <code>media-library.min.js</code>. Load the Dashboard / Plugins page → <strong>0 hits</strong>. Load <code>upload.php</code> or a post editor → both bundles load.</li>
<li>On <code>upload.php</code>, the base bundle is ~125 KB and <strong>no</strong> video.js chunk loads. Open a virtual/GoDAM-proxy video's detail panel → the video.js chunk (~682 KB) loads on demand and the player plays.</li>
</ol>
<hr>
<h2>8. [Performance] <code>wp_cache_flush()</code> during ZIP build + never-cleaned polling/observer + synchronous ZIP</h2>
<p><strong>What was happening earlier</strong>
(a) The ZIP builder called <code>wp_cache_flush()</code> under memory pressure, flushing the <strong>entire site object cache</strong> to build one download. (b) The React app ran a 500ms <code>setInterval</code> and a <code>document.body</code> subtree <code>MutationObserver</code> for the whole page lifetime (never cleared/disconnected) just to detect modal close — very chatty in the block editor. (c) <code>download_folder</code> built the archive synchronously in-request (batches of 50, files up to 500 MB), risking timeouts on large folders.</p>
<p><strong>Fix</strong></p>
<ul>
<li><strong>(a)</strong> <code>cleanup_memory()</code> now uses <code>wp_cache_flush_runtime()</code> (WP 6.0+, guarded) — frees only the in-memory runtime cache, not the persistent site cache.</li>
<li><strong>(b)</strong> Removed the <code>setInterval</code> and the <code>MutationObserver</code>; rely on the single authoritative <code>wp.media</code> <code>Modal.close()</code> hook (made idempotent).</li>
<li><strong>(c)</strong> ZIP now builds in a <strong>background job</strong>: <code>download_folder</code> registers a job, enqueues it (Action Scheduler when available, else WP-Cron nudged with <code>spawn_cron()</code>), and returns a <code>job_id</code>. A new <code>run_folder_zip_job()</code> builds the archive and stores the result in a transient; a new <code>GET download-folder-status/{job_id}</code> endpoint (gated by <code>upload_files</code> + owner check) reports status. The frontend starts the job, polls every 2s (up to 3 min), then downloads when <code>completed</code> — failures/timeouts surface a snackbar.</li>
</ul>
<p><strong>How to test</strong></p>
<ol>
<li>Download a folder → the request returns immediately with a <code>job_id</code> (status <code>queued</code>); the frontend polls and downloads once ready. Large folders no longer time out the request.</li>
<li>Poll an unknown/expired <code>job_id</code> → <strong>404</strong>; a job started by another user → <strong>403</strong>.</li>
<li>Block editor: with DevTools → Performance, confirm there's no perpetual 500ms task or constant mutation-observer activity from the media library while idle; opening/closing the media modal still resets the sidebar correctly.</li>
</ol>
<blockquote>
<p>Note: the background job runs on the site's cron (Action Scheduler / WP-Cron). On sites with functioning cron (normal traffic or system cron) it processes within seconds; if cron is entirely broken the frontend degrades gracefully with a "timed out preparing the ZIP" error after 3 minutes.</p>
</blockquote>
<hr>
<h2>9. [Bug] Media Library correctness cleanup (batch of low-severity issues)</h2>
<p><strong>What was happening earlier &amp; the fixes</strong></p>
<ul>
<li><strong><code>update_folder_meta_status</code></strong> had identical <code>if/else</code> branches that always incremented <code>updated_count</code> and never populated <code>failed_ids</code>, so lock/bookmark <strong>failures reported as success</strong>. → Now distinguishes a no-op (value already set → success) from a genuine write failure (populates <code>failed_ids</code>/<code>errors</code>).</li>
<li><strong><code>getAllMediaCount</code></strong> omitted <code>X-WP-Nonce</code>, so the "All Media" count ran as an anonymous request. → Added the nonce header.</li>
<li><strong>Drag-move</strong> persisted optimistically with no <code>.unwrap()</code>/error path. → <code>.unwrap()</code> in a try/catch; on failure it shows a snackbar and does not commit the move.</li>
<li><strong>Bookmark/Locked tab copies</strong> went stale on rename/delete. → Synced those copies (and the current selection) in <code>renameFolder</code>, and pruned them in <code>deleteFolder</code>.</li>
<li><strong>Selecting a folder from search</strong> stripped its meta (search requested only <code>id,name</code>), defeating lock checks. → Search now requests <code>id,name,parent,meta</code> and select passes the full folder object.</li>
<li><strong><code>parseInt</code> without radix</strong> on the <code>media-folder</code> URL param, and <strong><code>deleteFolder</code> didn't reset <code>selectedFolder</code></strong>. → Added radix <code>10</code>; reset the selection to "All" when the selected folder (or an ancestor) is deleted.</li>
</ul>
<p><strong>How to test</strong></p>
<ol>
<li>Bulk-lock an already-locked folder → reports <strong>success</strong> (no spurious error); a genuine failure now reports as failed rather than success.</li>
<li>With private/pending media present, the "All Media" count matches the authenticated grid count.</li>
<li>Drag a folder onto a locked destination → move is rejected with a snackbar and the tree snaps back.</li>
<li>Rename a bookmarked/locked folder → the Bookmark/Locked tabs show the new name immediately; delete the selected folder → the grid resets to "All".</li>
<li>Select a folder from search → its lock/bookmark state is respected afterward.</li>
</ol>
<hr>
<h2>Commits (one per point)</h2>

Point | Commit
-- | --
1 | Folder operations breaking in wp.media popup (+ folder-filter gap)
2 | Folder delete reports success even when the server rejects it
3 | New Folder button nesting under the last right-clicked folder
4 | Folder locked state enforced on the server
5 | Media folder ZIP export hardening
6 | IDOR on folder assignment
7 | Media Library JS bundles load on every admin page (+ video.js code-split)
8 | wp_cache_flush() → runtime flush; consolidated close-detectors; async ZIP job
9 | Media Library correctness cleanup batch

</body></html># Media Library — Improvements & Bug Fixes (Epic #23)

Closes / addresses **[[rtCamp/godam-plugin-wp#23](https://github.com/rtCamp/godam-plugin-wp/issues/23)](https://github.com/rtCamp/godam-plugin-wp/issues/23)** — *[Epic] Media Library – Improvement and Bug fixes*.

This PR works through all 9 points of the epic (3 bugs, 3 security, 2 performance, 1 correctness-cleanup batch), each in its own commit. Every point below documents **what was happening earlier**, **the fix**, and **how to test**.

> **Build note:** JS source changes require a rebuild — run `npm run build:js` before testing. `assets/build/**` is gitignored, so the build runs in CI/release.

---

## 1. Folder operations break when the Media Library is opened as a wp.media selector popup

**What was happening earlier**
Folder-tree operations worked on the standalone grid (`upload.php`) but broke inside a picker modal (Add Media, block image picker, featured image). Root causes: the React tree never re-mounted on reopen (roots were unmounted on close but only injected on `initialize()`, and reused frames call `open()`); three redundant close-detectors reset the active folder mid-use; folder clicks drove filtering through grid-only DOM (`#media-folder-filter` / `#post-query-submit`) that isn't present in the modal, so clicks silently no-op'd; and a duplicate `#rt-transcoder-media-library-root` id + a "last visible frame" heuristic targeted the wrong frame.

**Fix**
Re-inject and re-mount the tree on frame `open` (not only `initialize`); stop resetting UI state on unrelated modal churn (reset only once the last modal closes); make `triggerFilterChange` set the frame's Backbone `media-folder` collection prop directly (works without the grid DOM), which the server-side `ajax_query_attachments_args` filter then applies; and scope each root to its own frame via a `_godamFrame` handle instead of a shared id.

**How to test**
1. Edit a post/page → open the block image picker.
2. Select a folder → the grid filters to that folder's media.
3. Close and reopen the picker → the folder tree is still present and select/create/rename/delete/drag all work.
4. Repeat on first and subsequent opens without the active folder resetting.

---

## 2. Folder delete reports success even when the server rejects it

**What was happening earlier**
`DeleteModal.jsx` awaited `deleteFolderMutation` / `bulkDeleteFoldersMutation` **without `.unwrap()`**. RTK Query *resolves* (does not throw) on HTTP 4xx/5xx, so the `catch` never ran — the UI showed "Folder deleted successfully," removed the folder from the tree, and re-filtered, even though the folder still existed server-side.

**Fix**
Added `.unwrap()` to both mutation calls so a server rejection re-throws into the `catch`; the success snackbar and tree removal now only run on a real success (matching how Rename/Create already worked).

**How to test**
1. Force a delete to fail (e.g. temporarily return a 500 / a locked folder / insufficient caps).
2. The UI shows "Failed to delete folder" and the folder **stays** in the tree; a refresh confirms it still exists.
3. Happy path: a normal delete still succeeds and removes the folder.

---

## 3. "New Folder" button creates a sub-folder under the last right-clicked folder

**What was happening earlier**
The top-level **New Folder** button opened the creation modal without clearing `currentContextMenuFolder`. `FolderCreationModal` derives the new folder's parent from that value, and it was only cleared when clicking All/Uncategorized. So after right-clicking any folder, the top button created the new folder **nested under that folder** instead of at the root.

**Fix**
The top-level button now dispatches `setCurrentContextMenuFolder( null )` before opening the modal, so `FolderCreationModal` falls into its root-parent branch. The right-click → "New Folder" (sub-folder) path is untouched.

**How to test**
1. Right-click any folder (e.g. "Folder A"), then dismiss the menu.
2. Click the top-level **New Folder** button, create a folder → it appears at the **root**, not under A.
3. Regression: right-click "Folder A" → **New Folder** from the context menu → created **inside** A.

---

## 4. [Security] "Locked" folders were not enforced on the server

**What was happening earlier**
Folder locking was enforced **only in React**, and `window.MediaLibrary.roles` is client-controllable. No PHP handler checked the `locked` term-meta, so a direct REST call could rename/delete/re-parent/populate a locked folder — the disabled UI buttons were the only protection.

**Fix**
Added `is_folder_locked()` — a folder counts as locked if its own `locked` meta is truthy **or any ancestor's** is (a locked folder protects its whole subtree). Enforced it at every mutation path:
- **Native `wp/v2/media-folder` endpoints** (rename/delete/re-parent/create-under) — a `rest_pre_dispatch` guard returns **403** for `POST/PUT/PATCH/DELETE` on a locked term and for creating/moving into a locked parent. Reads are untouched.
- **`assign_images_to_folder`** — 403 when assigning into (or removing out of) a locked folder.
- **`bulk_delete_folders`** — 403 upfront if any target is locked (never partially deletes).

**How to test** (as a REST client / browser console with your nonce)
1. `POST wp/v2/media-folder/{lockedId}` with a new name → **403** `godam_folder_locked`.
2. `DELETE wp/v2/media-folder/{lockedId}?force=true` → **403**.
3. Rename/delete a **child** of the locked folder → **403** (cascade).
4. `POST wp/v2/media-folder` with `parent = lockedId` → **403**.
5. `POST .../assign-folder` with `folder_term_id = lockedId` → **403**.
6. `DELETE .../bulk-delete-folders` with a locked id in the list → **403** (unlocked ones not deleted).
7. Regression: the same operations on **unlocked** folders return 200.

---

## 5. [Security] Folder ZIP export was world-readable at a predictable URL and Contributor-triggerable

**What was happening earlier**
The export wrote to `uploads/godam/media-folder-{slug}.zip` — a **guessable, unauthenticated** URL in a directory with no listing protection, persisting 24h — and the endpoint was gated only by `edit_posts`, so Contributors could generate a ZIP of any folder (information disclosure + resource exhaustion).

**Fix**
- **Unguessable filename**: `media-folder-{slug}-{random-20-char-token}.zip`.
- **Directory hardening**: write `index.php` (blank) and `.htaccess` (`Options -Indexes`) into `uploads/godam/` to prevent enumeration.
- **Capability**: raised the download-folder endpoint from `edit_posts` to `upload_files` (Author+), excluding Contributors.

**How to test**
1. Download a folder → the returned `zip_url` ends in a 20-char random token.
2. Fetch the real URL → 200 (`application/zip`); fetch a guessed `media-folder-{slug}.zip` → **404**.
3. Visit `/wp-content/uploads/godam/` → **403** (no listing).
4. As a Contributor, `POST .../download-folder/{id}` → **403**; as admin/author → success.

---

## 6. [Security] Missing per-object capability check on folder assignment (IDOR)

**What was happening earlier**
`assign_images_to_folder` was gated only by the broad `edit_posts` capability and never checked `current_user_can( 'edit_post', $attachment_id )`, so a lower-privileged user could move **anyone's** attachments between folders. The remove branch (`folder_term_id = 0`) called `remove_all_terms_from_id()` on raw IDs without verifying they were attachments, allowing `media-folder` terms to be stripped from arbitrary objects.

**Fix**
Added one upfront guard (covering both the assign and remove paths): every ID must be a real `attachment` **and** pass `current_user_can( 'edit_post', $id )`. Any failure rejects the whole request — **400** for a non-attachment, **403** for a permission failure — so there are no partial moves.

**How to test**
1. Assign a **post/page** id via `.../assign-folder` → **400** `invalid_attachment` (both assign and remove paths).
2. A mixed batch (valid attachment + non-attachment) → **400**, and the valid one is **not** moved.
3. As an **Author**, try to move an **admin-owned** attachment → **403** `rest_forbidden`; moving the Author's **own** upload → **200**.

---

## 7. [Performance] Media Library JS bundles loaded on every admin page (+ video.js code-split)

**What was happening earlier**
Both media-library bundles — `media-library.min.js` (bundles video.js) and `pages/media-library.min.js` (React/Redux/RTK/PostHog) — were enqueued via a globally-hooked `admin_enqueue_scripts`, so they downloaded and parsed on **every** wp-admin screen, not just where the media library / `wp.media` modal is used.

**Fix**
- Added `godam_should_load_media_library_assets( $screen )` — loads the bundles only on the Media grid, the post/page/attachment editors (which also covers the Elementor and WPBakery editors, both `post.php`), and GoDAM's own admin pages; extensible via the `godam_should_load_media_library_assets` filter. Gated both enqueues.
- **Follow-up (code-split):** converted the static `import 'video.js'` to a dynamic `import()` (via a small `videojs-loader`), so video.js is now a separate on-demand chunk loaded only when a video attachment's detail panel is opened. The always-loaded `media-library.min.js` dropped from ~803 KB to ~125 KB.

**How to test**
1. DevTools → Network → filter `media-library.min.js`. Load the Dashboard / Plugins page → **0 hits**. Load `upload.php` or a post editor → both bundles load.
2. On `upload.php`, the base bundle is ~125 KB and **no** video.js chunk loads. Open a virtual/GoDAM-proxy video's detail panel → the video.js chunk (~682 KB) loads on demand and the player plays.

---

## 8. [Performance] `wp_cache_flush()` during ZIP build + never-cleaned polling/observer + synchronous ZIP

**What was happening earlier**
(a) The ZIP builder called `wp_cache_flush()` under memory pressure, flushing the **entire site object cache** to build one download. (b) The React app ran a 500ms `setInterval` and a `document.body` subtree `MutationObserver` for the whole page lifetime (never cleared/disconnected) just to detect modal close — very chatty in the block editor. (c) `download_folder` built the archive synchronously in-request (batches of 50, files up to 500 MB), risking timeouts on large folders.

**Fix**
- **(a)** `cleanup_memory()` now uses `wp_cache_flush_runtime()` (WP 6.0+, guarded) — frees only the in-memory runtime cache, not the persistent site cache.
- **(b)** Removed the `setInterval` and the `MutationObserver`; rely on the single authoritative `wp.media` `Modal.close()` hook (made idempotent).
- **(c)** ZIP now builds in a **background job**: `download_folder` registers a job, enqueues it (Action Scheduler when available, else WP-Cron nudged with `spawn_cron()`), and returns a `job_id`. A new `run_folder_zip_job()` builds the archive and stores the result in a transient; a new `GET download-folder-status/{job_id}` endpoint (gated by `upload_files` + owner check) reports status. The frontend starts the job, polls every 2s (up to 3 min), then downloads when `completed` — failures/timeouts surface a snackbar.

**How to test**
1. Download a folder → the request returns immediately with a `job_id` (status `queued`); the frontend polls and downloads once ready. Large folders no longer time out the request.
2. Poll an unknown/expired `job_id` → **404**; a job started by another user → **403**.
3. Block editor: with DevTools → Performance, confirm there's no perpetual 500ms task or constant mutation-observer activity from the media library while idle; opening/closing the media modal still resets the sidebar correctly.

> Note: the background job runs on the site's cron (Action Scheduler / WP-Cron). On sites with functioning cron (normal traffic or system cron) it processes within seconds; if cron is entirely broken the frontend degrades gracefully with a "timed out preparing the ZIP" error after 3 minutes.

---

## 9. [Bug] Media Library correctness cleanup (batch of low-severity issues)

**What was happening earlier & the fixes**
- **`update_folder_meta_status`** had identical `if/else` branches that always incremented `updated_count` and never populated `failed_ids`, so lock/bookmark **failures reported as success**. → Now distinguishes a no-op (value already set → success) from a genuine write failure (populates `failed_ids`/`errors`).
- **`getAllMediaCount`** omitted `X-WP-Nonce`, so the "All Media" count ran as an anonymous request. → Added the nonce header.
- **Drag-move** persisted optimistically with no `.unwrap()`/error path. → `.unwrap()` in a try/catch; on failure it shows a snackbar and does not commit the move.
- **Bookmark/Locked tab copies** went stale on rename/delete. → Synced those copies (and the current selection) in `renameFolder`, and pruned them in `deleteFolder`.
- **Selecting a folder from search** stripped its meta (search requested only `id,name`), defeating lock checks. → Search now requests `id,name,parent,meta` and select passes the full folder object.
- **`parseInt` without radix** on the `media-folder` URL param, and **`deleteFolder` didn't reset `selectedFolder`**. → Added radix `10`; reset the selection to "All" when the selected folder (or an ancestor) is deleted.

**How to test**
1. Bulk-lock an already-locked folder → reports **success** (no spurious error); a genuine failure now reports as failed rather than success.
2. With private/pending media present, the "All Media" count matches the authenticated grid count.
3. Drag a folder onto a locked destination → move is rejected with a snackbar and the tree snaps back.
4. Rename a bookmarked/locked folder → the Bookmark/Locked tabs show the new name immediately; delete the selected folder → the grid resets to "All".
5. Select a folder from search → its lock/bookmark state is respected afterward.

---

## Commits (one per point)

| Point | Commit |
|---|---|
| 1 | Folder operations breaking in wp.media popup (+ folder-filter gap) |
| 2 | Folder delete reports success even when the server rejects it |
| 3 | New Folder button nesting under the last right-clicked folder |
| 4 | Folder locked state enforced on the server |
| 5 | Media folder ZIP export hardening |
| 6 | IDOR on folder assignment |
| 7 | Media Library JS bundles load on every admin page (+ video.js code-split) |
| 8 | `wp_cache_flush()` → runtime flush; consolidated close-detectors; async ZIP job |
| 9 | Media Library correctness cleanup batch |

## Demo

https://github.com/user-attachments/assets/babe2f3e-208a-45dc-b213-fc5ed78a4788

https://github.com/user-attachments/assets/22870b14-7cbd-4a7c-93fb-28d3613019c6

https://github.com/user-attachments/assets/d9eab648-23e1-4765-ad2d-977534bc6f79

https://github.com/user-attachments/assets/e6857519-2bef-418c-a9f4-1b8fecd675e0

Screenshot 2026-08-04 at 3 34 26 PM

### DIFF
--- a/assets/src/js/media-library/index.js
+++ b/assets/src/js/media-library/index.js
@@ -92,6 +92,10 @@ function initializeMediaLibrarySidebar( frame ) {
 				menu.querySelectorAll( '#rt-transcoder-media-library-root' ).forEach( ( el ) => el.remove() );
 				const div = document.createElement( 'div' );
 				div.id = 'rt-transcoder-media-library-root';
+				// Keep a per-frame reference to the wp.media frame on its own root,
+				// so the React sidebar can filter THIS frame's collection directly
+				// (via Backbone props) instead of poking shared grid-only DOM.
+				div._godamFrame = frame;
 				// Append into the menu's first element child (the .media-menu
 				// container). firstChild could be a whitespace text node, which
 				// cannot take children and would throw a HierarchyRequestError.
@@ -137,6 +141,10 @@ function initializeMediaLibrarySidebar( frame ) {
 			} );
 			const div = document.createElement( 'div' );
 			div.id = 'rt-transcoder-media-library-root';
+			// Keep a per-frame reference to the wp.media frame on its own root, so
+			// the React sidebar can filter THIS frame's collection directly (via
+			// Backbone props) instead of poking shared grid-only DOM.
+			div._godamFrame = frame;
 			menu.appendChild( div );
 			// Pass the exact root so the React app renders into THIS frame's sidebar
 			// instead of guessing which frame is active.
--- a/assets/src/js/media-library/index.js
+++ b/assets/src/js/media-library/index.js
@@ -4,13 +4,9 @@
 import { __ } from '@wordpress/i18n';
 
 /**
- * External dependencies
- */
-import videojs from 'video.js';
-
-/**
  * Internal dependencies
  */
+import { getLoadedVideoJs } from './videojs-loader.js';
 import '../../libs/jquery-ui-1.14.2.draggable/jquery-ui';
 import './transcoding-status';
 
@@ -36,6 +32,10 @@ import { isFolderOrgDisabled, isUploadPage, addManageMediaButton } from './utili
  * @param {HTMLElement} container - The DOM element to search for Video.js players.
  */
 function destroyVideoJSPlayersInContainer( container ) {
+	// Only relevant if Video.js was actually loaded (i.e. a player was created this
+	// session). If it was never loaded there are no Video.js players to dispose.
+	const videojs = getLoadedVideoJs();
+
 	if ( ! container || ! videojs ) {
 		return;
 	}
--- a/assets/src/js/media-library/videojs-loader.js
+++ b/assets/src/js/media-library/videojs-loader.js
@@ -1,0 +1,46 @@
+/**
+ * Lazy loader for Video.js in the media library.
+ *
+ * Video.js is a large dependency but is only needed inside the media library when a
+ * (virtual/GoDAM proxy) video attachment's detail panel is opened — which most sessions
+ * never do. Statically importing it pulled the whole library into the always-loaded
+ * media-library bundle. Loading it through a dynamic `import()` splits it into its own
+ * async chunk (webpackChunkName "videojs") that is fetched on demand and then cached, so
+ * the base bundle stays small.
+ */
+
+let cachedVideojs = null;
+let loadingPromise = null;
+
+/**
+ * Dynamically load (and cache) Video.js.
+ *
+ * @return {Promise<Function>} Resolves with the videojs factory function.
+ */
+export function loadVideoJs() {
+	if ( cachedVideojs ) {
+		return Promise.resolve( cachedVideojs );
+	}
+
+	if ( ! loadingPromise ) {
+		loadingPromise = import( /* webpackChunkName: "videojs" */ 'video.js' ).then( ( mod ) => {
+			cachedVideojs = mod.default || mod;
+			return cachedVideojs;
+		} );
+	}
+
+	return loadingPromise;
+}
+
+/**
+ * Synchronously return the already-loaded Video.js, or null if it hasn't been loaded yet.
+ *
+ * Use this where Video.js is only relevant if a player already exists (e.g. cleanup, or
+ * updating a poster on a player that must have been created earlier): if it was never
+ * loaded, there is nothing to act on and the caller can safely skip.
+ *
+ * @return {Function|null} The videojs factory, or null when not yet loaded.
+ */
+export function getLoadedVideoJs() {
+	return cachedVideojs;
+}
--- a/assets/src/js/media-library/views/attachment-detail-two-column.js
+++ b/assets/src/js/media-library/views/attachment-detail-two-column.js
@@ -4,7 +4,6 @@
  * External dependencies
  */
 import DOMPurify from 'isomorphic-dompurify';
-import videojs from 'video.js';
 /**
  * WordPress dependencies
  */
@@ -15,6 +14,7 @@ import { __ } from '@wordpress/i18n';
  */
 import { addIcon, trashIcon, editIcon, barChartIcon } from '../media-library-icons';
 import { canManageAttachment } from '../utility';
+import { loadVideoJs, getLoadedVideoJs } from '../videojs-loader.js';
 
 const AttachmentDetailsTwoColumn = wp?.media?.view?.Attachment?.Details?.TwoColumn;
 
@@ -536,8 +536,10 @@ export default AttachmentDetailsTwoColumn?.extend( {
 		const virtual = this.model.get( 'virtual' );
 
 		// If the attachment is virtual (e.g. a GoDAM proxy video), override default preview.
-		if ( undefined !== virtual && virtual ) {
-			const videoPlayer = videojs( 'videojs-player-' + this.model.get( 'id' ) );
+		// The player was created earlier, so Video.js is already loaded and cached here.
+		const posterVideojs = getLoadedVideoJs();
+		if ( undefined !== virtual && virtual && posterVideojs ) {
+			const videoPlayer = posterVideojs( 'videojs-player-' + this.model.get( 'id' ) );
 			videoPlayer.poster( selected );
 		}
 
@@ -709,8 +711,10 @@ export default AttachmentDetailsTwoColumn?.extend( {
 				const virtual = model.get( 'virtual' );
 
 				// If the attachment is virtual (e.g. a GoDAM proxy video), override default preview.
-				if ( undefined !== virtual && virtual ) {
-					const videoPlayer = videojs( 'videojs-player-' + model.get( 'id' ) );
+				// The player was created earlier, so Video.js is already loaded and cached here.
+				const posterVideojs = getLoadedVideoJs();
+				if ( undefined !== virtual && virtual && posterVideojs ) {
+					const videoPlayer = posterVideojs( 'videojs-player-' + model.get( 'id' ) );
 					videoPlayer.poster( thumbnailURL );
 				}
 
@@ -1002,9 +1006,11 @@ export default AttachmentDetailsTwoColumn?.extend( {
 				` );
 
 				// Wait for DOM to fully render the core preview container.
-				setTimeout( () => {
+				setTimeout( async () => {
 					const videoElement = document.getElementById( videoId );
-					if ( videoElement && typeof videojs !== 'undefined' ) {
+					// Lazy-load Video.js only now that a video player is actually needed.
+					const videojs = videoElement ? await loadVideoJs() : null;
+					if ( videoElement && videojs ) {
 						// Calculate initial dimensions using 16:9 aspect ratio as default
 						const viewContainer = this.$el.closest( '.media-modal-content' ).find( '.attachment-media-view' );
 						const availableWidth = viewContainer.length ? viewContainer.width() : window.innerWidth * 0.65;
--- a/assets/src/js/media-library/views/attachment-detail-two-column.js
+++ b/assets/src/js/media-library/views/attachment-detail-two-column.js
@@ -1008,9 +1008,23 @@ export default AttachmentDetailsTwoColumn?.extend( {
 				// Wait for DOM to fully render the core preview container.
 				setTimeout( async () => {
 					const videoElement = document.getElementById( videoId );
+					if ( ! videoElement ) {
+						return;
+					}
+
 					// Lazy-load Video.js only now that a video player is actually needed.
-					const videojs = videoElement ? await loadVideoJs() : null;
-					if ( videoElement && videojs ) {
+					// Guard the async import: if the chunk fails to load (stale cache after an
+					// update, network hiccup) don't leave an unhandled rejection.
+					let videojs = null;
+					try {
+						videojs = await loadVideoJs();
+					} catch ( e ) {
+						return;
+					}
+
+					// The modal may have closed while the chunk was loading — re-check the
+					// element is still attached before initialising a player on it.
+					if ( videojs && videoElement.isConnected ) {
 						// Calculate initial dimensions using 16:9 aspect ratio as default
 						const viewContainer = this.$el.closest( '.media-modal-content' ).find( '.attachment-media-view' );
 						const availableWidth = viewContainer.length ? viewContainer.width() : window.innerWidth * 0.65;
--- a/inc/classes/class-assets.php
+++ b/inc/classes/class-assets.php
@@ -318,7 +318,14 @@ class Assets {
 		}
 
 		wp_set_script_translations( 'easydam-media-library', 'godam', RTGODAM_PATH . 'languages' );
-		wp_enqueue_script( 'easydam-media-library' );
+
+		// Only load the heavy media-library bundle (~2.65 MB, bundles video.js) where the
+		// media library / wp.media modal is actually used. It was previously enqueued on
+		// every admin screen. Registration + localization above stay unconditional (cheap,
+		// and inert unless the handle is enqueued).
+		if ( godam_should_load_media_library_assets( $screen ) ) {
+			wp_enqueue_script( 'easydam-media-library' );
+		}
 
 		/**
 		 * Dependency library for the date range picker. Its only consumers (the media-library
--- a/inc/classes/class-pages.php
+++ b/inc/classes/class-pages.php
@@ -1227,6 +1227,13 @@ class Pages {
 
 		wp_enqueue_style( 'wp-components' );
 
+		// The media-library React sidebar bundle (~1.1 MB) is the remaining work in this
+		// method; skip it on screens that never open the media library / wp.media modal.
+		// It was previously enqueued on every admin screen.
+		if ( ! godam_should_load_media_library_assets( $screen ) ) {
+			return;
+		}
+
 		// The bundle externalizes several @wordpress packages to wp.* globals
 		// (see the `pages` externals in webpack.config.js). On WP admin / the
 		// block editor these globals are already present, but in other contexts
--- a/inc/classes/class-pages.php
+++ b/inc/classes/class-pages.php
@@ -1243,9 +1243,13 @@ class Pages {
 		// only the handles actually registered in the current context: a handle
 		// missing on older WP / non-standard admin screens would otherwise block
 		// the enqueue entirely.
+		// `media-views` is included so the bundle's modal-close hook (which patches
+		// wp.media.view.Modal in pages/media-library/index.js) has wp.media defined by the
+		// time it runs — otherwise the ordering is incidental and the hook can silently
+		// no-op, leaving React roots unmounted and UI state un-reset.
 		$media_library_deps = array_values(
 			array_filter(
-				array( 'wp-element', 'wp-i18n', 'wp-components', 'wp-api-fetch', 'wp-primitives', 'wp-data', 'wp-notices' ),
+				array( 'wp-element', 'wp-i18n', 'wp-components', 'wp-api-fetch', 'wp-primitives', 'wp-data', 'wp-notices', 'media-views' ),
 				static function ( $handle ) {
 					return wp_script_is( $handle, 'registered' );
 				}
--- a/inc/classes/media-library/class-media-folder-create-zip.php
+++ b/inc/classes/media-library/class-media-folder-create-zip.php
@@ -161,6 +161,9 @@ class Media_Folder_Create_Zip {
 			}
 		}
 
+		// Harden the directory: prevent listing/enumeration of the generated archives.
+		$this->protect_godam_dir( $godam_dir );
+
 		// Full path to the ZIP file.
 		$zip_path = $godam_dir . '/' . $zip_name;
 
@@ -359,8 +362,45 @@ class Media_Folder_Create_Zip {
 	}
 
 	/**
+	 * Harden the uploads/godam directory against enumeration.
+	 *
+	 * The generated ZIPs are static files served from this directory. Writing an
+	 * index.php (blank) and an .htaccess that disables auto-indexing prevents a
+	 * visitor from listing the directory to discover archive filenames. Combined
+	 * with the random token in each ZIP name, direct guessing becomes infeasible.
+	 * Both files are only written when missing, so this is cheap to call repeatedly.
+	 *
+	 * @since 1.3.0
+	 *
+	 * @param string $dir Absolute path to the godam upload directory.
+	 * @return void
+	 */
+	private function protect_godam_dir( $dir ) {
+		global $wp_filesystem;
+
+		if ( empty( $wp_filesystem ) || ! is_a( $wp_filesystem, 'WP_Filesystem_Base' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+			WP_Filesystem();
+		}
+
+		if ( empty( $wp_filesystem ) ) {
+			return;
+		}
+
+		$index = trailingslashit( $dir ) . 'index.php';
+		if ( ! $wp_filesystem->exists( $index ) ) {
+			$wp_filesystem->put_contents( $index, "<?php\n// Silence is golden.\n", FS_CHMOD_FILE );
+		}
+
+		$htaccess = trailingslashit( $dir ) . '.htaccess';
+		if ( ! $wp_filesystem->exists( $htaccess ) ) {
+			$wp_filesystem->put_contents( $htaccess, "Options -Indexes\n", FS_CHMOD_FILE );
+		}
+	}
+
+	/**
 	 * Cleanup the ZIP file after a scheduled time.
-	 * 
+	 *
 	 * @since 1.3.0
 	 *
 	 * @param string $zip_path The path to the ZIP file to be cleaned up.
--- a/inc/classes/media-library/class-media-folder-create-zip.php
+++ b/inc/classes/media-library/class-media-folder-create-zip.php
@@ -356,8 +356,14 @@ class Media_Folder_Create_Zip {
 			gc_collect_cycles();
 		}
 
-		if ( function_exists( 'wp_cache_flush' ) ) {
-			wp_cache_flush();
+		// Free only the in-memory (runtime) object cache that accumulates while iterating
+		// attachments — NOT the entire persistent site cache. The previous wp_cache_flush()
+		// flushed every group in the object cache (hurting every other request on the site)
+		// just to reclaim memory for one download. wp_cache_flush_runtime() (WP 6.0+) drops
+		// the runtime cache without touching the persistent backend; on older WP we rely on
+		// gc_collect_cycles() plus the memory-threshold guard in the batch loop.
+		if ( function_exists( 'wp_cache_flush_runtime' ) ) {
+			wp_cache_flush_runtime();
 		}
 	}
 
--- a/inc/classes/media-library/class-media-folder-create-zip.php
+++ b/inc/classes/media-library/class-media-folder-create-zip.php
@@ -370,11 +370,13 @@ class Media_Folder_Create_Zip {
 	/**
 	 * Harden the uploads/godam directory against enumeration.
 	 *
-	 * The generated ZIPs are static files served from this directory. Writing an
-	 * index.php (blank) and an .htaccess that disables auto-indexing prevents a
-	 * visitor from listing the directory to discover archive filenames. Combined
-	 * with the random token in each ZIP name, direct guessing becomes infeasible.
-	 * Both files are only written when missing, so this is cheap to call repeatedly.
+	 * The generated ZIPs are static files served from this directory. A blank index.php
+	 * stops a visitor from listing the directory to discover archive filenames on both
+	 * Apache and nginx; combined with the random token in each ZIP name, direct guessing
+	 * becomes infeasible. (We deliberately do NOT write an `.htaccess` with
+	 * `Options -Indexes`: it's a no-op on nginx and, on Apache configs where AllowOverride
+	 * excludes Options, it 500s the whole directory — breaking the very downloads it would
+	 * protect. The index.php alone is enough.) The file is only written when missing.
 	 *
 	 * @since 1.3.0
 	 *
@@ -386,21 +388,22 @@ class Media_Folder_Create_Zip {
 
 		if ( empty( $wp_filesystem ) || ! is_a( $wp_filesystem, 'WP_Filesystem_Base' ) ) {
 			require_once ABSPATH . 'wp-admin/includes/file.php';
-			WP_Filesystem();
+
+			// WP_Filesystem() returns false when a non-direct transport needs credentials —
+			// but can still leave a non-connected object in $wp_filesystem, so gate on its
+			// return value rather than on emptiness of the global.
+			if ( ! WP_Filesystem() ) {
+				return;
+			}
 		}
 
-		if ( empty( $wp_filesystem ) ) {
+		if ( empty( $wp_filesystem ) || ! is_a( $wp_filesystem, 'WP_Filesystem_Base' ) ) {
 			return;
 		}
 
 		$index = trailingslashit( $dir ) . 'index.php';
 		if ( ! $wp_filesystem->exists( $index ) ) {
 			$wp_filesystem->put_contents( $index, "<?php\n// Silence is golden.\n", FS_CHMOD_FILE );
-		}
-
-		$htaccess = trailingslashit( $dir ) . '.htaccess';
-		if ( ! $wp_filesystem->exists( $htaccess ) ) {
-			$wp_filesystem->put_contents( $htaccess, "Options -Indexes\n", FS_CHMOD_FILE );
 		}
 	}
 
--- a/inc/classes/rest-api/class-media-library.php
+++ b/inc/classes/rest-api/class-media-library.php
@@ -79,8 +79,10 @@ class Media_Library extends Base {
 
 		$term_id = isset( $matches[1] ) ? (int) $matches[1] : 0;
 
-		// Renaming / re-parenting / deleting a locked folder (or a child of one).
-		if ( $term_id > 0 && $this->is_folder_locked( $term_id ) ) {
+		// Renaming / re-parenting / deleting a locked folder (or a child of one). A
+		// bookmark toggle is a personal flag rather than a modification of the folder's
+		// protected content, so it is still allowed on a locked folder.
+		if ( $term_id > 0 && $this->is_folder_locked( $term_id ) && ! $this->is_bookmark_only_folder_update( $request, $term_id ) ) {
 			return new \WP_Error(
 				'godam_folder_locked',
 				__( 'This folder is locked and cannot be modified.', 'godam' ),
@@ -102,6 +104,67 @@ class Media_Library extends Base {
 		}
 
 		return $result;
+	}
+
+	/**
+	 * Whether a native media-folder update request only toggles the `bookmark` meta and
+	 * changes nothing else (name, slug, description, parent, or the `locked` meta itself).
+	 *
+	 * Bookmarking is a per-user convenience flag, not a modification of the folder's
+	 * protected content, so it stays allowed on a locked folder — but this must NOT become
+	 * a hole through which a locked folder is renamed, re-parented, or unlocked.
+	 *
+	 * @param \WP_REST_Request $request The request being dispatched.
+	 * @param int              $term_id The media-folder term id.
+	 * @return bool True if the request only changes the bookmark meta.
+	 */
+	private function is_bookmark_only_folder_update( $request, $term_id ) {
+		if ( 'DELETE' === $request->get_method() ) {
+			return false;
+		}
+
+		$term = get_term( $term_id, 'media-folder' );
+
+		if ( ! $term || is_wp_error( $term ) ) {
+			return false;
+		}
+
+		// Any change to the folder's own fields disqualifies it.
+		$name = $request->get_param( 'name' );
+		if ( ! is_null( $name ) && (string) $name !== (string) $term->name ) {
+			return false;
+		}
+
+		$slug = $request->get_param( 'slug' );
+		if ( ! is_null( $slug ) && (string) $slug !== (string) $term->slug ) {
+			return false;
+		}
+
+		$description = $request->get_param( 'description' );
+		if ( ! is_null( $description ) && (string) $description !== (string) $term->description ) {
+			return false;
+		}
+
+		$parent = $request->get_param( 'parent' );
+		if ( ! is_null( $parent ) && (int) $parent !== (int) $term->parent ) {
+			return false;
+		}
+
+		// The lock state itself must not change through this path (unlocking is a
+		// capability-gated operation handled by the bulk-lock endpoint).
+		$meta = $request->get_param( 'meta' );
+		if ( is_array( $meta ) && array_key_exists( 'locked', $meta ) ) {
+			$current_locked = get_term_meta( $term_id, 'locked', true );
+			$current_bool   = ( '1' === (string) $current_locked || 1 === $current_locked || true === $current_locked || 'true' === $current_locked );
+			$requested      = $meta['locked'];
+			$requested_bool = ( true === $requested || 1 === $requested || '1' === (string) $requested || 'true' === $requested );
+
+			if ( $current_bool !== $requested_bool ) {
+				return false;
+			}
+		}
+
+		return true;
 	}
 
 	/**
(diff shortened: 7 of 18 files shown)


KMchaudhary:
## Code review — Media Library improvements & bug fixes (Epic #23)

Reviewed the full diff (17 files, `fix/godam-media-libary` → `fix/godam-media-library-issues`) against the 9 points in [rtCamp/godam-plugin-wp#23](https://github.com/rtCamp/godam-plugin-wp/issues/23). The direction is right on every point — below is what I think needs another pass before merge, ordered by severity.

---

### 🔴 Blocking

#### 1. `godam_should_load_media_library_assets()` never matches the GoDAM dashboard — wrong screen id

`inc/helpers/custom-functions.php:1519-1522`

```php
$godam_pages = array(
    'toplevel_page_godam',
    'godam_page_rtgodam_media_editor',
);
```

The menu slug is `rtgodam` (`class-pages.php:26`), so the dashboard screen id is **`toplevel_page_rtgodam`** (`class-pages.php:86`, `$menu_page_id`). `toplevel_page_godam` matches nothing, so with point 7 in place both media-library bundles now silently stop loading on the GoDAM dashboard. (The same typo already exists in `godam_should_load_auth_detector_script()` at `:1477` — it was harmless-ish there, it isn't here.)

#### 2. Point 7 breaks the Tools page: `window.easydamMediaLibrary` disappears

`inc/classes/class-assets.php:322-328`

The comment says registration + localization "stay unconditional (cheap, and inert unless the handle is enqueued)". `wp_localize_script()` data is only *printed* for handles that end up in `WP_Scripts::$to_do` — i.e. only when the handle is enqueued (or pulled in as a dependency). So gating the enqueue also drops the `easydamMediaLibrary` global everywhere else, and it has out-of-bundle consumers:

- `pages/tools/components/tabs/RetranscodeTab.jsx:58` — `nonce === window.easydamMediaLibrary?.godamToolsNonce`. The Tools screen (`godam_page_rtgodam_tools`) is not allowlisted, so the "Retranscode selected" hand-off from the media library will now always fail with *"The requested operation is not allowed. The nonce provided in the URL is invalid or expired."*
- `pages/video-editor/components/SidebarLayers.js:128` is fine (media-editor page *is* allowlisted).

Either localize onto a handle that always loads, or add `godam_page_rtgodam_tools` to the allowlist.

#### 3. Screen allowlist misses several screens that do open `wp.media`

`inc/helpers/custom-functions.php:1507-1526`

`Pages::admin_enqueue_scripts()` calls `wp_enqueue_media()` for **all** GoDAM pages (`class-pages.php:544-555`), and these open the modal:

- `pages/godam/components/tabs/GeneralSettings/BrandImageSelector.jsx`, `.../VideoSettings/VideoWatermark.jsx` → `godam_page_rtgodam_settings`
- `pages/analytics/index.js`, `pages/analytics/Analytics.js` → `godam_page_rtgodam_analytics`

On those screens the modal now renders without GoDAM's frame overrides — no GoDAM cloud tab, no virtual-media handling, no folder sidebar. Same gap for core screens: `widgets.php`, `site-editor` (FSE), `customize`, `term.php`.

#### 4. Lock enforcement gap: an attachment can still be moved *out* of a locked folder

`inc/classes/rest-api/class-media-library.php:768-775`

The assign branch only checks the **destination**:

```php
if ( $this->is_folder_locked( $folder_term_id ) ) { … 403 … }
…
wp_set_object_terms( $attachment_id, $folder_term_id, 'media-folder' );
```

`wp_set_object_terms()` *replaces* the folder assignment, so `POST /assign-folder` with `folder_term_id = <any unlocked folder>` empties a locked folder one item at a time. The remove branch (`folder_term_id = 0`) correctly checks the source at `:729-743`; the assign branch needs the same source check.

#### 5. The new `rest_pre_dispatch` guard breaks bookmarking a locked folder

`inc/classes/rest-api/class-media-library.php:341-347` + `pages/media-library/components/context-menu/ContextMenu.jsx:284`

Single-folder bookmark toggling goes through `updateFolderMutation` → `POST /wp/v2/media-folder/{id}`, which the guard now rejects with 403 for any locked folder. The Bookmark button is **not** disabled for locked folders (`ContextMenu.jsx:443`, `disabled={ isSpecialFolder }`), so it just fails with *"Failed to update bookmark status"*. Inconsistently, the *multi-select* bookmark path uses `godam/v1/.../bulk-bookmark-folders`, which is unguarded and still succeeds.

Decide whether bookmark counts as "modification"; if not, exempt meta-only bookmark writes (or route the single case through the bulk endpoint too).

#### 6. `pollZipJobStatus()` can never observe a failed poll

`pages/media-library/redux/api/folders.js:32-43`

```js
const json = await res.json();
const data = json?.data || {};
if ( data.status === 'completed' || data.status === 'failed' ) { return data; }
```

`res.ok` is never checked, and a `WP_Error` serialises as `{code, message, data: {status: 404}}` — so `data.status` is the **HTTP code**, not the job status. The PR description's test case ("poll an unknown/expired `job_id` → 404") therefore doesn't surface: the client keeps polling for the full 180 s (90 requests) and then reports the generic "Timed out while preparing the ZIP file." Same for the 403 owner-check path.

#### 7. Point 2 isn't fixed for the bulk-delete case

`pages/media-library/components/modal/DeleteModal.jsx:62` + `class-media-library.php:1508-1518`

`bulk_delete_folders()` returns **HTTP 200 with `success: true`** on partial failure. `.unwrap()` only throws on a rejected `baseQuery`, so the multi-select path still hits the success branch, shows *"Folders deleted successfully"*, and drops every selected folder from the tree even when some deletions failed — which is the exact bug the PR is fixing for the single-folder case. Either return a non-2xx for partial failure or inspect `response.errors` client-side.

---

### 🟠 Should fix

#### 8. ZIP export: still exports other users' media, and archives are now unbounded

`class-media-library.php:1298-1324`, `:1381`

- Raising the cap to `upload_files` doesn't address the information-disclosure half of issue #5: an Author can still ZIP a folder full of media they can't edit. The job runs in cron as user 0, so nothing filters by capability at build time either.
- The random-token filename removes the previous *overwrite* behaviour. Every download now writes a **new** full-folder archive retained for 24 h, with no dedupe and no rate limit — N clicks = N copies of the folder on disk, plus N cron jobs each re-reading every file. That's a bigger resource-exhaustion surface than the synchronous version it replaced. Consider reusing an existing fresh archive per folder, or capping concurrent/pending jobs per user.

#### 9. "New Folder" stays disabled after right-clicking a locked folder

`pages/media-library/App.js:177`

```jsx
disabled={ selectedFolder?.meta?.locked || currentFolder?.meta?.locked || contextSelectedFolder?.meta?.locked }
```

Now that the click handler clears `currentContextMenuFolder` and always creates at the root, `contextSelectedFolder?.meta?.locked` (and `selectedFolder?.meta?.locked`) are stale gating — the button is greyed out for a root-level create that the lock has no bearing on. This is the same state the fix at `:174` exists to neutralise.

#### 10. Removing both fallbacks makes modal-close detection order-dependent and unrecoverable

`pages/media-library/index.js:53`, `:92-94`

`setupMediaModalCloseDetection()` runs at module evaluation and returns permanently if `wp.media.view.Modal` isn't defined yet — there's no retry and no `DOMContentLoaded` re-attempt. `media-library-react` doesn't declare `media-views`/`media-editor` as a dependency, so the ordering is incidental (it currently works because `Assets::admin_enqueue_scripts` runs `wp_enqueue_media()` before `Pages` enqueues). If that ordering ever shifts, React roots are never unmounted and `resetUIState()` never fires, with nothing to catch it. Cheapest fix: add `media-views` to `$media_library_deps` in `class-pages.php:1248`, or retry on `DOMContentLoaded`.

#### 11. Lazy video.js: unhandled rejection + possibly-detached element

`assets/src/js/media-library/views/attachment-detail-two-column.js:1009-1012`

```js
setTimeout( async () => {
    const videoElement = document.getElementById( videoId );
    const videojs = videoElement ? await loadVideoJs() : null;
```

No `try/catch`: if the `videojs` chunk 404s (stale cache after an update, CDN hiccup) the rejection is unhandled and the player silently never renders — where the old static import couldn't fail. Also `videoElement` is captured *before* the await; by the time the chunk resolves the modal may have closed, so it's worth re-checking `videoElement.isConnected` before `videojs( videoElement, … )`.

#### 12. N+1 queries in the new IDOR guard

`class-media-library.php:711-743`

`get_post_type()` per id, then `wp_get_object_terms()` **per id** in the remove branch, then `is_folder_locked()` per folder. A 100-attachment drag-move goes from ~2 queries to a few hundred. `wp_get_object_terms()` accepts an array of object ids, and `_prime_post_caches( $attachment_ids )` collapses the type checks into one query.

---

### 🟡 Nits / worth a look

#### 13. `godam_should_load_media_library_assets()` is a verbatim copy of `godam_should_load_auth_detector_script()`

`inc/helpers/custom-functions.php:1502` vs `:1457` — identical `$media_screens`, identical `$godam_pages` (including the bad `toplevel_page_godam`). Two lists that must stay in sync will drift. Extract the screen list once and have both call it.

#### 14. The new async chunk skips the repo's own multi-compilation workaround

`webpack.config.js:100-105`

The `mediaLibrary` compilation inherits `sharedConfig.output` — default `uniqueName`, `chunkFilename: '[name].min.js'` — and now emits `assets/build/js/videojs.min.js` into the same directory and onto the same `webpackChunkgodam` loading global as `godamPlayerFrontend`, which already emits `videojs-quality-levels` / `videojs-contrib-ads` / `fontawesome-*` async chunks. The comment at `:117-131` on `godamImageLayersFrontend` documents that this exact setup produced `TypeError: o[t] is not a function`; worth giving `mediaLibrary` its own `uniqueName` + prefixed `chunkFilename` for the same reason. Also, `[name].min.js` has no content hash, so a browser can serve a stale `videojs.min.js` after a plugin update.

#### 15. Lock enforcement is anchored to a route string rather than the taxonomy layer

`class-media-library.php:329` — `preg_match( '#^/wp/v2/media-folder(?:/(\d+))?$#', … )`

Everything that isn't that exact REST route bypasses the lock: `wp_update_term()` / `wp_delete_term()` from WP-CLI, other plugins, or future internal code. The issue's suggestion (`pre_delete_term` + a term-update guard) protects the data rather than one entry point, with the REST filter reduced to producing a nicer 403. Lower risk today because the taxonomy is `show_ui => false`, but it's a fragile anchor.

#### 16. Minor

- `class-media-library.php:778` — the assign loop still re-checks `get_post_type( $attachment_id ) !== 'attachment'`; the new upfront guard at `:711` already rejected those, so it's dead code now.
- `class-media-folder-create-zip.php:401-404` — `Options -Indexes` is a no-op on nginx, and on Apache configs where `AllowOverride` excludes `Options` it makes the whole `uploads/godam/` directory return **500**, breaking the very downloads it protects. The blank `index.php` alone already stops listing on both servers.
- `class-media-folder-create-zip.php:384-394` — `protect_godam_dir()` reaches for `WP_Filesystem()` while the rest of the class writes with `ZipArchive` / `wp_mkdir_p()` directly. If `get_filesystem_method()` isn't `direct`, `WP_Filesystem()` returns false but leaves a *non-connected* transport object in `$wp_filesystem`, which `if ( empty( $wp_filesystem ) )` won't catch.
- `redux/slice/folders.js:165` — `if ( window.godam )` guards the write, whereas `changeSelectedFolder` at `:61` does `window.godam = window.godam || {}`. If `window.godam` doesn't exist yet, the reset to "All" isn't mirrored out.
- `redux/slice/folders.js:99-128` — `renameFolder` syncs `folders` / `bookmarks` / `lockedFolders` / `selectedFolder`, but not `state.currentContextMenuFolder.name`, which is what the rename modal itself renders.

---

### What looks good

- `.unwrap()` on delete + drag-move, the `parseInt` radix, the `X-WP-Nonce` on `getAllMediaCount`, and the `_fields: 'id,name,parent,meta'` widening in search are all clean, targeted fixes.
- `update_folder_meta_status()` — reading the current value first to distinguish a no-op from a real write failure is the right call, and `(string) $current === (string) $value` handles the `false → ''` case correctly.
- `wp_cache_flush()` → `wp_cache_flush_runtime()` with the `function_exists` guard: correct and well explained.
- Dropping the `document.body` subtree `MutationObserver` + 500 ms `setInterval` is a real win in the block editor.
- `is_folder_locked()` walking ancestors so a lock protects the whole subtree is the right semantics, and rejecting the whole `bulk_delete_folders` request up front (never partially deleting) is the right shape.

---

*Review focus: correctness, security, and regressions from the enqueue gating and the sync → async ZIP change. Not verified at runtime — `assets/build/**` is gitignored, so the code-split chunk and the enqueue gating were reasoned about from source only.*

rudrakshi-gupta:
Thanks for the thorough review, @KMchaudhary. I've resolved the feedback and self-tested once.
 Summary per point:

## Blocking
1. Wrong dashboard screen id — `toplevel_page_godam` -> `toplevel_page_rtgodam`. Consolidated both screen checks into a single `godam_is_media_library_screen()` so the lists can't drift (also covers #13). Verified the bundle now loads on the GoDAM dashboard.
2. `window.easydamMediaLibrary` disappearing — the allowlist now includes every GoDAM admin page (dashboard, tools, settings, analytics, media-editor, help, what's-new), so the global + `godamToolsNonce` are present on the Tools screen again. Verified on Dashboard + Tools.
3. Allowlist missed wp.media screens — added all GoDAM pages + `widgets` + `site-editor`.
4. Move-out-of-locked gap — the assign path now checks the source too: 403 if an attachment currently lives in a locked folder other than the destination. Verified 403.
5. Bookmark broke on a locked folder — added `is_bookmark_only_folder_update()`; a bookmark-only meta write is allowed on a locked folder while rename / re-parent / unlock stay 403. Verified: bookmark 200, rename 403, unlock 403.
6. `pollZipJobStatus` couldn't observe a failure — now checks `res.ok` (a WP_Error serialises `data.status` as the HTTP code), so a 404/403 fails fast instead of polling the full 180s. Verified the status endpoint returns 404 for an unknown job.
7. Bulk-delete partial failure — `DeleteModal` now inspects the 200 body's `errors[]` and surfaces a failure instead of "deleted successfully". Verified the endpoint returns { success:true, errors:[...], deleted_count } on partial.

## Should fix
9. New Folder stayed disabled — removed the stale lock-based `disabled` (the top button always creates at root now). Verified the button is enabled.
10. Modal-close order dependency — added `media-views` to the React bundle deps so `wp.media` is defined when the close hook patches it. Verified the hook is installed.
11. Lazy video.js — wrapped `loadVideoJs()` in try/catch and re-check `videoElement.isConnected` before initialising.
12. N+1 in the IDOR guard — prime `_prime_post_caches()` + `update_object_term_cache()` once.

## Nits
- 13 — deduped the two screen helpers into `godam_is_media_library_screen()`.
- 14 — gave the `mediaLibrary` compilation its own `uniqueName` + prefixed, content-hashed `chunkFilename` so its async chunks can't collide with the player build's.
- 16a — removed the now-dead `get_post_type` recheck in the assign loop.
- 16b — dropped the `.htaccess` (Options -Indexes is a no-op on nginx and 500s on restrictive Apache); the blank `index.php` alone stops listing on both.
- 16c — gate on `WP_Filesystem()`'s return value rather than emptiness of the global.
- 16d — window.godam = window.godam || {} before mirroring the reset.
- 16e — `renameFolder` now also syncs `currentContextMenuFolder.name`.

## Deferred to a follow-up
- #8 (ZIP still exports media the user can't edit + unbounded per-download archives) — needs capability-filtering at build time + per-user job dedupe/rate-limiting; better scoped as its own PR.
- #15 (anchor lock enforcement at the taxonomy layer via pre_delete_term / term-update guards) — defense-in-depth; the REST route is the only real entry point today since the taxonomy is show_ui => false.

Thanks again — the source-check and bookmark-exemption catches were spot on.

KMchaudhary (approved):

Issue #465, opened 2025-06-19
Text Domain loading issue


subodhr258:
### PHP Notice: 
Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>godam</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /var/www/htdocs/current/wp-includes/functions.php on line 6121

This notice occurs even though we are not using any `load_plugin_textdomain()` function for the 'godam' text domain.
It could be that the issue is due to `wp_set_script_translations()` functions.

To Reproduce:
Ensure WordPress debugging is enabled (e.g., `define('WP_DEBUG', true);` in wp-config.php).

By moving our translation loading function later, we should resolve the "Translation loading ... triggered too early" notice.

mi5t4n:
@subodhr258 Let me take this one, it should be an easy fix.

mi5t4n:
@subodhr258 Were you able to reproduce this issue consistently? I've been trying to replicate it, but it seems to happen randomly.

subodhr258:
Okay @mi5t4n , thanks!

justlevine:
This is not fixed, even with all other plugins disabled and TT5 active it's everywhere.

<img width="1854" height="988" alt="Image" src="https://github.com/user-attachments/assets/843a41d5-59bf-4d46-9c20-e5d13c612563" />

subodhr258:
The previous open PR partially solved the issue.

Four call sites in GoDAM, all reachable on a default install:

| Call site | Ran on | Fix |
| :--- | :--- | :--- |
| WPForms video field, translates labels in the constructor | `wpforms_loaded` | moved to `init` |
| Fluent Forms recorder field, same | `fluentform/loaded` | moved to `init` |
| `Addon_Registry::warn_incompatible_addons()` | `plugins_loaded` (15) | moved to `admin_init` |
| `Addon_Registry::boot_addons()`, both notice branches | `plugins_loaded` (15) | notice text built by a callback at render time |

Plus one in the add-on: `Godam_For_Woo_Addon::get_dependencies()`, which GoDAM calls during `plugins_loaded`. `GODAM_WOO_MIN_GODAM_VERSION` must be updated to this fixed version with its release.

[...]

PR #2061, opened 2026-08-05
Fix translation issue during plugins_loaded

elifvish:
This pull request addresses several issues related to translation timing and admin notices in WordPress 6.7+, improves translation coverage, and introduces new GitHub Actions workflows for CI. The main focus is to prevent warnings about premature translation loading, enhance the translation workflow, and ensure admin notices and columns are registered only when appropriate. It also updates the plugin version and changelog.

**Translation timing and admin notice improvements:**

* Refactored add-on dependency checks and admin notice registration to defer translation until after the text domain is loaded, preventing `_load_textdomain_just_in_time was called incorrectly` warnings on WordPress 6.7+ (`class-abstract-addon.php`, `class-addon-registry.php`). Admin notices are now registered on `admin_init` and accept a callback for translated content (`Fbe047b1L21R21`, `[[1]](diffhunk://#diff-fd532c481766a2d90bbfa7827391bf0c8c4216f664de57934a4ad6257d5eacd8R89-R94)`, `[[2]](diffhunk://#diff-fd532c481766a2d90bbfa7827391bf0c8c4216f664de57934a4ad6257d5eacd8L217-R246)`, `[[3]](diffhunk://#diff-fd532c481766a2d90bbfa7827391bf0c8c4216f664de57934a4ad6257d5eacd8L284-R324)`, `[[4]](diffhunk://#diff-bd6a11d6674add225c2a77e203febc6903a5d5506dc536ee0c1e8cd9dd157586L72-R82)`, `[[5]](diffhunk://#diff-bd6a11d6674add225c2a77e203febc6903a5d5506dc536ee0c1e8cd9dd157586R91-R102)`, `[[6]](diffhunk://#diff-bd6a11d6674add225c2a77e203febc6903a5d5506dc536ee0c1e8cd9dd157586R112-R135)`).

* Changed the Fluent Forms integration to load its field on `init` instead of `fluentform/loaded`, ensuring translations are available before use (`class-init.php`).

**Admin UI and API key validation:**

* Updated the logic for registering the "Transcode Status" column in the Media Library: column and sortability registration now only occur if the API key is valid, and only on the relevant admin screen, avoiding unnecessary API calls and warnings (`godam-transcoder-functions.php`). [[1]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901R285-R289) [[2]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901R298-R301) [[3]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901L374-R421) [[4]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901L393-R436)

**Translation and build process improvements:**

* Enhanced the `composer pot` script to run in two passes, ensuring both built and source references are included in the POT file for better JS translation support. Added a script description for clarity (`composer.json`).

* Added a new GitHub Actions workflow to check that the committed POT file is up to date with the build, and another to run PHP unit tests on pull requests (`pot_up_to_date_on_pull_request.yml`, `php_unit_tests_on_pull_request.yml`). [[1]](diffhunk://#diff-1bb68a707724fd01c140fb328ed4fcc55462cfd1041b3456ebac5ed3c9399debR1-R54) [[2]](diffhunk://#diff-b0ac40beef58645b7be7361c18f6789692ca4a0b69e6736fd7509dffd9e18f1dR1-R39)

**Changelog and version updates:**

* Updated plugin version to 2.1.1 and documented fixes and tweaks in `CHANGELOG.md`, `godam.php`, and `README.md`. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR3-R8) [[2]](diffhunk://#diff-11ad0cbfc6e1732b964b5298a09e2b1cd2df3efc7f60babe88d5d0e88d748916L6-R6) [[3]](diffhunk://#diff-11ad0cbfc6e1732b964b5298a09e2b1cd2df3efc7f60babe88d5d0e88d748916L46-R46) [[4]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L13-R13)

**Distribution and test configuration:**

* Updated `.distignore` to exclude the `/tests/` directory and clarified comments about test exclusions.
* Minor CI workflow step simplification in JS unit tests workflow.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -32,11 +32,14 @@ bin/
 .packagejsonignore
 .distignore
 
-# JS unit tests + their config — never ship in the distribution build.
+# Tests + their config — never ship in the distribution build.
 # (Tests under assets/src/js are already covered by that dir's exclusion above;
-# these patterns also catch co-located tests under pages/.)
+# the *.test.js patterns also catch co-located tests under pages/.)
 jest.config.js
 *.test.js
 *.spec.js
 __tests__/
 phpunit.xml.dist
+# Leading slash anchors this to the plugin root: an unanchored `tests/` would
+# also strip vendor/**/tests/, which is not this file's call to make.
+/tests/
--- a/.github/workflows/js_unit_tests_on_pull_request.yml
+++ b/.github/workflows/js_unit_tests_on_pull_request.yml
@@ -15,8 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Set up Node
         uses: actions/setup-node@v4
--- a/.github/workflows/php_unit_tests_on_pull_request.yml
+++ b/.github/workflows/php_unit_tests_on_pull_request.yml
@@ -1,0 +1,41 @@
+on:
+  pull_request:
+    branches:
+      - develop
+      - main
+
+name: PHP Unit Tests
+
+permissions:
+  contents: read
+
+jobs:
+  runPHPUnitTests:
+    name: Run PHP unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      # 8.4, not the plugin's minimum of 7.4: composer.lock pins dev tools that
+      # require PHP ^8.4 (doctrine/instantiator, symfony/finder), so
+      # `composer install` fails its platform check on anything older. This
+      # constrains the test toolchain only — runtime PHP support is unaffected.
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.4'
+          coverage: none
+          tools: composer
+
+      # Dev dependencies are required here: the deploy and plugin-check
+      # workflows install with --no-dev, so phpunit is absent in those.
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-progress
+
+      # Pure unit tests via tests/bootstrap.php stubs — no WordPress install and
+      # no database, so no service containers are needed. Includes the #465
+      # hook-timing guards, which only prevent a regression if CI runs them.
+      - name: Run PHP unit tests
+        run: composer phpunit
--- a/.github/workflows/php_unit_tests_on_pull_request.yml
+++ b/.github/workflows/php_unit_tests_on_pull_request.yml
@@ -15,8 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
 
       # 8.4, not the plugin's minimum of 7.4: composer.lock pins dev tools that
       # require PHP ^8.4 (doctrine/instantiator, symfony/finder), so
--- a/.github/workflows/pot_up_to_date_on_pull_request.yml
+++ b/.github/workflows/pot_up_to_date_on_pull_request.yml
@@ -1,0 +1,54 @@
+on:
+  pull_request:
+    branches:
+      - develop
+      - main
+
+name: Translation Template
+
+permissions:
+  contents: read
+
+jobs:
+  checkPotIsUpToDate:
+    name: Check the committed POT matches a fresh build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.4'
+          coverage: none
+          tools: composer
+
+      - name: Install dependencies
+        run: |
+          composer install --prefer-dist --no-progress
+          npm ci
+
+      # assets/build is gitignored, so the POT cannot be regenerated correctly
+      # without building first: `composer pot` reads the built files to record the
+      # reference WordPress hashes when it looks up JS translations at runtime.
+      - name: Build assets
+        run: npm run build:prod
+
+      - name: Regenerate the POT
+        run: composer pot
+
+      # POT-Creation-Date is stamped on every run, so it is the one line allowed
+      # to differ. Anything else means the committed POT is stale — most likely
+      # regenerated without a build, which silently drops every built-file
+      # reference and leaves JS strings untranslatable.
+      - name: Fail if the committed POT is stale
+        run: |
+          if ! git diff -I 'POT-Creation-Date' --exit-code -- languages/; then
+            echo "::error::languages/ is out of date. Run 'npm run build:prod && composer pot' and commit the result."
+            exit 1
+          fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,10 @@
 # Changelog #
 
-## v2.1.1 (August 5, 2026) ##
+## v2.1.1 (August 6, 2026) ##
 
-- Fix: Removed every GoDAM translation call that ran before the `init` action, so WordPress 6.7+ no longer reports `Function _load_textdomain_just_in_time was called incorrectly` for the `godam` text domain. Covers the WPForms and Fluent Forms field registrations, the add-on compatibility and dependency notices, and the media library transcode status column.
-- Tweak: The media library transcode status column no longer re-verifies the API key over HTTP on front-end, REST and cron requests.
-- Fix: The translation template is now built from source instead of compiled assets, adding 47 strings that could not previously be translated and dropping stale ones that no longer exist in the plugin.
-- Chore: PHP unit tests now run on every pull request.
+- Fix: GoDAM no longer fills the debug log with `Function _load_textdomain_just_in_time was called incorrectly` on WordPress 6.7 and later. On sites that display errors, the warning could also appear on the page itself, in the admin and on the front end.
+- Fix: 34 more pieces of on-screen text can now be translated, covering the video player's transcript, chapters, comments, sharing and sign-in prompts, which no language pack could reach before.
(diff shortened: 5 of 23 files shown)


KMchaudhary:
## Code review — rtCamp/godam#2061 + rtCamp/godam-for-woo#190

Reviewed both PRs together, since the add-on contract change spans them. The core fix (deferring translation past `init`) is right and the deferred-callback design is clean. Two things below would regress users if merged as-is, and a handful of smaller items follow.

---

### 🔴 Blockers

**1. The POT change breaks JS translation lookups — `composer.json` (both repos)**

Dropping `assets/src` from the `make-pot` excludes and adding `assets/build` moves every JS string's source reference from the *shipped* file to a path `.distignore` strips from the release.

Measured on the two POTs in this PR:
- **1313 of 1316** entries lost their `assets/build/**.js` reference (woo: `406 → 0`)
- **376** entries now have their only `.js` reference under `assets/src/{blocks,css,js}` — all three of which `.distignore` removes from the distribution

That matters because WordPress resolves JS translations by path hash:

```
languages/godam-<locale>-md5( 'assets/build/js/media-library.min.js' ).json
```

…while `wp i18n make-json` / Loco Translate derive the filename from the POT reference, which is now `md5( 'assets/src/js/media-library/views/attachment-details.js' )`. The two never match, so every string behind the 17 `wp_set_script_translations()` calls (video editor, analytics, dashboard, onboarding, media library, recorder — plus woo's 5) stays untranslated. The shipped POT also references files that don't exist in the release ZIP.

The goal is right — 49 genuinely new strings show up, and the 3 dropped ones are stale build artifacts. Suggested fix: **keep `assets/build` in the scan** (exclude only `node_modules`, `vendor`, `tests`) so each entry carries both a source and a built reference, or generate a `--use-map` src→build mapping.

**2. `GODAM_WOO_MIN_GODAM_VERSION` bump to `2.1.1` disables the whole add-on on GoDAM 2.1.0 — `godam-for-woo.php:46`**

Plugin updates are independent, so a site on GoDAM 2.1.0 that updates GoDAM for Woo first hits `is_godam_version_compatible() === false` → `boot_addons()` shows the "requires GoDAM 2.1.1" notice and `continue`s → `boot()` never runs → `inc/class-bootstrap.php` is never loaded and **every Shoppable Video block, product hotspot and reel stops rendering on the front end**.

And the bump isn't needed. With the new `get_missing_dependency_messages()` override (`inc/class-godam-for-woo-addon.php:132`), the add-on already resolves the closure itself, so GoDAM 2.1.0's `show_admin_notice( $message ) → wp_kses_post( $message )` receives a plain string and works fine. As written, the constant trades a cosmetic `_load_textdomain_just_in_time` notice for a total loss of add-on functionality. Recommend leaving it at `2.1.0`.

---

### 🟠 Correctness

**3. `init` drops the "WPForms actually booted" guarantee — `inc/classes/wpforms/class-wpforms-integration.php:55`**

`class-wpforms-field-godam-video.php` declares the class only inside `if ( class_exists( 'WPForms_Field' ) )`, and `init_godam_video_field()` does an unguarded `new …WPForms_Field_GoDAM_Video()`. On `wpforms_loaded` that was safe by construction — WPForms fired the hook, so its includes had run. On `init` the only gate left is `is_plugin_active()` at line 32, which is true whenever the plugin file is in `active_plugins` even if WPForms aborted its own bootstrap (requirements check, license bail, fatal-recovery mode). In that state the class is never declared and `init` fatals with `Class "RTGODAM\Inc\WPForms\WPForms_Field_GoDAM_Video" not found` on **every** request, front end included.

Please add a `class_exists()` guard in `init_godam_video_field()`, and the equivalent for `Recorder_Field` / `BaseFieldManager` at `inc/classes/fluentforms/class-init.php:93`.

**4. `$dep['check']` still read without `isset()` — `class-abstract-addon.php:95` and `:114`**

The same loop gained `$dep['message'] ?? ''`, but `check` stays unguarded. A dependency entry with only `name` + `message` (easy to assume now that `message` is optional) emits `Warning: Undefined array key "check"` from every `plugins_loaded`, twice per add-on per request — and with `WP_DEBUG_DISPLAY` on it prints into the page, i.e. the exact symptom class this PR removes. Same gap in the woo copy at `inc/class-godam-for-woo-addon.php:136`.

**5. Sortable column still registered unconditionally at file load — `admin/godam-transcoder-functions.php:419`**

`manage_upload_sortable_columns` is left at file scope, so it runs on front-end / REST / cron requests too — undercutting the hunk's stated goal — and registers `convert_status` as sortable even when `rtgodam_register_media_status_columns()` bailed on an invalid API key and the column doesn't exist. (Separately: nothing in the repo handles a `convert_status` orderby, so the header is inert either way.) Moving this `add_filter()` inside the new function alongside the other two would make the hunk self-consistent.

---

### 🟡 Design / depth

**6. `admin_init` is broader than this column needs — `admin/godam-transcoder-functions.php:402`**

`rtgodam_is_api_key_valid()` → `rtgodam_get_user_data()` re-verifies over HTTP once the cache is an hour old. On `admin_init` that fires from `post.php`, `options.php`, `admin-post.php`, every settings save — none of which render `manage_media_columns`. `load-upload.php` fires after `admin_init` but before `upload.php` constructs `WP_Media_List_Table`, so it still registers in time while confining the remote call to the one screen that needs it.

**7. Dependency checks now run twice per admin request — `class-addon-registry.php:242`**

`dependencies_met()` runs `get_dependencies()` + all checks, then the deferred closure runs `get_missing_dependency_messages()` which does it all again (and reallocates the closures). Free for woo's `class_exists()`, but the documented contract on `'check'` promises nothing about purity — an add-on checking an option, transient or remote endpoint pays twice. Sharing one resolved result between the two would avoid it.

**8. CI is pinned to PHP 8.4, so the declared minimum is never tested — `php_unit_tests_on_pull_request.yml:28`**

The comment is accurate (`doctrine/instantiator 2.1.0` needs `^8.4`, `symfony/finder v8.0.3` needs `>=8.4`), but the root cause is `composer.lock` having `"platform": {}` — it was resolved on an 8.4 machine. The result is that no job ever exercises the plugin on `Requires PHP: 7.4`–8.3, so 8.0+-only syntax (nullsafe, named args, `str_contains`) would ship green and fatal on a 7.4 host. Adding `"config": { "platform": { "php": "7.4" } }` and re-locking would let this run a real version matrix instead.

**9. `get_missing_dependency_messages()` is a verbatim copy across two repos — `inc/class-godam-for-woo-addon.php:133`**

Defensible as a back-compat shim, but worth marking as one with an explicit removal condition (e.g. "delete once `GODAM_WOO_MIN_GODAM_VERSION >= 2.1.1`") rather than a permanent override — the docblock at line 122 explicitly declines to version-gate it, so nothing signals when it can go.

---

### 🟡 Tests

**10. The #465 guards assert on source text, not behaviour — `tests/php/EarlyTranslationHooksTest.php`**

They're pinned tight in one direction and loose in the other:
- `test_wpforms_field_is_registered_on_init()` fails on **any** `add_action( 'wpforms_loaded', … )` in that file, so an unrelated future WPForms hook breaks CI for no reason (same for `fluentform/loaded`).
- Adding a fresh `__( 'x', 'godam' )` inside `boot_addons()` or at the top of `godam-transcoder-functions.php` reintroduces the exact bug while all five assertions still pass — `test_addon_boot_notices_are_deferred()` only checks the `show_admin_notice( callable $x )` signature.

A behavioural guard (attach a `gettext` filter, fire `plugins_loaded` against the bootstrap stubs, assert nothing translated) would cover the class rather than five instances. `AddonDependencyMessagesTest` is good as-is — the "string that matches a function name" case is a nice catch.

**11. The woo override has no tests and no PHP test workflow — godam-for-woo**

godam-for-woo has no `tests/` dir, no `phpunit.xml.dist`, no `phpunit` script, and only `build_artifact_on_push.yml` + `js_unit_tests_on_pull_request.yml`. The copy that actually prevents the wp-admin fatal is the untested one. If someone later simplifies its guard to a plain `is_callable()`, the `'phpversion'`-style case `AddonDependencyMessagesTest` covers here would slip through there.

**12. Tests run on the head SHA, not the merge result — `php_unit_tests_on_pull_request.yml:19`**

`ref: ${{ github.event.pull_request.head.sha }}` opts out of checkout's default `refs/pull/N/merge`, so a base change that breaks this branch only surfaces after merge. Mirrors `js_unit_tests_on_pull_request.yml`, so if that's deliberate it'd be worth a comment in both.

---

### 🔵 Worth confirming

**13. Is the "WooCommerce inactive" path actually reachable? — `inc/class-godam-for-woo-addon.php:88`**

`godam-for-woo.php:16` declares `Requires Plugins: godam, woocommerce` and the header requires WP 6.9 — well past 6.5, where core added plugin dependencies and started blocking activation on unmet ones. If that holds, `dependencies_met()` returning `false` is close to unreachable in supported configurations, and the whole callable-message contract (core API change + woo override + version bump) is defending a path users can't hit. Worth confirming before locking `'message' => callable` into the public `Abstract_Addon` API — `class-abstract-addon.php:75` now makes it a requirement for every future add-on.

---

### 👍 Nice

- Correctly spotted that `warn_incompatible_addons()` on `admin_init` is both later *and* admin-only, so the old `is_admin()` guard becomes redundant rather than being dropped silently.
- The empty-message early return in the `admin_notices` closure quietly handles the "dependency became satisfied between `plugins_loaded` and render" case (e.g. activating WooCommerce) — no stale notice.
- Moving the media-column registration off file scope incidentally fixes `rtgodam_get_user_data()` persisting a `currentUserId` resolved before the current user was available.
- `/tests/` in `.distignore` with the leading-slash anchor is right for the rsync-based 10up deploy action, and the comment explains why.

KMchaudhary (dismissed):

KMchaudhary (approved):

PR #2063, opened 2026-08-06
Fix built-file references and guard form-field hooks in init

elifvish:
This pull request focuses on improving the plugin’s reliability and translation coverage, especially for WordPress 6.7+, and introduces a new workflow to ensure translation templates remain up-to-date. It fixes issues with translation initialization, avoids unnecessary remote API calls, and hardens add-on dependency handling to prevent warnings. The changes also update the build and test workflows and improve test coverage.

**Translation and Internationalization Improvements:**

* The translation template (`godam.pot`) is now built from both source and compiled assets, ensuring 34 additional UI strings (e.g., video player transcript, chapters, comments, sharing, and sign-in prompts) are translatable, and dropping obsolete ones. A new workflow (`pot_up_to_date_on_pull_request.yml`) automatically checks that the committed POT file is current on every pull request. (`[[1]](diffhunk://#diff-1bb68a707724fd01c140fb328ed4fcc55462cfd1041b3456ebac5ed3c9399debR1-R54)`, `[[2]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L24-R30)`, `[[3]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL3-R7)`, `[[4]](diffhunk://#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L227-R231)`)
* Translation calls are now guaranteed not to run before the `init` action, preventing debug log warnings like `Function _load_textdomain_just_in_time was called incorrectly` on WordPress 6.7+ in both admin and frontend contexts. (`[[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL3-R7)`, `[[2]](diffhunk://#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L227-R231)`)

**Media Library Status Column and Remote Calls:**

* The transcode status column in the media library now checks API key validity only when building the media list table, avoiding unnecessary remote calls on front-end, REST, or cron requests. The sortable status column is registered only if the API key is valid. (`[[1]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901R285-R289)`, `[[2]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901R298-R301)`, `[[3]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901L384-R399)`, `[[4]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901L394-R421)`, `[[5]](diffhunk://#diff-881e50398e30fb2291e73ff71173e13e2c57d2d141bfbd76f9971eb94aa27901L415-R436)`)

**Add-on Dependency Handling:**

* Add-on dependency checks are hardened: missing or non-callable `'check'` keys are treated as satisfied dependencies, preventing `Undefined array key "check"` warnings and possible front-end errors under `WP_DEBUG_DISPLAY`. Tests are added to cover these cases. (`[[1]](diffhunk://#diff-bd6a11d6674add225c2a77e203febc6903a5d5506dc536ee0c1e8cd9dd157586R91-R102)`, `[[2]](diffhunk://#diff-bd6a11d6674add225c2a77e203febc6903a5d5506dc536ee0c1e8cd9dd157586L114-R123)`, `[[3]](diffhunk://#diff-800c4bddf39e44e03d5187926c09b594d5ff6e43d0b1481ce8063ea1ea14dcf3R170-R213)`)

**Plugin Integration Guards:**

* Guards are added to ensure that integration code for WPForms and FluentForms only runs if the required parent classes are loaded, preventing fatal errors if those plugins are deactivated or fail to initialize properly. (`[[1]](diffhunk://#diff-c2fccfa5cfb0493e85a2f48f7c5330dea73542d042b3edd374146f44c65a6887R91-R104)`, `[[2]](diffhunk://#diff-e7eeaa700bc30f8ef6e3bd83ea8647705e5f96ba3b6dd7f6d5916eb1827b9470R155-R172)`)

**Build, Test, and Workflow Updates:**

* The JS and PHP unit test workflows are simplified by removing explicit `ref` usage in `actions/checkout`. The translation template workflow is added, and the PHPCS configuration now excludes test stubs for translation sniffing. (`[[1]](diffhunk://#diff-f8a39e5f8e23e1c77e608b6c7e7bb1cde14482c454d296252ac946a5e17bd573L18-L19)`, `[[2]](diffhunk://#diff-b0ac40beef58645b7be7361c18f6789692ca4a0b69e6736fd7509dffd9e18f1dL18-L19)`, `[[3]](diffhunk://#diff-daa856500eea114bbf034fd35b7600050d965e5d19bb0700885c075747848fd8R30-R35)`)

These changes make the plugin more robust, improve translation coverage, and ensure that automated checks keep the codebase and translation files aligned.


KMchaudhary (approved):

CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
...
 
-## v2.1.1 (August 5, 2026) ##
+## v2.1.1 (August 6, 2026) ##
 
 - Fix: Removed every GoDAM translation call that ran before the `init` action, so WordPress 6.7+ no longer reports `Function _load_textdomain_just_in_time was called incorrectly` for the `godam` text domain. Covers the WPForms and Fluent Forms field registrations, the add-on compatibility and dependency notices, and the media library transcode status column.
-- Tweak: The media library transcode status column no longer re-verifies the API key over HTTP on front-end, REST and cron requests.
-- Fix: The translation template is now built from source instead of compiled assets, adding 47 strings that could not previously be translated and dropping stale ones that no longer exist in the plugin.
+- Tweak: The media library transcode status column no longer re-verifies the API key over HTTP except on the media library screen itself.
+- Fix: The translation template now covers both the plugin source and the built assets it ships, adding 47 strings that could not previously be translated and dropping stale ones that no longer exist in the plugin.
 - Chore: PHP unit tests now run on every pull request.
```
subodhr258: Do you think we should have this PHP unit test - related line in the changelog for our users?

Other 2 in changelog also need to be more user-oriented

subodhr258:
Reviewed. The three PHP changes hold up under checking:

- Verified `get_columns()` runs before `get_sortable_columns()` in `WP_List_Table::get_column_info()`, and `rtgodam_get_user_data()` refreshes `timestamp` before `update_option()`, so the two `rtgodam_is_api_key_valid()` calls cost one remote verify per hour, not two.
- Verified the two-pass pot: 1354 of 2064 entries now carry an `assets/build/…` reference (base had 0), and the build paths are hash-free, which is what `make-json` and the runtime `md5(relative_src)` lookup need.

Four things worth fixing before merge:

1. **PR description is stale.** It says the `pot` script now calls `bin/make-pot.sh`. That file does not exist on this branch; `composer.json` uses an inline two-command array.

2. **The changelog's "+47 strings" is not what this PR does.** I diffed the msgid sets: base and head are identical at 2064 entries. This pot diff is 100% reference churn, zero strings added or removed (the +47 landed earlier on develop). As written, a reviewer will hunt 18k lines for 47 new strings and find none. Suggest rewording to something like "now also records the built-asset reference for each string, which is what WordPress hashes to find JS translations at runtime".

3. **Pass 2 should keep excluding `assets/build`.** Pass 1's `--include="assets/build"` already covers everything under build, min or not, so dropping the exclude from pass 2 buys no reference and costs duplicates: 17 entries now carry both `assets/build/**/render.php` and `assets/src/**/render.php`, and 40 carry doubled `block.json` refs. Restoring `--exclude="assets/build,assets/node_modules,node_modules,tests,vendor"` on pass 2 loses nothing.

4. **The `head.sha` workflow change probably belongs in its own PR.** The comment is honest that it gives up merge-result testing, but "consistency with the JS workflow" is an argument for fixing that workflow rather than matching it. `refs/pull/N/merge` is the default so base-branch drift surfaces at PR time instead of at release.

Smaller notes:

- The docblock on `rtgodam_register_media_status_columns()` says each of the three hooks "defers to `rtgodam_is_api_key_valid()` there". `rtgodam_add_status_columns_content()` does not (and does not need to, since the action only fires for a column `get_columns()` returned), so the comment is inaccurate as written.
- The new guard tests assert source shape, not behavior: `/function\s+init_godam_video_field\(\s*\)\s*\{\s*if\s*\(\s*!\s*class_exists\(/` breaks on a comment or blank line before the guard, or on extracting it into a helper, even when behavior is unchanged. A behavioral test (don't declare `WPForms_Field`, call the method, assert no fatal) would survive refactors.
- `${TMPDIR:-/tmp}` is POSIX-shell only, so `composer pot` breaks on Windows, and the temp pot is never cleaned up.
- `assets/build/` is gitignored, so the committed pot encodes line numbers into artifacts nothing verifies. A CI job would close that: `npm ci && npm run build:prod && composer pot && git diff --exit-code languages/`.
- The 23 entries with a src-JS ref and no build ref are all from `assets/src/blocks/godam-player/track-uploader.js`, which nothing in the repo imports. Dead code, not a gap in this fix, but worth deleting separately.

PR #2072, opened 2026-08-10
feat(ci): audit rtgodam_before/after_attachment_lookup hook integrity

Pathan-Amaankhan:
## Summary

Adds automated CI + pre-commit checks that guard the `rtgodam_before_attachment_lookup` / `rtgodam_after_attachment_lookup` hook pair against silent regressions. These hooks let a plugin that centralizes attachments across a multisite network (e.g. wp-dam) switch site context around every place GoDAM reads/writes attachment data directly, then switch back. GoDAM only fires them — nothing previously stopped a hook from quietly disappearing, or new attachment-access code from shipping without the wrapping it needs.

## What's included

- `bin/hook-check/balance.php` — one check: every `rtgodam_before_attachment_lookup` in a scope must reach a matching `rtgodam_after_attachment_lookup` before the scope ends, and no `after` may fire with nothing open. Branch-aware around early `return`/`throw`/`exit`/`die`/`wp_die()`-style exits, so a guard clause that closes the wrap and returns early isn't a false alarm. Deliberately doesn't track *whether* a hook was removed or *how much* attachment-access code exists — that's a GitHub reviewer's job reading the diff, and a removed hook whose access call stays is caught directly (not historically) by the sibling coverage check below.
- `bin/hook-check/coverage.php` — a finer-grained, per-call-site pass: flags every attachment-access call (`get_post_meta`, `wp_get_attachment_url`, etc.) or attachment-shaped query pattern (`WP_Query`, `get_posts`, `get_children`, raw `$wpdb` queries) that has no bracket open at that exact point, unless its argument is provably a pass-through of an already-safe value. Includes interprocedural call-graph tracing (arbitrary hops, with codebase-wide-uniqueness safeguards) so a parameter-sourced call covered only by a caller's caller doesn't false-positive.
- `bin/hook-check/shared.php` — tokenizer/scope-walking primitives shared by both scripts, plus the `godam-coverage-*` directive parser (below). Extracted so the two scripts can't drift to different definitions of "what counts as firing a hook."
- **Inline `godam-coverage-ignore` / `-disable` / `-enable` / `-ignore-file` comments** — the only acceptance mechanism, replacing an earlier baseline-JSON + `update-baseline` design entirely. Mirrors `phpcs:ignore`/`disable`/`enable`/`ignoreFile` minus a rule-code argument, since each checker only ever checks one thing:
  ```php
  get_post_meta( $id, 'x', true ); // godam-coverage-ignore -- not attachment data, see docblock above

  // godam-coverage-disable -- whole migration is host-post-scoped, see class docblock
  ...several lines...
  // godam-coverage-enable
  ```
  A `-disable` with no matching `-enable` before end of file is a hard error, not a silent "rest of the file is fine." No baseline files, no `update-baseline` mode — every run judges only the code and its own inline comments as they currently stand.
- `.github/workflows/wp_dam_hook_integrity_on_pull_request.yml` — runs both checks in CI, scoped to PRs touching `inc/`, `admin/`, `assets/src/blocks/`, or the checker scripts themselves.
- `bin/hook-check/pre-commit.sh` (wired into `.husky/pre-commit`) — runs the same two checks locally pre-commit, over the same path scope.
- `package.json` — `godam-hooks:balance-check` / `godam-hooks:coverage-check`, combined into `lint:godam-hooks` via `npm-run-all --continue-on-error` so a failure in one doesn't hide the other's findings.

## Notable detection fixes along the way

- A raw `$wpdb` query referencing `{$wpdb->posts}`/`{$wpdb->postmeta}` via string interpolation previously went undetected — PHP's tokenizer splits that into separate `$wpdb`/`->`/`posts` tokens with no `wp_posts` substring for the old check to find.
- A `WP_Query`/`get_posts`/`get_children` call whose arguments come from a separate helper function (e.g. `new WP_Query( build_args() )`) previously read as "not attachment-shaped" and wasn't flagged at all, since the post_type literal lives inside a callee the scan never looks into. Now defaults to flagging it, matching this tool's existing philosophy elsewhere: a false positive costs a one-line dismissal comment, a false negative is a silent gap.

## How it works

Single mode — the code plus its own inline `godam-coverage-*` comments are the complete, self-contained source of truth:

```
php bin/hook-check/balance.php
php bin/hook-check/coverage.php
```

A real finding means either add the hook pair, or — if the call genuinely isn't centralization-relevant attachment data — add the matching inline comment with a reason, right at the call site.


package.json
```diff
@@ -164,6 +164,7 @@
 		"lint:php": "vendor/bin/phpcs",
 		"lint:php:fix": "vendor/bin/phpcbf",
 		"lint:package-json": "wp-scripts lint-pkg-json --ignorePath .packagejsonignore",
+		"lint:godam-hooks": "php bin/godam-wp-dam-hook-check.php check && php bin/godam-interprocedural-leak-check.php check && php bin/godam-attachment-access-coverage-check.php check",
```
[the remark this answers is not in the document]
Pathan-Amaankhan: Resolved in https://github.com/rtCamp/godam/pull/2072/commits/2f96043650fe55071f4e4885a58b9dd0dcab3f5c

KMchaudhary:
@Pathan-Amaankhan 
- This CI will cause pre-commit failures for non-attachment post code.

And will you be able to segregate whether the hooks are called in the context of `attachment` or `post` on VideoJet's feature plugins?

Also, what if the developer tries to perform some post operation on a non-attachment post type where `rtgodam_before_attachment_lookup` & `rtgodam_after_attachment_lookup` hooks are not required, in that case, Husky will show the pre-commit error.

bin/hook-check/shared.php
```diff
@@ -0,0 +1,1516 @@
...
+				break;
+
+			case 'enable':
+				if ( null !== $open_disable ) {
+					$disabled_ranges[] = array(
+						'start'  => $open_disable['line'],
+						'end'    => $line,
+						'reason' => $open_disable['reason'],
+					);
+					$open_disable = null;
```
rtBot: :warning: **Warning**: Equals sign not aligned with surrounding assignments; expected 6 spaces but found 1 space (*Generic.Formatting.MultipleStatementAlignment.NotSameWarning*).

bin/hook-check/shared.php
```diff
@@ -0,0 +1,1516 @@
...
+
+	return $uses;
+}
+
+/**
+ * Function names whose typical call shape passes a callback (directly, or
+ * nested in an args array) that WordPress stores and invokes later,
+ * potentially on a different request than the one that registered it.
+ *
+ * register_activation_hook()/register_deactivation_hook() take __FILE__
```
rtBot: :no_entry_sign: **Error**: Doc comment long description must start with a capital letter (*Generic.Commenting.DocComment.LongNotCapital*).

rtBot (dismissed):

**phpcs** scanning turned up:

:no_entry_sign: 1 error 

:warning: 1 warning 


***


.github/workflows/wp_dam_hook_integrity_on_pull_request.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  pull_request:
+    paths:
+      - 'inc/**/*.php'
+      - 'admin/**/*.php'
+      - 'assets/src/blocks/**/*.php'
```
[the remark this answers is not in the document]
Pathan-Amaankhan: Updated in https://github.com/rtCamp/godam/pull/2072/commits/2fbeebca634fea67676c13cfd9eebb5c1d457009

.github/workflows/wp_dam_hook_integrity_on_pull_request.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Set up PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.4'
+        coverage: none
+    - name: Run hook integrity check
+      run: php bin/hook-check/balance.php check
+    - name: Run attachment-access coverage check
+      if: always()
+      run: php bin/hook-check/coverage.php check
```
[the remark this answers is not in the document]
Pathan-Amaankhan: hooks will be added via https://github.com/rtCamp/godam/pull/2080

KMchaudhary (approved):

bin/hook-check/shared.php
Pathan-Amaankhan: Resolved

bin/hook-check/shared.php
Pathan-Amaankhan: Resolved

PR #2075, opened 2026-08-13
fix(transcoder-admin): correct spelling of 'Trialling' to 'Trialing'

KMchaudhary:
Ref: https://rtcamp.slack.com/archives/C25QZJ50B/p1786602456414059?thread_ts=1786515787.338739&cid=C25QZJ50B

### DIFF
--- a/admin/class-rtgodam-transcoder-admin.php
+++ b/admin/class-rtgodam-transcoder-admin.php
@@ -101,7 +101,7 @@ class RTGODAM_Transcoder_Admin {
 		$grace_period_days   = 30;
 
 		// If the user is in trial mode, show trial expiry notice.
-		if ( 'Trialling' === $subscription_status && $trial_end ) {
+		if ( 'Trialing' === $subscription_status && $trial_end ) {
 			$days_until_trial_end = ceil( ( $trial_end - $current_time ) / DAY_IN_SECONDS );
 
 			if ( $days_until_trial_end > 0 ) {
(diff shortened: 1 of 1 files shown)


mayan-000 (approved):

elifvish (changes requested):
Requesting changes — I think this needs to accept both spellings rather than swap one for the other.

`subscription_status` isn't internal copy; it arrives verbatim from the central API and is cached in the `rtgodam-usage` option (written at `admin/class-rtgodam-transcoder-handler.php:591`, read at `admin/class-rtgodam-transcoder-admin.php:97`). Two gaps remain with the rename alone, even given that the API is fixed and backward compatible:

1. **Stale caches.** `rtgodam-usage` only refreshes on API-key re-save or a transcode callback. `rtgodam_reconcile_api_key_state()` (`inc/classes/class-update.php:108`) runs on version bump but early-returns for any site with a persisted key status, and never touches `rtgodam-usage` in either branch; the hourly `rtgodam_get_user_data()` refresh writes a different option (`rtgodam_user_data`). So a site that is trialing at upgrade time keeps `Trialling` cached, and the trial-expiry notice silently stops rendering for exactly the users it exists for.

2. **Version cutover.** The API's backward compatibility branches on the `plugin_version` we send (`admin/godam-transcoder-functions.php:500`). If godam-core's threshold lands above the release this ships in (currently `2.1.1`), trialing sites receive `Trialling` against code expecting `Trialing` — the same silent bug, just relocated.

One line covers both, with no migration and no extra HTTP call on upgrade:

```php
if ( in_array( $subscription_status, array( 'Trialing', 'Trialling' ), true ) && $trial_end ) {
```

elifvish (approved):

PR #2078, opened 2026-08-13
fix(analytics): show only the hotspot label in the layer rail

subodhr258:
Connected ticket - https://github.com/rtCamp/godam-analytics/issues/196

### What

In the Video Layer Timeline detail panel, every hotspot row in the "Hotspots in this layer" rail repeated the parent layer name and then truncated the part that identifies the hotspot. Every row read "Hotspot layer at 4.59s ..." so you could not tell the hotspots apart.

### Why it happened

The player tracker writes each hotspot event's `layer_name` as the parent layer name, then a separator dash, then "Hotspot N". The rail already shows the parent layer as its card title, so `groupRows` was rendering that whole string as the row name, and the rail's `truncate` style then clipped off the "Hotspot N".

Woo product rows were already fine, because they carry a bare `product_name` ("Lamp", "Sofa").

### The fix

In `groupRows` (`pages/analytics/hooks/useVideoLayerData.js`), strip the parent-name prefix from a hotspot sub-row's name so each row reads just "Hotspot N". It only runs for rows with no `product_name` (so Woo is untouched) and only when the name starts with the exact “<parent> — ” prefix the tracker composes (matching the full separator, not just the parent name, so a custom label that merely begins with the parent name or one that itself starts with a dash is left intact). It reads what is already stored, so it fixes existing and historical data too, with no change to the wire format.

### Verification

- Unit tests: 15/15 pass, including 5 new ones (the strip, the untouched Woo case, a custom name that does not start with the parent prefix, a label that only shares the parent name as a prefix, and a label that itself begins with a dash). The last two fail on the pre-review greedy logic and pass with the exact-separator match.
- On `godam-dev.local` pointed at the develop analytics service, video "People": the 4.59s hotspot rail now reads "Hotspot 1 / Hotspot 2", and the Woo layer still reads "Lamp / Sofa".

### Automation impact

Rail row labels changed from the prefixed form ("<parent layer name>" then "Hotspot N") to just "Hotspot N". Any test asserting the old, prefixed label needs updating.

### Known follow-up (not this PR)

Hotspots are numbered by position ("Hotspot N"), so when one is removed or reordered the numbers can collide (you can see two "Hotspot 2" rows, one active and one "Removed"). Giving each hotspot a stable identity label is a separate change.

### Before
Screenshot 2026-08-13 at 5 08 21 PM

### After
image

### DIFF
--- a/pages/analytics/hooks/useVideoLayerData.js
+++ b/pages/analytics/hooks/useVideoLayerData.js
@@ -432,7 +432,24 @@ export function groupRows( rows, layerType, configIndex ) {
 				// path the parent uses so a raw composite UUID never reaches
 				// the rail. Falls back to a generic ordinal "<TypeLabel> #N"
 				// when nothing usable.
-				const rawSubName = md.product_name || row.layer_name || '';
+				let rawSubName = md.product_name || row.layer_name || '';
+				// The hotspot tracker packs the parent layer name, a separator
+				// dash, and "Hotspot N" into layer_name. The rail's card title
+				// already shows the parent, so that prefix is pure repetition
+				// and, once the row is truncated, it hides the "Hotspot N" that
+				// actually names the row. Drop the parent prefix (and the
+				// separator after it), keeping only the hotspot's own label.
+				// Woo rows carry a bare product_name and never match this, so
+				// they are left alone.
+				if (
+					! md.product_name &&
+					md.parent_layer_name &&
+					rawSubName.startsWith( md.parent_layer_name )
+				) {
+					rawSubName = rawSubName
+						.slice( md.parent_layer_name.length )
+						.replace( /^[\s\p{Pd}]+/u, '' );
+				}
 				let subName;
 				if (
 					rawSubName &&
--- a/pages/analytics/hooks/useVideoLayerData.js
+++ b/pages/analytics/hooks/useVideoLayerData.js
@@ -143,6 +143,17 @@ const TRACKER_AUTO_NAME_RE = new RegExp(
 );
 
 /**
+ * Exact separator the hotspot tracker puts between the parent layer name
+ * and the hotspot's own label when it composes `layer_name`. It builds
+ * `<parentLayerName> — <subLabel>` (space, em dash, space) in
+ * `hotspotLayerManager.js`; keep this in sync with that string. Matching
+ * the full separator — rather than just the parent name — is what lets us
+ * strip the prefix without touching a hotspot label that legitimately
+ * begins with a dash.
+ */
+const HOTSPOT_NAME_SEPARATOR = ' — ';
+
+/**
  * Whether a server-delivered layer_name is missing, a bare UUID, or a
  * tracker-side auto-generated name — i.e. anything we should replace with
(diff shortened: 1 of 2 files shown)


pages/analytics/hooks/useVideoLayerData.js
```diff
@@ -432,7 +432,24 @@ export function groupRows( rows, layerType, configIndex ) {
...
+				// they are left alone.
+				if (
+					! md.product_name &&
+					md.parent_layer_name &&
+					rawSubName.startsWith( md.parent_layer_name )
+				) {
+					rawSubName = rawSubName
+						.slice( md.parent_layer_name.length )
+						.replace( /^[\s\p{Pd}]+/u, '' );
+				}
```
KMchaudhary, line 467: The separator cleanup is greedy across dashes, so it can also eat a dash that belongs to the hotspot's own label.

`/^[\s\p{Pd}]+/u` consumes *every* leading whitespace/dash character, not just the one separator. A hotspot named `-50% off` inside a layer named `Summer sale` is tracked as `Summer sale — -50% off` (`hotspotLayerManager.js` composes `${parentLayerName} — ${subLabel}`). After `slice()` the remainder is `" — -50% off"`, and the character class then swallows space + em dash + space **and** the label's own hyphen, so the rail row reads `50% off` — the opposite meaning. Same for something like `—Limited—`.

Anchoring the strip to a single separator occurrence keeps the label intact, e.g.:

```js
rawSubName = rawSubName
	.slice( md.parent_layer_name.length )
	.replace( /^\s*\p{Pd}\s*/u, '' );
```

(or only strip when the composite actually matches `parent + SEP`). Minor/edge-case — everything else in the change looks right to me: the `! md.product_name` guard correctly exempts Woo, `startsWith` avoids touching unrelated names, `parent_layer_name` landed in the same commit as the composite name so there's no historical-data gap, and an empty result still falls through to the `"<Type> #N"` ordinal.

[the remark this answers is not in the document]
subodhr258: Fixed in b91bdab4. The strip now requires the exact `<parent> — ` prefix (parent name + the tracker's ` — ` separator) instead of `startsWith(parent)`, so a bare label that only shares the parent name — parent `Sale`, label `Sale special` — is left intact rather than becoming `special`. Added a prefix-collision test that fails on the old logic and passes now.

subodhr258: Fixed in b91bdab4. Went with your "only strip when the composite actually matches `parent + SEP`" option, which resolves this greedy-dash case and the Copilot prefix-collision case in one check. The strip now matches the exact `<parent> — ` prefix and slices just that — no character-class replace — so a label that begins with its own dash keeps it: `Summer sale — -50% off` -> `-50% off`, not `50% off`.

The separator is now a module constant `HOTSPOT_NAME_SEPARATOR = ' — '` with a comment pointing at `hotspotLayerManager.js:213`, since that is the source of truth for the composite. Added two tests (leading-dash and prefix-collision); both fail on the old logic and pass now. 15/15 in the suite.

KMchaudhary (approved):

Issue #2073, opened 2026-08-11
GoDAM Document GB, Elementor, WPBakery block supports .docx, .xlsx, .pptx, ODF, .txt, .csv files same as GoDAM Central


KMchaudhary:
# Document block — support .docx, .xlsx, .pptx, ODF, .txt, .csv files.

## Context

GoDAM Central shipped a **document pipeline** ([godam-core#1307](https://github.com/rtCamp/godam-core/pull/1307), spec in [issue #1170](https://github.com/rtCamp/godam-core/issues/1170#issuecomment-5250385396)). Central now accepts Word, Excel, PowerPoint, OpenDocument, TXT and CSV alongside PDF, converts each to a **preview PDF** server-side (Gotenberg/LibreOffice), and exposes it as a new job field `preview_pdf_url`.

The WordPress plugin is still PDF-only end to end: `godam/pdf` (title "Document") accepts `application/pdf` only, the transcoder handler only sends `.pdf` to Central, and the front end embeds the file in `<object type="application/pdf">`.

**Goal:** upload any supported document in WP → it transcodes on Central → the block renders its preview PDF on the front end through a chrome-less pdf.js viewer, so a visitor looking at a `.docx` never sees that it is really a PDF. Downloads always serve the original file, never the generated preview.

### Decisions taken

- pdf.js renders **every** document, native `.pdf` included — one code path, consistent chrome-less look.
- Same library as Central: `react-pdf@9.2.1` + `pdfjs-dist@4.8.69` (react-pdf pins that exact pdfjs version).
- Files with no preview (password-protected) get a **download-only panel**; no in-browser unlock this iteration.
- In scope: the block, the transcode/callback plumbing, the Media Library "GoDAM" tab, the Elementor / WPBakery / shortcode wrappers, and retranscoding existing documents.

kishan-gondaliya-7270:
Fixed and released

PR #2079, opened 2026-08-14
Extend the Document block beyond PDF to docx/doc, xlsx/xls, pptx/ppt, odt/ods/odp, txt and csv.

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam/issues/2073

This pull request adds robust support for Office, OpenDocument, plain-text, and CSV files as previewable documents alongside PDFs. It ensures that only supported document formats are processed, both in the WordPress admin and in the block editor, and introduces new logic to handle document-specific transcoding, error reporting, and preview management. The changes also improve user feedback and maintain consistency between the client and server regarding supported formats.

**Document Format Support and Consistency**

- Added a comprehensive list of supported document MIME types and extensions (including DOCX, XLSX, PPTX, ODT, CSV, TXT, etc.) in both the backend (`RTGODAM_Transcoder_Handler` and helpers) and frontend (`SUPPORTED_DOCUMENT_TYPES` in `constants.js`), ensuring the block and server always agree on which formats are allowed. [[1]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868R95-R111) [[2]](diffhunk://#diff-568a16992779119b9bba5c54ec3ed179c9a67230591e9d2f4eea878bcaa75141R1-R163)
- Updated the block description and keywords in `block.json` to reflect support for a wide range of document types, not just PDFs.

**Transcoding Logic and Preview Handling**

- Modified transcoding logic so that Office/OpenDocument/text files are converted to a preview PDF by GoDAM Central, while PDFs are passed through directly. Only supported extensions are accepted for transcoding, preventing unsupported files (e.g., `.srt`, `.c`) from being misrouted. [[1]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868L313-R366) [[2]](diffhunk://#diff-4f7d5e41b95ad49276b75df0eccbc9295818361e0ad9f8949e6d7f5d7e079868R388-R401)
- Implemented logic to clear outdated previews when retranscoding or replacing a document, preventing stale previews from being shown if the file changes.
- On callback, stores or deletes the preview PDF URL as appropriate, ensuring accurate rendering and download behavior for both PDFs and converted documents.

**Error Reporting and Status Handling**

- Extended the REST API to accept and store a machine-readable `error_code` for failed transcodes, allowing the UI to distinguish between retryable and permanent errors. [[1]](diffhunk://#diff-2c8260c84a22b264dd8ccfd5d58e5306e47db95a97a82118b0d72176c3b2baf2R125-R151) [[2]](diffhunk://#diff-2c8260c84a22b264dd8ccfd5d58e5306e47db95a97a82118b0d72176c3b2baf2R257) [[3]](diffhunk://#diff-2c8260c84a22b264dd8ccfd5d58e5306e47db95a97a82118b0d72176c3b2baf2R275-R284)
- Defined a strict allow-list of in-progress transcoding statuses in the frontend, ensuring only recognized statuses are treated as "in flight" and preventing indefinite waiting on unrecognized states.

**Media Library and Admin Integration**

- Updated media row actions and transcoded URL field logic to use attachment-based checks (not just MIME type), so only truly supported documents offer previews or transcoding actions, excluding files like `.srt` or `.c` that share text/plain. [[1]](diffhunk://#diff-0d4b5d9fe699dc8791e18e5ba3ea3c125885c17e6c1ff6815ee718e53481e7afR243-R251) [[2]](diffhunk://#diff-2664911a55379dd72ac8c95fceda9c602329340edbc5fd2f1d70ed10ba5a0131R32-R38)

These changes together provide seamless, accurate support for a wide variety of document formats, improving both the backend processing and the user experience in the editor and media library.

## Demo

**Gutenberg Blocks**

https://github.com/user-attachments/assets/a24b0896-a024-47fe-95f5-0ee72df13692

**Elementor Widget**

https://github.com/user-attachments/assets/30e4b0ac-9ddc-4187-bb9d-fa9634d62642

**WPBakery Element**
Screenshot 2026-08-19 at 1 20 03 PM


assets/src/blocks/godam-pdf/constants.js
```diff
@@ -0,0 +1,117 @@
...
+	'text/plain': 'txt',
+	'text/csv': 'csv',
+	// Some servers report .csv as application/csv; Central accepts both.
+	'application/csv': 'csv',
+};
+
+/**
+ * MIME types, for MediaUpload's `allowedTypes`.
+ */
+export const ALLOWED_MEDIA_TYPES = Object.keys( SUPPORTED_DOCUMENT_TYPES );
```
subodhr258, line 31: `ALLOWED_MEDIA_TYPES` includes `text/plain`, and it is used as `allowedTypes` / `library.type` for every document picker (block, Elementor, WPBakery). WordPress maps `.srt`, `.asc`, `.c`, `.cc`, `.h` to `text/plain` too, so those files appear as selectable in the library picker even though they have no conversion path.

The upload dialog is limited by extension via `DOCUMENT_ACCEPT`, and the transcoder guards by extension (`rtgodam_is_supported_document_attachment`), but library *selection* filters by MIME. So a user can pick an existing `.srt`/`.c`/`.h`, `isSupportedDocument()` accepts it on MIME, and the block then only ever shows the 'no preview available' panel.

Worth deciding whether that is acceptable, or whether the picker / `isSupportedDocument()` should also apply the extension gate so these text/plain lookalikes are not offered as documents.

inc/helpers/custom-functions.php
```diff
@@ -707,51 +707,95 @@ function godam_is_audio_file( $file_path_or_url ) {
...
+ * block only ever renders a PDF whatever the author uploaded — but the *upload* has to be
+ * accepted here first, and the transcoder has to label the job `document` rather than `pdf`.
  *
- * Callers use this to skip front-end output entirely and to show an "unsupported
- * format" notice in the editors instead.
+ * This is the single source of truth on the PHP side; the editor's copy lives in
+ * assets/src/blocks/godam-pdf/constants.js and the two are asserted equal by
+ * tests/php/DocumentSupportTest.php.
+ *
+ * @since 2.2.0
```
subodhr258: This branch still carries `@since n.e.x.t` placeholders for other unreleased changes (`inc/classes/class-assets.php:171`, `inc/classes/shortcodes/class-godam-player.php:252`), and the plugin is currently `2.1.1`. This PR hardcodes `@since 2.2.0` in 7 places, which fixes the next release number before it is decided and skips the release tooling that replaces `n.e.x.t`. Suggest using `@since n.e.x.t` to match the rest of the branch.

assets/src/blocks/godam-pdf/view.js
```diff
@@ -1,268 +1,165 @@
...
+	// React owns the container from here, so clear the server-rendered fallback first.
+	wrapper.textContent = '';
+
+	const root = createRoot( wrapper );
+
+	root.render(
+		createElement( DocumentViewer, {
+			url: previewUrl,
+			title,
+			onError: () => {
```
subodhr258, line 128: `react-pdf` can invoke more than one error callback for a single load: `onLoadError`, `onSourceError` and `onPassword` all route to `handleError`, which calls this `onError`. If it fires twice, `root.unmount()` runs twice and `renderNativeFallback()` appends a second `<object>`. Consider making the fallback run once, e.g. a `handled` flag set before the first `renderNativeFallback()`.

assets/src/blocks/godam-pdf/edit.js
```diff
@@ -421,30 +486,121 @@ function PdfEdit( {
...
+						{ __( 'Loading preview…', 'godam' ) }
+					</p>
+				}
 			>
-				<embed
-					src={ src }
-					type="application/pdf"
-					width="100%"
-					height="100%"
+				<DocumentViewer
```
subodhr258: The front end passes an `onError` to `DocumentViewer` that falls back to a native `<object>` and keeps the download link, but the editor canvas does not pass one here. On a CORS or load failure the author sees only 'This preview could not be loaded.' with no download, while a visitor on the published page still gets the file via the native fallback. Consider passing an `onError` here that surfaces the same download-only / unavailable panel, so the editor matches what visitors actually see.

admin/class-rtgodam-transcoder-handler.php
```diff
@@ -92,6 +92,20 @@ class RTGODAM_Transcoder_Handler {
...
+	 * Office / OpenDocument / plain-text formats, which GoDAM Central converts to a preview
+	 * PDF. They are kept apart from $other_extensions because those map the extension
+	 * straight onto the job type ('pdf' => job_type 'pdf'), whereas every format here shares
+	 * the single job type 'document'.
+	 *
+	 * @since    2.2.0
+	 * @access   public
+	 * @var      string    $document_extensions    Document extensions with comma separated.
+	 */
+	public $document_extensions = ',docx,doc,xlsx,xls,pptx,ppt,odt,ods,odp,txt,csv';
```
subodhr258: Minor / latent: `explode( ',', $this->document_extensions )` produces a leading empty-string entry (same shape as `$other_extensions`), so `in_array( $extension, ..., true )` is true for a file with no extension. It is currently masked (the early document-MIME gate returns first, and the audio branch also matches `''`), but it is a footgun. Dropping the leading comma, or `array_filter()`-ing the exploded list, removes the empty entry.

subodhr258 (commented):
Reviewed the full diff (excluding the lockfile). This is a careful, well documented change and the core design holds together:

- the preview-vs-download URL split is consistent across `render.php`, the editor, and the `rtgodam_get_document_preview_url()` / `rtgodam_get_document_download_url()` resolvers;
- legacy PDF blocks keep rendering with no migration (`test_preview_url_falls_back_for_legacy_pdfs`);
- `pdfjs-dist` (4.8.69) matches what `react-pdf` 9.2.1 depends on, so there is no worker/API version mismatch;
- the `preview_pdf_url` path is wired end to end (Central to ajax to JS import, plus the transcoder callback), and download always serves the original upload;
- the PHP-to-JS constant-sync test is a solid guard against the two lists drifting.

A few non-blocking things left inline:
- Document pickers request `text/plain`, which also surfaces `.srt/.c/.h/.asc` as selectable (medium).
- `@since 2.2.0` is hardcoded while the branch still uses `@since n.e.x.t` elsewhere.
- `view.js` error fallback is not idempotent.
- The editor viewer has no `onError`/download fallback, unlike the front end.
- `document_extensions` leading-comma `explode` footgun.

One more that could not be anchored inline (the line is outside the diff): `assets/src/blocks/godam-pdf/edit.js:278` still reads `// Reset so useEffect re-counts pages for the new file.`, but the `useEffect` was removed in this PR (counting now happens in `onPreviewLoaded`), so that comment is now stale.

assets/src/blocks/godam-pdf/viewer/index.js
```diff
@@ -0,0 +1,391 @@
...
+			},
+			{ root: container, rootMargin: `${ RENDER_MARGIN_PX }px 0px` },
+		);
+
+		slotRefs.current.forEach( ( slot ) => slot && observer.observe( slot ) );
+
+		return () => observer.disconnect();
+		// Re-observes when the page count changes, which is when the slots themselves are
+		// replaced. Page width is deliberately not a dependency: a resize moves the same slots.
+	}, [ numPages ] );
```
subodhr258, line 263: The on-demand rendering can leave a document showing only the first 2 pages if the viewer loads while its container is zero-width.

This effect is keyed on `[numPages]` and observes whatever slots exist when it runs (`slotRefs.current.forEach( ... observe )`), but the slots below only render once `pageWidth > 0`. If `numPages` is set while the container is zero-width (a Document block inside a collapsed tab/accordion, which is exactly the "hidden viewer" case the `INITIAL_PAGES` comment describes), there are no slots to observe, and this effect never re-runs when the panel is later expanded and the slots appear. `pagesInView` then stays at the `INITIAL_PAGES` seed, so only the first 2 pages mount and the rest stay blank placeholders.

A background tab is fine (non-zero width, so slots exist and get observed), but a zero-width-at-load container is not. Re-running the observer once slots first appear fixes it, e.g. add a `pageWidth > 0` boolean to this effect's deps, or call `observer.observe()` from the slot ref callback.

subodhr258:
Re-reviewed (commits through `dd0d907`). Every point from the first review round is addressed, and the classifier changes now have test coverage in `tests/php/DocumentSupportTest.php`. Verified each:

**Copilot's points**
1. Classifier no longer MIME-only: `godam_is_supported_document()` now requires the MIME type and the extension to agree, falling back to MIME alone only when the file cannot be resolved. Tested (`.srt/.asc/.c/.cc/.h` rejected, `.txt` kept). ✓
2. `isSupportedDocument()` mirrors that on the client via `hasUnsupportedDocumentExtension()`. ✓
3. Editor uses `sourceUrl` before `src`, through an ordered `previewCandidates` list with per-URL failure fallback. ✓
4. `get_mime_type_for_job_type()` infers the MIME from `orignal_file_name` instead of labelling converted docs `application/pdf`; PDF is the last resort. ✓
5. Page count is held in state and committed from an effect when the block becomes selected. ✓
6. Viewer renders pages on demand via IntersectionObserver, with correct-height placeholders. ✓
7. Scroll container is now a focusable `role="region"`. ✓
8. Stale preview is cleared at queue time in the handler, so failed or pending replacements no longer serve the old document. ✓

**My earlier points**
9. text/plain lookalikes: still selectable from the library picker (WordPress filters that picker by MIME), but they now surface the unsupported panel instead of a broken block. Reasonable resolution. ✓
10. `@since 2.2.0` reverted to `@since n.e.x.t`. ✓
11. `handledError` guard prevents the double unmount and double `<object>`. ✓
12. Editor canvas now has an error path (error boundary plus candidate fallback into the download-only panel). ✓
13. Leading empty extension fixed (dropped the comma and added a `'' !== $extension` guard). ✓

The per-document pdf.js worker handover in `viewer/worker.js` is also the right fix for the shared-worker teardown, and the JS/PHP parity test guards the two lists from drifting.

Two things before merge:

**1. On-demand viewer can miss pages when it loads at zero width** (`viewer/index.js`, inline comment added). The IntersectionObserver effect is keyed on `[numPages]` and observes only the slots that exist when it runs, but slots render only once `pageWidth > 0`. If `numPages` arrives while the container is zero-width (a Document block inside a collapsed tab or accordion, the "hidden viewer" case the `INITIAL_PAGES` comment calls out), the observer attaches to nothing; when the panel is expanded later the slots render but the effect does not re-run, so only the 2 seed pages ever mount and the rest stay blank placeholders. Re-running the observer once slots first appear fixes it (add a `pageWidth > 0` signal to the effect deps, or call `observer.observe()` from the slot ref callback).

**2. Merge conflict.** The PR currently conflicts with `develop` in `assets/src/js/media-library/views/attachment-detail-two-column.js` (develop's media-library work touched the same file). It needs a rebase or merge before it can go in.

Otherwise this looks good.

assets/src/js/media-library/views/attachment-detail-two-column.js
```diff
@@ -1175,7 +1175,10 @@ export default AttachmentDetailsTwoColumn?.extend( {
 			this.renderAudioActions();
 		}
 
-		if ( this.model.get( 'type' ) === 'application' && this.model.get( 'subtype' ) === 'pdf' ) {
+		// Any convertible document, not just PDF: Central rasterises page 0 of the preview for
+		// all of them, and set_media_library_thumbnail() puts it on `image`, so an .xlsx has a
+		// real preview to show here instead of the generic document icon.
+		if ( isDocumentModel( this.model ) ) {
```
subodhr258, line 1181: Blocking: `isDocumentModel` is used here but is not imported in this file, so this throws `ReferenceError: isDocumentModel is not defined`.

The `develop` merge (222e273b) resolved the import-line conflict to `import { canManageAttachment } from '../utility';` (line 16), dropping the `isDocumentModel` the pre-merge PR tip imported. The symbol is still exported by `utility.js` and imported correctly in `attachment.js`, just not here. Because this sits inside `render()` (line 991) and the call evaluates the condition unconditionally, opening the two-column attachment details for **any** attachment throws at the end of render.

Fix: add it back to the import on line 16, `import { canManageAttachment, isDocumentModel } from '../utility';`.

subodhr258:
Verified both items from the last round are fixed:

- **Zero-width viewer** (`f015e021`): the IntersectionObserver is now stashed in a ref and each slot observes itself from its ref callback, so slots that first appear when a collapsed / zero-width viewer is revealed get picked up. Correct, and the unobserve-on-replace and unmount handling is right.
- **Merge conflict resolved** (`222e273b`). The new WPBakery inline-editor asset loading (`0414c618`) also looks sound (double-mount guard before the first `await`, Elementor `element_ready` plus a WPBakery preview observer, all idempotent).

One blocking regression the `develop` merge introduced, so I can't approve yet:

`assets/src/js/media-library/views/attachment-detail-two-column.js` uses `isDocumentModel( this.model )` in `render()` (line 1181), but the merge dropped `isDocumentModel` from this file's import (line 16 now imports only `canManageAttachment`). The symbol is still exported by `utility.js` and imported fine in `attachment.js`, so it is just this one file. As written it throws `ReferenceError: isDocumentModel is not defined` every time the two-column attachment details render, for any attachment. One-line fix (inline comment).

Once that import is restored I'll approve.

KMchaudhary:
Thanks for catching that — you're right, and it wasn't document-specific: `render()` calls `isDocumentModel` unconditionally, so the missing import threw for every attachment in the two-column details view.

Fixed in b451498f by restoring the symbol to this file's import:

```js
import { canManageAttachment, isDocumentModel } from '../utility';
```

I also linted the whole `media-library` and `godam-pdf` trees to check the merge hadn't dropped anything else — no `no-undef` anywhere, so this was indeed the only affected file. The remaining lint errors there are pre-existing and untouched by this PR (`import/no-unresolved` on `@wordpress/block-editor` in `caption.js`/`edit.js`, and `no-unsafe-wp-apis` on the `__experimentalToggleGroupControl` imports).

Ready for another look.

subodhr258:
The code is good now — `isDocumentModel` is restored to the import in `attachment-detail-two-column.js` (`b451498f`), the zero-width viewer fix is correct, and every point from the earlier rounds is addressed. So this is approvable on the code.

Holding the approval on one CI thing: **Run JS unit tests is failing at `npm ci`** with `Missing: … from lock file` (dompurify@3.2.7, @babel/core@7.25.7, markdown-it, @opentelemetry/*, date-fns, yaml, and more), which means `package-lock.json` is out of sync with `package.json` on this branch. This is specific to this PR, not the old base-wide drift: #2085 passes the same check on the same `develop` base. Most likely the `develop` merge left the lockfile inconsistent.

Fix: run `npm install` on the branch and commit the regenerated `package-lock.json`, then confirm the JS unit tests job goes green. Once it does I'll approve.

(The red POT and Plugin Check are the usual pre-existing/expected ones.)

subodhr258 (approved):

PR #2080, opened 2026-08-16
feat: add rtgodam_before_attachment_lookup and rtgodam_after_attachment_lookup hooks around media

Pathan-Amaankhan:
## Summary

Adds the `rtgodam_before_attachment_lookup` / `rtgodam_after_attachment_lookup` hook pair around every place across the plugin that directly reads or writes WordPress attachment data — a new capability that lets a plugin centralizing attachments across a multisite network (e.g. wp-dam) know when GoDAM is about to touch attachment data on the current request, so it can switch to whichever site actually holds it, then switch back. Companion to #2072, which adds the CI + pre-commit checks that guard this pair against silently regressing.

## Where hooks were added

- **Transcoder pipeline** (`admin/`) — thumbnail generation, retranscoding, job-status polling and callbacks.
- **REST API endpoints** (`inc/classes/rest-api/`) — media library, transcoding, transcription, video migration, analytics, ads, engagement, sync.
- **Block render templates** (`assets/src/blocks/`, `inc/templates/`) — audio, image, PDF, player, video-duration, video-thumbnail, video-gallery.
- **One-time migrations** (`inc/classes/migrations/`) — legacy gallery/shortcode formats, discontinued CPT cleanup.
- **Page-builder integrations** — Elementor, WPBakery, SureForms, Ninja Forms, WPForms.
- **Shared helpers** — `inc/helpers/custom-functions.php`, media-usage tracking/backfill, SEO schema caching.

## Multisite-safety fix bundled in

Media-usage tracking's reverse-index meta was previously keyed by bare post ID, which can collide across sites on a centralized network — two different sites can have the same numeric post ID. Now keyed by a composite `"{blog_id}:{post_id}"` string (`make_usage_key()` / `make_post_ref_key()`), captured *before* any site-context switch so it always reflects the site the data actually belongs to, not whichever site the hook pair happens to be switched to at that moment.

## Coverage

- 116 call sites reviewed and marked with an inline `godam-coverage-ignore` comment (the sibling coverage-checker from #2072), each with its own reason directly at the call site — mostly host-post-scoped reads/writes (SEO schema cache, Elementor edit-mode flags, media-usage tracking markers, one-time migrations scoped by `post_status`/CPT) that never touch attachment data despite superficially matching the checker's heuristics.
- Fixed one real gap a reviewer found: the gallery block's "query" mode ran its own `WP_Query` outside any hook — on a centralized site that query would return nothing locally, so the (correctly-wrapped) per-item resolver right next to it would never even run. Now bracketed too, with restoration before `wp_reset_postdata()`.
- `@since 2.2.0` across all 128 new docblocks — this is a backwards-compatible new capability (pure `do_action()` calls, no default listeners), so it lands as a minor version bump from the current 2.1.1.

## Also included

Two small player improvements: the layer-manager's rich-content fields (`html`/`text`) are now stripped from the player's JSON config before encoding — the layer JS already reads the rendered HTML from the DOM, so duplicating it into `data-options` was unnecessary — and shortcodes now run in layer CTA HTML by default (e.g. a form shortcode), rather than requiring a separate filter listener to enable it.


inc/classes/rest-api/class-analytics.php
```diff
@@ -1039,6 +1059,14 @@ public function fetch_top_videos( WP_REST_Request $request ) {
 			}
 		}
 
+		/**
+		 * Fires after enriching top-video rows, so integrations can restore
+		 * the site context switched in `rtgodam_before_attachment_lookup`.
+		 *
+		 * @since 1.8.0
```
KMchaudhary: @since should be n.e.x.t so we can change all the next release version number at release time

Pathan-Amaankhan: Updated in https://github.com/rtCamp/godam/pull/2080/commits/7aaf12cbffcb429df10acafdd18506f857b39896

assets/src/js/media-library/views/attachment-browser.js
```diff
@@ -20,15 +20,24 @@ const AttachmentsBrowser = wp?.media?.view?.AttachmentsBrowser;
...
-		// Initialize the parent class.
+		// Initialize the parent class. Its own initialize() already calls
+		// this.bindEvents() (dispatched through the prototype chain, so it
+		// runs the override below) — calling it again here would double-bind
+		// every listener it registers.
 		AttachmentsBrowser.prototype.initialize.apply( this, arguments );
 
 		this.updateCollectionObserve();
-		this.bindEvents();
 		this.addUploadParam();
```
KMchaudhary: **`bindEvents()` is now never called — both `collection.props` listeners are unbound.**

This comment isn't accurate for `AttachmentsBrowser`. Checked against WP 7.0.4 (`wp-includes/js/media-views.js`):

- `AttachmentsBrowser = View.extend({...})` starts at line **4389**; there is no `bindEvents` anywhere in 4389–5095, and its `initialize()` calls only `createSidebar` / `createUploader` / `createToolbar` / `createAttachmentsHeading` / `createAttachmentsWrapperView` / `createLoadMoreView` / `updateContent` / `updateLoadMoreView`.
- The `bindEvents` at line **4016** belongs to a *different* view — `wp.media.view.Attachments` (declared line 3893) — which reaches it via its own `this.on( 'ready', this.bindEvents )`.
- Neither `wp.media.View` nor `wp.Backbone.View` defines it either.

So nothing calls `bindEvents()` any more. The one-shot `updateCollectionObserve()` / `addUploadParam()` in `initialize()` still run, so the browser looks fine on open — but `this.collection.props.on( 'change', ... )` (lines 41-42) never binds, so changing a folder / filter / search term no longer re-runs either handler. The `if ( AttachmentsBrowser.prototype.bindEvents )` guard below is dead code for the same reason.

```suggestion
		// Initialize the parent class. AttachmentsBrowser's own initialize()
		// does NOT call bindEvents() (that belongs to wp.media.view.Attachments),
		// so the override below has to be invoked explicitly.
		AttachmentsBrowser.prototype.initialize.apply( this, arguments );

		this.updateCollectionObserve();
		this.bindEvents();
		this.addUploadParam();
```

inc/classes/class-media-library-ajax.php
```diff
@@ -1325,41 +1428,55 @@ public function filter_rtgodam_content_img_tag( $filtered_image, $context, $atta
...
+		/**
+		 * Fires before reading this attachment's virtual-media original-ID
+		 * and GoDAM CDN image-size/transcoded-URL meta, so integrations
+		 * that centralize media on another site can switch context first.
+		 * Wrapped in try/finally because this method returns from several
+		 * points while deciding whether to rewrite the rendered <img> src.
+		 *
+		 * @since 2.2.0
+		 */
+		do_action( 'rtgodam_before_attachment_lookup' );
```
KMchaudhary, line 1460: **The wrap below is unreachable — the `get_post_mime_type()` gate above it isn't wrapped.**

Line 1426 reads `$mime_type = get_post_mime_type( $attachment_id )` and line 1428 returns early if it isn't an image — but the pair only opens here at 1440. `get_post_mime_type()` resolves the attachment post row's `post_mime_type` column, exactly like the `get_post()` calls this PR wraps everywhere else. If that row needs the switch, so does this read.

So on a centralized install line 1426 returns empty, line 1428 returns, and the `get_post_meta()` calls this docblock was written for never run. CDN image rewriting in post content silently no-ops.

`filter_virtual_media_srcset()` has the identical shape a few methods up: mime read at **1260**, early return at **1262**, bracket at **1279**. Same for `upload_media_to_frappe_backend()` — mime read at **303**, bracket at **348**.

Outside this file, `inc/templates/video-preview.php:21` does the same read with no bracket anywhere in that file, which makes `$godam_is_image` / `$godam_is_audio` both false and labels every image and audio preview "Video Preview".

Root cause is in the companion PR: `get_post_mime_type` isn't in `ACCESS_FUNCTIONS` in `bin/hook-check/coverage.php`, so the gate is blind to it. Worth adding it there and re-running.

admin/godam-transcoder-actions.php
```diff
@@ -19,117 +19,131 @@
...
+			$transcoded_url = rtgodam_get_transcoded_url_from_attachment( $post->ID );
+			$job_id         = rtgodam_get_job_id_by_attachment_id( $post->ID );
 
-		$easydam_settings = get_option( 'rtgodam-settings', array() );
+			$easydam_settings = get_option( 'rtgodam-settings', array() );
 
-		// Determine if the site has a valid API key (i.e., Premium user).
-		$api_key = get_option( 'rtgodam-api-key', '' );
+			// Determine if the site has a valid API key (i.e., Premium user).
+			$api_key = get_option( 'rtgodam-api-key', '' );
```
KMchaudhary: **`get_option()` here reads the *media* site's options, so the Premium gate uses the wrong site's API key.**

The pair opens at line 37 and its `finally` closes at 146 — the whole function body. `switch_to_blog()` calls `$wpdb->set_blog_id()`, so inside that span:

- `get_option( 'rtgodam-settings', array() )` (59) and `get_option( 'rtgodam-api-key', '' )` (62) come from the media site's `wp_options`, but the API key is per-site config.
- Line 64's `if ( ! empty( $api_key ) )` therefore gates on the wrong site: a site with its own valid key gets the locked "Available in Premium version" upsell (120-141) when the media site has none, and a site *without* a key sees the oEmbed / Transcoded CDN URL fields when the media site has one.
- `admin_url( 'admin.php?page=rtgodam_settings#video-settings' )` (137) then points the "API key" link at the media site's admin, which the user may have no access to.

Only lines 41, 49, 56, 57 and 105 are actual attachment reads. Tightening the bracket to just those fixes all three at once — and lines 64-141 are pure string/HTML building that doesn't want to be switched anyway.

Minor, while you're here: `$easydam_settings` (59) is assigned and never used in this function.

inc/classes/class-media-usage-tracker.php
```diff
@@ -424,76 +463,157 @@ public function unregister_media_usage( $attachment_id, $post_id, $source = 'con
...
-				$normalized[ (int) $post_id ] = array_values( array_filter( array_map( 'strval', (array) $sources ) ) );
+			foreach ( $map as $map_key => $entry ) {
+				if ( is_array( $entry ) && isset( $entry['post_id'], $entry['sources'] ) ) {
+					// Current shape already.
+					$blog_id = isset( $entry['blog_id'] ) ? (int) $entry['blog_id'] : get_current_blog_id();
+					$post_id = (int) $entry['post_id'];
+					$sources = array_values( array_filter( array_map( 'strval', (array) $entry['sources'] ) ) );
+				} else {
+					// Legacy shape: bare int post_id => string[] sources.
+					$blog_id = get_current_blog_id();
```
KMchaudhary: **This `get_current_blog_id()` runs *inside* the caller's switch, so legacy entries migrate under the media site's blog ID.**

`register_media_usage()` gets this exactly right — it captures `$blog_id` at line 349 *before* firing `before()` at 362. But it then calls `get_usage_sources()` at 364 and `save_usage_sources()` at 380, both still inside that open bracket (closed at 384). `switch_to_blog()` sets `$GLOBALS['blog_id']`, so `get_current_blog_id()` at 524, 529 and 549 returns the **media** site.

Concretely, a legacy entry `123 => ['content']` migrates to key `"<media_blog>:123"`, while the caller looks up `"<post_blog>:123"` — a guaranteed miss. Knock-on effects:

1. `$was_referenced` is false for an already-registered pair, so `schedule_log_media_view()` re-notifies GoDAM Central.
2. A second entry is written next to the mis-keyed one and persisted by `save_usage_sources()`.
3. `unregister_media_usage()` (line 420 — same correct capture, same wrong helper) also computes the post-site key, so it can never delete the orphan. It also double-counts post 123 in the derived `ATTACHMENT_META_KEY` list.

`save_usage_sources()`'s own fallback at line 574 is switched for the same reason (its `$clean` loop runs before its *own* `before()` at 588, but it's already inside the caller's).

Cleanest fix is to pass `$blog_id` down into both helpers rather than re-reading it — the callers already have the correct value.

inc/classes/class-media-library-ajax.php
```diff
@@ -953,10 +989,24 @@ private function godam_replace_attachment_with_external_file( $attachment_id, $f
...
+		try {
+			// Validate the attachment.
+			$attachment = get_post( $attachment_id );
+			if ( ! $attachment || 'attachment' !== $attachment->post_type ) {
+				return new \WP_Error( 'invalid_attachment', __( 'Invalid attachment ID.', 'godam' ) );
+			}
+		} finally {
+			do_action( 'rtgodam_after_attachment_lookup' );
```
KMchaudhary: **Closing the bracket here leaves the two attachment *writes* below unwrapped.**

This function only has brackets at 1001-1009 and 1068-1070. Everything in between runs unswitched, including:

- `update_attached_file( $attachment_id, $new_filepath )` (**1045**) — writes the attachment's own `_wp_attached_file` meta
- `wp_update_post( array( 'ID' => $attachment_id, 'post_mime_type' => ... ) )` (**1050**) — writes the attachment's post row

So the read at 1001-1009 resolves against the media site while both writes land locally: the file replacement silently has no effect on the real attachment, and may create or clobber a same-ID row on the local site.

Neither `update_attached_file` nor `wp_update_post` is in `ACCESS_FUNCTIONS`, which is why the checker stays green here. Extending this bracket down over 1045-1055 (or wrapping that block in its own pair) covers it.

inc/classes/media-library/class-media-folder-utils.php
```diff
@@ -57,80 +57,96 @@ public function get_attachment_count( $folder_id, $force_refresh = false, $mime_
...
+		 * this covering the transient too, each site would cache its own
+		 * possibly-stale copy of what's supposed to be one shared count.
+		 *
+		 * @since 2.2.0
+		 */
+		do_action( 'rtgodam_before_attachment_lookup' );
+		try {
+			// Try to get cached count first (unless force refresh is requested).
+			if ( ! $force_refresh ) {
+				$cached_count = get_transient( $cache_key );
```
KMchaudhary: **The transient is keyed by a *local* term ID but stored on the media site.**

`$cache_key` (line 58) is `'attachment_count_' . $folder_id . '_' . $mime_type_key`, and `$folder_id` is a term ID in the **calling** site's own `media-folder` taxonomy. Both `get_transient()` here and `set_transient()` at 144 sit inside the `try` opened at 71-72 — transients are per-blog options, and `switch_to_blog()` switches both `$wpdb` and the object-cache prefix, so on a centralized network the cache lives on the media site and is shared by every site on the network.

Multisite term IDs are per-site (`wp_<N>_terms`), so site A's folder 5 and site B's folder 5 are different folders that collide on `attachment_count_5_all`. Site B renders site A's count for an unrelated folder for up to `CACHE_EXPIRATION`, and each site keeps overwriting the other's value.

This is the same collision class the PR deliberately fixes in `Media_Usage_Tracker` (capture site-local state before the switch). Moving the transient get/set outside the bracket — the `$wpdb` query is the only part that needs the switch — resolves it.

inc/templates/godam-player.php
```diff
@@ -998,7 +1049,13 @@ class="easydam-layer hidden"
 									<div id="layer-<?php echo esc_attr( $godam_instance_id . '-' . $godam_layer['id'] ); ?>" class="easydam-layer hidden" style="background-color: <?php echo isset( $godam_layer['bg_color'] ) ? esc_attr( $godam_layer['bg_color'] ) : '#FFFFFFB3'; ?>">
 										<?php if ( $godam_render_html_cta ) : ?>
 											<div class="easydam-layer--cta-html">
-												<?php echo wp_kses_post( $godam_layer['html'] ); ?>
+												<?php
+												// Run shortcodes in CTA HTML by default (e.g. a Marketo form
+												// shortcode) rather than requiring an extension-side filter
+												// listener to enable it; `rtgodam_cta_html_content` remains as
+												// an optional override point.
+												echo apply_filters( 'rtgodam_cta_html_content', do_shortcode( wp_kses_post( $godam_layer['html'] ) ), $godam_layer ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- wp_kses_post() already sanitizes; do_shortcode()/the filter run after, same pattern as godam_preview_page_content().
```
KMchaudhary: **Sanitize-then-expand: `wp_kses_post()` runs *before* `do_shortcode()`, so the shortcode's output is emitted unsanitized.**

Three separate concerns on this one line:

1. **Ordering.** The `phpcs:ignore` says "wp_kses_post() already sanitizes; do_shortcode()/the filter run after" — but that's the problem, not the justification. kses filters the *input*; whatever `do_shortcode()` then expands, and whatever the filter returns, goes out raw with `EscapeOutput` suppressed.
2. **New default behaviour.** Shortcodes in layer CTA HTML were previously inert; they now execute for anonymous visitors on every public page load. That means anyone who can edit a GoDAM video layer gains shortcode execution on the front end, where before it required an extension to opt in via a filter listener.
3. **The cited precedent doesn't hold.** `godam_preview_page_content()` (`inc/helpers/custom-functions.php:1337`) calls `do_shortcode( $godam_shortcode )` on a shortcode string *the plugin itself builds*. This call passes arbitrary editor-authored HTML. Not the same pattern.

If the expansion is intended (e.g. for a form shortcode), it'd be good to say so explicitly and sanitize *after* expansion, or gate it on a capability.

Separately: `rtgodam_cta_html_content` here and `rtgodam_video_layers_for_js` at line 551 are both **new in this PR** — neither name appears anywhere else in `inc/`, `assets/` or `pages/`, and `rtgodam_cta_html_content` isn't in the base template at all. But both comments say they "remain as an optional override point", which reads as pre-existing API. Given this PR adds 128 `@since 2.2.0` action docblocks, these two `apply_filters()` calls having no docblock, `@since` or `@param` stands out.

admin/godam-transcoder-actions.php
```diff
@@ -160,8 +174,17 @@ function rtgodam_save_transcoded_url_field( $post, $attachment ) {
 			if ( ! current_user_can( 'edit_post', $post['ID'] ) ) {
```
KMchaudhary: **`current_user_can( 'edit_post', $post['ID'] )` needs the attachment row — but this one isn't wrapped, while the callback right above it wraps its entire body.**

`current_user_can( 'edit_post', $id )` routes through `map_meta_cap()`, which does `$post = get_post( $args[0] )` and returns `do_not_allow` when that's null (and for an `inherit`-status attachment it also resolves `get_post( $post->post_parent )`). So on a centralized install this check fails and line 175 returns `$post` unchanged — saving a transcoded URL from the attachment-details panel silently does nothing.

Worth noting `switch_to_blog()` does **not** re-initialise `$current_user` or `WP_Roles` (`wp-includes/ms-blogs.php` only touches `$wpdb`, `$table_prefix`, `$GLOBALS['blog_id']` and the object cache), so wrapping a meta-cap check is safe as well as necessary: the post lookup gets the media site while capability resolution correctly stays local. `inc/classes/rest-api/class-transcription.php:133` already does exactly this.

The contrast inside this one file is the tell — `rtgodam_add_transcoded_url_field()` above wraps 37-146, its save-side counterpart wraps nothing. A bracket around just the cap check (and any attachment reads below) would do it; `get_option( 'rtgodam-settings' )` at 164 is correctly left outside, so please keep it there.

Same gap at `inc/classes/class-pages.php:425` and `inc/classes/rest-api/class-media-library.php:818` — see the general review comment.

inc/classes/class-media-library-ajax.php
```diff
@@ -323,6 +338,14 @@ public function upload_media_to_frappe_backend( $attachment_id, $manual_retransc
...
 		$api_url = RTGODAM_API_BASE . '/api/resource/Transcoder Job' . ( empty( $transcoding_job_id ) ? '' : '/' . $transcoding_job_id );
 
+		/**
+		 * Fires before reading this attachment's URL/title/author, so
+		 * integrations that centralize media on another site can switch
+		 * context first.
+		 *
+		 * @since 2.2.0
+		 */
+		do_action( 'rtgodam_before_attachment_lookup' );
```
KMchaudhary, line 352: **`get_site_url()` inside this bracket reports the media site, and that value is what gets sent to the SaaS.**

`$site_url = get_site_url()` at line 357 takes no `$blog_id`, so while switched it returns the current (media) site's `siteurl`. It's then sent verbatim as `'wp_site' => $site_url` in the `wp_remote_request()` transcoding-job payload below.

On a centralized network that means every job submitted from any site is attributed to the media site, which breaks per-site attribution, quota and analytics on the SaaS side.

`Media_Usage_Tracker` goes to real trouble to get this right — `get_wp_site()` plus an explicit pre-switch `get_current_blog_id()` capture — so the two paths currently disagree about which site "owns" an upload. Reading `$site_url` before the `do_action()` on this line lines them up.

assets/src/blocks/godam-audio/render.php
```diff
@@ -18,6 +18,18 @@
 $godam_description   = ! empty( $attributes['description'] ) ? $attributes['description'] : '';
 $godam_thumbnail     = ! empty( $attributes['thumbnail'] ) ? esc_url( $attributes['thumbnail'] ) : '';
 
+/**
+ * Fires before reading this attachment's title/thumbnail/chapters/transcript
+ * meta and its audio source URLs (below), so integrations that centralize
+ * media on another site can switch context first. Closed explicitly before
```
KMchaudhary: **This claim isn't true — and `godam-pdf/render.php` in this same PR depends on the opposite.**

> since a bare `return` at file scope doesn't unwind through a `finally` the way a function's would

A file-scope `return` inside a `try` *does* run the `finally` before returning to the includer. Verified on PHP 8.4.5 (the version `.github/workflows/wp_dam_hook_integrity_on_pull_request.yml` pins):

```php
// inc.php
try { return; } finally { echo "FINALLY RAN\n"; }
// include 'inc.php';  =>  prints "FINALLY RAN"
```

`assets/src/blocks/godam-pdf/render.php` relies on exactly that: pair opened at 36, `try` over 38-77, bare file-scope `return`s at **62** and **76**, closed in `finally` at 78-79. So the two sibling templates now disagree — one documents a hazard that doesn't exist, the other correctly assumes its absence.

Practical cost here: three `do_action( 'rtgodam_after_attachment_lookup' )` calls (128, 141, 146) where one `try`/`finally` would do, and every future early `return` added to this template is a leak waiting to happen. Converting this file to the PDF template's shape would make them consistent and drop the manual bookkeeping.

(If there *is* a supported PHP version where this differs, worth naming it — because then `godam-pdf/render.php` is the broken one.)

admin/class-rtgodam-retranscodemedia.php
```diff
@@ -631,33 +668,33 @@ public function transcoded_thumbnails_added( $media_id = '' ) {
 			return;
 		}
 
-		$current_thumbnail = get_post_meta( $media_id, 'rtgodam_media_video_thumbnail', true );
-		$custom_thumbnails = get_post_meta( $media_id, 'rtgodam_custom_media_thumbnails', true );
+		$current_thumbnail = get_post_meta( $media_id, 'rtgodam_media_video_thumbnail', true ); // godam-coverage-ignore -- transcoded_thumbnails_added(): covered transitively — sole trigger is add_media_thumbnails()'s do_action( 'rtgodam_transcoded_thumbnails_added' ), which always fires from inside handle_callback()'s before/after pair.
```
KMchaudhary: **These dismissals rest on who fires a *public* action — which the plugin can't guarantee, and the checker can never re-verify.**

`transcoded_thumbnails_added()` is an `add_action( 'rtgodam_transcoded_thumbnails_added', ... )` listener (registered line 90). Eight `godam-coverage-ignore` comments in this method (671, 672, 680, …) all say the same thing: "sole trigger is `add_media_thumbnails()`'s `do_action(...)`, which always fires from inside `handle_callback()`'s before/after pair."

That's true today — I traced it: `add_media_thumbnails()` fires it at `class-rtgodam-transcoder-handler.php:741`, and `class-rtgodam-transcoder-rest-routes.php` wraps the call chain at 311-316. Same for `rtgodam_before_thumbnail_store` and `rtgodam_before_transcoded_media_store`.

The problem is that `rtgodam_transcoded_thumbnails_added` is a documented extension point. Any third-party plugin, WP-CLI command, or future GoDAM code path that fires it runs these eight `get_post_meta`/`update_post_meta` calls on an attachment ID completely unwrapped — and because the justification is a provenance claim rather than a structural one, nothing will ever flag it again.

Since nesting is safe for a stack-based listener (`switch_to_blog()` keeps `$_wp_switched_stack`), firing the pair defensively inside each listener is cheaper than maintaining this invariant by hand.

inc/classes/class-media-usage-tracker.php
```diff
@@ -344,26 +344,52 @@ public function register_media_usage( $attachment_id, $post_id, $source = 'conte
...
+		 * Fires before reading/writing the attachment-keyed usage index, so
+		 * integrations that centralize media on another site can switch
+		 * context first. $post_id belongs to whatever site is currently
+		 * active and must stay untouched, so the switch is scoped tightly to
+		 * the attachment-meta calls below and is always restored before
+		 * get_post_type( $post_id ) or schedule_log_media_view() run.
+		 *
+		 * @since 2.2.0
+		 */
+		do_action( 'rtgodam_before_attachment_lookup' );
```
KMchaudhary, line 372: **Re-entrant firing: worth documenting the contract, and worth counting the switches.**

This bracket wraps `get_usage_sources()` (364) and `save_usage_sources()` (380), each of which fires its **own** pair. Ten such nesting sites across the PR: here and 427/429, `class-dynamic-gallery.php` 156/163, `class-meta-rest-fields.php` 148/149, `class-dynamic-shortcode.php` 89/103-104, `class-transcription.php` 155/164, `godam-player.php` 239/242-243, `godam-video-gallery.php` 389/396.

**Contract.** A `switch_to_blog()`/`restore_current_blog()` listener handles this fine — WP keeps `$_wp_switched_stack`. But the obvious naive implementation ("before → switch if not already switched; after → restore") unwinds to the wrong site on the first inner `after()` and silently corrupts every read after it. Since GoDAM is the side *defining* these hooks, it'd be worth stating the guarantee in the docblocks: *may be fired re-entrantly; listeners must balance with a stack.*

**Cost.** Each firing is a real round-trip that resets `$wpdb`'s prefixes, re-primes the object-cache prefix and fires `switch_blog` (which plenty of plugins hook to re-init):
- `godam-player.php:239` wraps six helpers that each fire their own pair → **7** switch/restore cycles per player render.
- `godam-video-gallery.php` calls the bracketed `rtgodam_gallery_v2_get_video_data()` once per item (396, 409) → **50** round-trips for a 50-video gallery, nested inside the outer bracket in query mode.
- Five `before()` sites sit directly inside loops: `class-retranscode-failed-media.php` 123/156, `class-video-metadata.php` 165, `class-video-migration.php` 688/814.

Hoisting the bracket to the loop/batch and exposing unwrapped `*_after_lookup` variants — the idiom already used for `create_virtual_attachment()` — would collapse most of these.

KMchaudhary:
### Review notes that don't sit on a changed line

The balance side of this PR holds up well — I ran three independent detectors over the head (`61a4c49`) and all came back clean: **zero** unguarded early exits (`return`/`throw`/`exit`/`wp_die`/`wp_send_json*`) while a bracket is open, zero unguarded `continue`/`break`, and zero top-level scopes ending with an open `before()`. The `try { … } finally { … }` idiom is applied consistently, and the multi-exit templates that can't use it close on every path. Both checkers pass (0 uncovered, 92 transitive, 124 dismissed).

Three items below are on files/lines outside the diff, so they can't be inline comments.

---

#### 1. `inc/classes/rest-api/class-media-library.php` — `assign_images_to_folder()` has no hook pair at all

`grep attachment_lookup` over lines 780-880 returns **0**, yet this REST route resolves the attachment's post row three times:

- `_prime_post_caches( $attachment_ids )` — line 810
- `get_post_type( $attachment_id )` — line 814
- `current_user_can( 'edit_post', $attachment_id )` — line 818

On a centralized install line 814 returns `''` and the loop bails immediately with `WP_Error 'invalid_attachment'` (400), so assigning images to a media folder fails for **every** user, admins included. Line 818 would fail the same way one line later.

#### 2. `inc/classes/class-pages.php:425` — video editor page meta-cap check unwrapped

```php
if ( $attachment_id && ! current_user_can( 'edit_post', $attachment_id ) ) {
```

`map_meta_cap()` calls `get_post( $args[0] )` and returns `do_not_allow` when it's null, so every visit to the video editor with `?id=<centralized attachment>` renders *"Sorry, you are not allowed to edit this item."* — for all users. `inc/classes/rest-api/class-transcription.php:133` wraps exactly this check, so the omission looks like an oversight rather than a decision.

#### 3. `inc/templates/video-preview.php:21` — file got no hooks

```php
$godam_media_mime = $godam_video_id ? (string) get_post_mime_type( $godam_video_id ) : '';
```

Zero `attachment_lookup` occurrences in the file. `$godam_is_image` / `$godam_is_audio` (22-23) both go false, so every image and audio preview renders with the "Video Preview" label and the video-only Analytics link. Its sibling `inc/templates/video-embed.php` *was* wrapped in this PR for the equivalent read.

---

### The common root cause

All three above, plus the inline notes on `class-media-library-ajax.php` (mime gates at 303 / 1260 / 1426, writes at 1045 / 1050), share one cause: **`coverage.php`'s `ACCESS_FUNCTIONS` omits the functions actually being used.** Missing today and in use on attachment IDs:

| function | why it matters |
|---|---|
| `get_post_mime_type` | reads `post_mime_type` off the attachment row; used as the pre-bracket gate in 4+ places |
| `current_user_can` (meta caps) | `map_meta_cap` → `get_post()`; fails closed, locking users out |
| `get_post_type` | same row read as `get_post()` |
| `update_attached_file` | write counterpart of the tracked `get_attached_file` |
| `wp_update_post` | writes the attachment post row |
| `get_the_title` / `get_post_field` | read the attachment row |
| `_prime_post_caches` | primes the attachment post cache |

Adding these to `ACCESS_FUNCTIONS` in #2072 and re-running would surface all of the above mechanically instead of by hand. That matters because a green run currently reads as stronger evidence than it is.

---

### Retracting one finding from my earlier pass

I previously flagged `inc/classes/rest-api/class-transcription.php:133` (`current_user_can( 'edit_post', $attachment_id )` inside the bracket) as resolving authorization against the wrong site. **That was wrong — the code is correct as written.**

`switch_to_blog()` (`wp-includes/ms-blogs.php`) only sets `$_wp_switched_stack`, `$wpdb->set_blog_id()`, `$table_prefix`, `$GLOBALS['blog_id']` and `wp_cache_switch_to_blog()`, then fires `switch_blog`. It does **not** call `WP_User::for_blog()` or re-initialise `WP_Roles`, so capability resolution stays on the original site across a switch. Meanwhile `map_meta_cap( 'edit_post', … )` genuinely needs the attachment row from the media site. So that wrap gets both halves right, and the unbracketed `current_user_can( 'upload_files' )` at line 122 is also right because it takes no post argument.

The real defects are the *inverse* cases — items 1 and 2 above, and `admin/godam-transcoder-actions.php:174` — where the same check was left unwrapped.

---

### One release-process note

`godam.php:6` and `readme.txt:7` still read `2.1.1` (as does `RTGODAM_VERSION` at `godam.php:46`) while this PR adds 128 docblocks tagged `@since 2.2.0`. Those files aren't in this diff, so this is presumably intentional sequencing — just worth confirming the release is cut as 2.2.0 and not 2.1.2, or the tags will all be off by one.

admin/godam-transcoder-actions.php
KMchaudhary: @Pathan-Amaankhan take a look into this suggetion

KMchaudhary (approved):

PR #2081, opened 2026-08-17
Fix: Media Library Delete folder issue

rudrakshi-gupta:
## Root cause

The folder tree renders from state.FolderReducer.folders. When you delete, the deleteFolder reducer removes the folder from that array — but the tree re-syncs from the cached folders query via setTree, and setTree only ever merges (add/update); it never removes. So the deleted folder gets re-added from the still-cached query data. It only disappears on a hard refresh, because then the slice starts empty and setTree takes its "replace" branch. (The merge exists on purpose — it's how the "Load more" pagination appends pages — which is why the bug was subtle.)

## The fix (3 coordinated changes, all JS)

- redux/slice/folders.js — setTree now supports removal. An array payload keeps the legacy merge/append (used by drag-reorder and attachment-count updates, which pass the full list). A new { folders, replace } payload replaces the list, so deletions actually stick.
- components/folder-tree/FolderTree.jsx — the sync effect now replaces on page 1 (reflecting deletes) and appends on load-more pages, and only syncs once the fetch has settled (!isFetching) so an in-flight/stale response can't momentarily re-add a just-deleted folder.
- redux/api/folders.js — cache invalidation. Tagged getFolders with 'Folder'; createFolder, deleteFolder, and bulkDeleteFolders now invalidatesTags: ['Folder'], so after any add/remove the tree refetches authoritative data instead of drifting from the cache.

Net effect: delete a folder → it's removed optimistically and the refetched authoritative list replaces the tree, so it disappears immediately and stays gone (single and multi-select).

## Testing instruction

On `http://<site>/wp-admin/upload.php` after a `npm run build:js`:

- Create a folder, delete it (single) → it disappears immediately, no refresh.
- Multi-select a few folders, delete → all disappear immediately.
- Delete a folder with subfolders → the whole subtree goes.
- If you have >20 folders: "Load more", then delete → the deleted one goes and the loaded pages stay.

## Demo

https://github.com/user-attachments/assets/494499ae-2948-4b0d-8d87-7807bb4744eb

### DIFF
--- a/pages/media-library/components/folder-tree/FolderTree.jsx
+++ b/pages/media-library/components/folder-tree/FolderTree.jsx
@@ -67,10 +67,17 @@ const FolderTree = ( { handleContextMenu } ) => {
 	const [ updateFolderMutation ] = useUpdateFolderMutation();
 
 	useEffect( () => {
-		if ( folders ) {
-			dispatch( setTree( openLocalStorageItem( folders?.data ) ) );
+		// Only sync once the query has settled — syncing an in-flight (stale) response
+		// could momentarily re-introduce a just-deleted folder. On the first page we
+		// REPLACE the list so server-side removals (deletes) are reflected; on load-more
+		// pages we append, preserving the already-loaded pages.
+		if ( folders && ! isFetching ) {
+			dispatch( setTree( {
+				folders: openLocalStorageItem( folders?.data ),
+				replace: currentPage === 1,
+			} ) );
 
-			if ( Array.isArray( folders?.data ) && ! isFetching ) {
+			if ( Array.isArray( folders?.data ) ) {
 				// If no folders are returned, reset to the first page
 				dispatch( updatePage( { totalPages: folders.totalPages } ) );
 			}
--- a/pages/media-library/components/modal/DeleteModal.jsx
+++ b/pages/media-library/components/modal/DeleteModal.jsx
@@ -56,17 +56,21 @@ const DeleteModal = () => {
 			// folder from the tree even though it still exists on the server. unwrap()
 			// re-throws the error payload so failures reach the catch (matching how
 			// Rename/Create already handle their mutations).
+			let partialMessage = null;
+
 			if ( ! isMultiSelecting ) {
 				await deleteFolderMutation( selectedFolder.id ).unwrap();
 			} else if ( multiSelectedFolderIds && multiSelectedFolderIds.length ) {
 				// The bulk endpoint returns HTTP 200 with `errors` on *partial* failure, so
-				// .unwrap() alone won't throw — inspect the payload and surface it as a
-				// failure rather than reporting "deleted successfully" and dropping every
-				// selected folder from the tree.
+				// .unwrap() alone won't throw. On a partial failure we still remove the whole
+				// selection optimistically and surface a WARNING (not "deleted successfully");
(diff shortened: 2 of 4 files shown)


KMchaudhary:
### Code review

The root-cause analysis is right, and the two mechanisms (tag invalidation + a `replace` branch in `setTree`) do fix the reported delete bug on page 1. Four issues with the `replace` path, all reproducible:

---

**1. A newly created folder can disappear from the tree — `redux/api/folders.js:158`**

`createFolder` now `invalidatesTags: [ 'Folder' ]`, so the page‑1 refetch replaces the whole list and drops the optimistic entry pushed by `FolderCreationModal`.

Repro: site with >20 top-level folders (`get_media_folders` paginates top-level terms 20/page, `orderby name ASC`), user on page 1. Create a folder named `zzz` → snackbar says "Folder created successfully", `createFolder` reducer pushes it, then the invalidation refetch lands and `replace: currentPage === 1` sets `state.folders` to page 1's 20 folders only. The new folder is gone until Load More / reload. Before this PR there was no refetch, so the optimistic entry survived.

Suggestion: keep the created folder by merging it back in (or skip `invalidatesTags` on `createFolder`, since the optimistic push + `updatePage` already covers it).

---

**2. The Z‑A sort order is silently reset — `components/folder-tree/FolderTree.jsx:77`**

`setSortOrder` sorts `state.folders` in place and is never re-applied afterwards. With the new replace path, deleting or creating any folder while on page 1 replaces the list with the server's `name ASC` order — the tree flips back to A‑Z while the dropdown (`App.js:189`) still reads "By Name (Z‑A)". The old merge branch preserved the order because it only `Object.assign`ed in place.

Suggestion: re-apply `state.sortOrder` at the end of the replace branch in `setTree`.

---

**3. Load-more pages are dropped when the page resets to 1 — `components/folder-tree/FolderTree.jsx:75`**

`resetUIState` sets `page.current = 1` but deliberately preserves `folders` ("for performance and user experience"). In the Edit Post media modal: Load More twice (state holds pages 1‑3, `currentPage = 3`), close the modal → `resetUIState`, reopen → cached page‑1 data resolves with `isFetching === false`, effect fires with `replace = true`, and `state.folders` collapses to page 1. Recoverable via Load More, but it defeats what `resetUIState` explicitly tries to keep.

---

**4. Deletions still aren't reconciled once the user has clicked Load More — `components/folder-tree/FolderTree.jsx:77`**

Since `replace` is gated on `currentPage === 1`, the original class of bug survives on paginated views wherever the optimistic reducer doesn't run.

Repro: `currentPage = 2`, bulk-delete 3 folders where the server deletes 2 and returns HTTP 200 with `errors` for the third. `DeleteModal` throws on `bulkResult?.errors?.length`, so `deleteFolder()` is never dispatched and all 3 remain in the tree. The invalidation refetch of `getFolders({ page: 2 })` then takes the merge branch, which can only add/update — the 2 folders that really were deleted linger until a hard refresh.

Suggestion: instead of `replace: currentPage === 1`, reconcile by page — replace the slice belonging to the fetched page (or prune ids that the authoritative response no longer contains for pages already loaded).

---

Nothing blocking on the API changes themselves — `providesTags` / `invalidatesTags` are correctly scoped, and leaving `updateFolder` un-invalidated is the right call so drag-reorder isn't clobbered.

KMchaudhary (approved):

PR #2083, opened 2026-08-17
feat(tools): let Retranscode Media fetch audio, PDF and image attachments

KMchaudhary:
This pull request adds support for selecting the media type to fetch and retranscode in the retranscoding tool, both in the REST API and the UI. Users can now choose which type of media (video, audio, PDF, image, or all) to fetch for retranscoding, and the backend and frontend have been updated to support this selection. Additionally, UI improvements make the options clearer and more accessible.

**REST API enhancements:**
- Added a `media_type` parameter to the `not-transcoded` REST API route, with validation and a default value, allowing clients to specify which type of media to fetch for retranscoding. [[1]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837R26-R54) [[2]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837R136-R145)
- Updated backend logic to use the selected media type for querying and counting attachments, ensuring accurate results and compatibility with the new parameter. [[1]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837R395-R403) [[2]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837L365-R414) [[3]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837L387-R436) [[4]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837L417-R471) [[5]](diffhunk://#diff-e31d1f786c9ba0b8d050fe0f21b0d057c41e530c240ad36008d78d8404871837L438-R495)

**Frontend/UI improvements:**
- Added a media type selector (`SelectControl`) to the retranscoding tab, allowing users to choose the type of media to fetch. The UI now dynamically shows/hides fetch options based on context. [[1]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R14) [[2]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R24-R48) [[3]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R58) [[4]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R68-R73) [[5]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83L391-R465)
- Updated the logic for fetching media to include the selected media type in API requests, and improved handling of responses and user notifications based on the selected or returned media type. [[1]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R168-R175) [[2]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R184-R204) [[3]](diffhunk://#diff-127e7446d128bc2c4ed75e0b2af4d1152fa3d977361ad7c2a1b1e385e3745d83R308)
- Improved the styling of the retranscoding options section for better usability and clarity.

These changes make the retranscoding tool more flexible and user-friendly, allowing for targeted retranscoding operations and a better overall experience.

## Demo

https://github.com/user-attachments/assets/8d23a7e0-6366-4d86-970d-c12443a9d210


subodhr258:
## Adversarial review

I ran this PR through several independent review passes, each trying to break a different part, then a second pass that tried to disprove every issue found. Below are the ones that held up, most important first. At the end I list what I checked and found solid, so you know what I looked at.

Nothing here is a merge blocker on correctness of the happy path. The main asks are the `@since` fix and some tests.

### Worth fixing

**1. `@since 2.1.2` should be `@since n.e.x.t`**
`inc/classes/rest-api/class-transcoding.php` lines 29, 51, 70, 567, 629.
The five new symbols (`DEFAULT_MEDIA_TYPE`, `get_media_type_mime_map`, `get_document_mime_types`, `get_document_attachments`, `count_transcoded_attachments`) hardcode `@since 2.1.2`. The plugin is on 2.1.1 and the next release number is not settled, so 2.1.2 is a guess. The rest of this feature's new code in `inc/helpers/custom-functions.php` already uses the `@since n.e.x.t` placeholder that the release tooling stamps with the real number at release time. Change all five to `@since n.e.x.t`.

**2. No tests added for the new server logic**
The PR changes two files and no test file. The untested parts are the ones most likely to drift: the media-type to MIME list, the document vs audio/video/image split, removing the `trash` key before summing `wp_count_attachments()`, the eligible vs not-yet-transcoded document scan in `get_document_attachments()`, `count_transcoded_attachments()`, and the new "unsupported document" skip in `retranscode_media()`. `tests/php/DocumentSupportTest.php` only exercises the helper functions in `custom-functions.php`, not this route. A PHPUnit test asserting `total_media_count` / `transcode_count` / `retranscode_count` for each media type (including `document`, where the audio/video/image list is empty, and `all`) would lock the counting math down.

**3. An extra database query runs on every fetch and its result is thrown away**
`inc/classes/rest-api/class-transcoding.php:475`.
On every fetch of video/audio/image/all, `count_transcoded_attachments()` runs a `WP_Query` so the response can carry `transcode_count` and `retranscode_count`. But `fetchRetranscodeMedia` in `RetranscodeTab.jsx` reads only `data`, `total_media_count`, `media_type`, `storage_exceeded`, and `message` from this response. It never reads `transcode_count` / `retranscode_count` here (those come from the separate `check-transcoded-status` route on the selected-IDs path). So this query is wasted work on every fetch. Either drop it and the two unused response fields, or wire the counts into the UI if you meant to show them.

**4. Unconvertible documents are skipped but not counted in the final summary**
`pages/tools/components/tabs/RetranscodeTab.jsx:291`, reacting to `class-transcoding.php:849`.
When someone bulk-selects document-type files whose extension cannot be converted (for example a `.srt`, `.c`, or `.h` stored as `text/plain`), the server correctly returns `200 {skipped:true, reason:'unsupported_document'}`. But `startRetranscoding` only tallies virtual media and successes, so this reason increments no counter. If a whole batch is skipped this way, the final notice reads "Operation completed without any media files to retranscode," which contradicts the earlier "N will be transcoded" message and the per-file skip lines in the log. The per-file log is still correct and nothing is mis-sent, so this is a wrong-summary problem, not a data problem. Suggest counting `unsupported_document` (and ideally the other skip reasons) into a "skipped" tally and naming it in the summary.

### Optional / your call

**5. On Copilot's `application/ogg` point**
I checked this closely and it is narrower than it looks. On a normal WordPress upload a `.ogg` or `.oga` audio file is stored with `post_mime_type` = `audio/ogg` (WordPress sets the stored type from the file extension, not from what libmagic reports), and `audio/ogg` is matched by both the `audio` and `all` options. So the miss only affects attachments explicitly typed `application/ogg`, which normal uploads do not produce but a migration or a custom mime filter could. Since the transcoder itself lists `application/ogg` as supported, adding `'application/ogg'` to the `audio` and `all` entries in `get_media_type_mime_map()` makes the selector match the transcoder exactly. Cheap and safe, but low priority.

**6. Pre-existing, not introduced by this PR: a network error mid-batch can freeze the tool**
`pages/tools/components/tabs/RetranscodeTab.jsx:300`.
The catch block does `const data = err.response.data;` without checking `err.response` first. If a retranscode POST fails with no HTTP response (server restart, connection drop), this throws inside the catch, the loop's finishing setters never run, and the progress state gets stuck (the Abort button still recovers it, so no reload is strictly needed). This line is not in your diff, but since you are already editing this file, guarding `err.response` the way `fetchRetranscodeMedia` already does on line 233 is cheap hardening.

### Checked and found solid
- The trash double-count Copilot raised is already handled: the current code removes the `trash` key before summing at `class-transcoding.php:472`.
- `media_type` is validated by a fixed allowed-value list plus `rest_validate_request_arg`, so unknown values are rejected with 400 rather than silently defaulting.
- Audio and image attachments do get `rtgodam_transcoded_url` set on completion, so the "not yet transcoded" check works correctly for them, not just for video.
- The `document` case guards against an empty MIME filter, which would otherwise pull the whole media library.
- The JavaScript `MEDIA_TYPE_OPTIONS` values match the PHP allowed-value list.
- The storage-limit early return still returns an accurate `total_media_count` (the document scan that feeds it runs before the check on purpose).

subodhr258 (approved):

PR #2084, opened 2026-08-19
Fix/godam blockspacing in columns

KMchaudhary:
Partially Fixes: https://github.com/rtCamp/godam-core/issues/1341

This pull request refactors how player wrapper styles are loaded for the GoDAM player. Previously, these styles were output as an inline `<style>` tag at render time, which caused layout issues in some contexts. Now, the wrapper styles are shipped as a real stylesheet and registered as a dependency of the main player CSS, ensuring consistent loading across all render paths and eliminating the phantom spacing bug.

**Player Wrapper Styles Refactor:**

* The inline `<style>` output for player wrapper styles in `godam-player.php` has been removed and replaced with logic to enqueue the `godam-player-wrapper-style` stylesheet if it is registered. This prevents layout issues caused by the inline style tag and ensures styles are loaded consistently.
* The `register_scripts()` method in `class-godam-player.php` now registers `godam-player-wrapper-style` as a real stylesheet and makes it a dependency of `godam-player-style`, so it is automatically enqueued wherever the player is used.

**Documentation and Comments:**

* Updated comments in `godam_skin_styles_enqueue()` to reflect that wrapper styles are now enqueued as a stylesheet, not injected inline.
* Updated comments in `class-wpforms-field-godam-video.php` to clarify that while most player variants now use the stylesheet, the whitelist for inline `<style>` remains for legacy or special cases (e.g., audio no-JS block).

## Demo

Screenshot 2026-08-19 at 11 32 42 AM

- [x] Self tested on Elementor's containers as well.

### DIFF
--- a/inc/classes/shortcodes/class-godam-player.php
+++ b/inc/classes/shortcodes/class-godam-player.php
@@ -156,10 +156,31 @@ class GoDAM_Player {
 			);
 		}
 
+		/**
+		 * Player wrapper styles (placeholder / poster / loading states).
+		 *
+		 * Shipped as a real stylesheet rather than an inline <style> emitted by
+		 * inc/templates/godam-player.php: the style tag rendered as a sibling of
+		 * the player markup, so in a flow-layout container (core/column, core/group)
+		 * the player became the second child and picked up the container's
+		 * `margin-block-start` blockGap rule, adding phantom spacing above the video.
+		 * The CSS is static — nothing per-video — so there is nothing to inline.
+		 *
+		 * Registered as a dependency of `godam-player-style` so every existing
+		 * enqueue point (block.json `style`, shortcodes, Elementor `depended_styles`,
+		 * gallery) pulls it in with no further changes.
+		 */
+		wp_register_style(
+			'godam-player-wrapper-style',
+			RTGODAM_URL . 'assets/build/css/godam-player-wrapper.css',
+			array(),
+			filemtime( RTGODAM_PATH . 'assets/build/css/godam-player-wrapper.css' )
+		);
+
 		wp_register_style(
 			'godam-player-style',
 			RTGODAM_URL . 'assets/build/css/godam-player.css',
-			array(),
+			array( 'godam-player-wrapper-style' ),
 			filemtime( RTGODAM_PATH . 'assets/build/css/godam-player.css' )
 		);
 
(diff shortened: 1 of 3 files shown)


subodhr258:
Reviewed the full diff (all PHP plus the player and media-library JS). Overall this is solid and I did not find any blocking issues.

What stood out as well done:

- The #465 work is consistent: translation-triggering calls are moved off `plugins_loaded` to `init` / `admin_init`, add-on notice text is passed as callables and built at render time, and the FluentForms/WPForms field loaders now guard on `class_exists()` before touching the parent class.
- Folder locking now has a real server-side guard (`enforce_locked_media_folder` on the native `wp/v2/media-folder` routes, plus the IDOR/ownership checks in `assign_images_to_folder`), so a direct REST call can no longer mutate a locked folder. The regex matches because the `media-folder` taxonomy registers with the default rest base.
- The lightbox / deep-link handling is careful about history (single entry, replace vs push), focus trapping, and the iOS custom-fullscreen lock cleanup on close.
- Lazy-loading Video.js in the media library and gating the heavy bundles per screen are good payload wins.

One non-blocking note:

**Analytics keepalive chunking measures characters, not bytes** — `assets/src/js/godam-player/analytics.js:459`

`eventBytes = JSON.stringify( event ).length + 1` counts UTF-16 code units, but `MAX_BYTES_PER_REQUEST = 32 * 1024` is meant to keep each chunk under the ~64 KiB keepalive body budget measured in bytes. For multibyte layer names or metadata (non-Latin text, emoji) the real UTF-8 size is 2 to 4x the character count, so a chunk that looks under 32 KB can be well over it in bytes. When that happens the keepalive POST is rejected, `.catch(()=>{})` swallows it, and the buffer is cleared unconditionally right after, so those events are lost. The 2x headroom (32 KB cap against a 64 KiB limit) covers ASCII and most Latin text, so this only bites on multibyte-heavy payloads. If you want it exact: `new TextEncoder().encode( JSON.stringify( event ) ).length`.

**Squash subject** — the PR title covers only one of the eight commits. Since these squash-merge, the merge subject will understate what is actually a 2.1.1 batch (translation fixes, lightbox, media-library fixes, ZIP hardening, new CI workflows). Worth a more descriptive subject at merge time. The description itself is thorough.

subodhr258:
Re-reviewed against `develop` (the base changed from `main`, per your DM). The diff is now just the wrapper-CSS blockspacing fix across 3 files, which is what the title describes. My earlier comment on this PR was against the old `main` base and no longer applies, so please ignore it.

The change is clean and correct. The player-wrapper CSS moves from an inline `<style>` (which rendered as a sibling of the player markup and so picked up the flow container's `margin-block-start` blockGap) to a registered `godam-player-wrapper-style`, set as a dependency of `godam-player-style`. Every enqueue point (block.json `style`, the shortcode's `wp_enqueue_style( 'godam-player-style' )`, Elementor `depended_styles`, gallery) pulls it in through that dependency, and the template keeps a direct-`require` fallback. The build artifact `assets/build/css/godam-player-wrapper.css` is present.

One thing worth a quick sanity check: the removed inline block was documented as deliberately inline because `wp_head` / `wp_footer` do not fire in some render paths (Elementor editor preview, Elementor Pro REST widget refresh, custom REST renders). Since the wrapper is now a dependency of `godam-player-style`, which already reaches those paths the same enqueue way and renders fine, this should be covered. Confirming the wrapper CSS still applies on the Elementor editor canvas closes the loop.

LGTM.

subodhr258 (approved):

PR #2087, opened 2026-08-21
Single-currency Revenue KPI card + base_currency proxy

subodhr258:
## What

The dashboard side of single store currency revenue (pairs with rtCamp/godam-analytics#257). GoDAM commits to one store currency: no conversion, no reliance on multi-currency plugins, no external FX.

- **WP proxy (`class-analytics.php`)**: pass `base_currency = get_option('woocommerce_currency')` to `/dashboard/metrics/fetch/` and `/dashboard/top-products/`. The service then returns base-currency revenue plus a count of orders in other currencies. The `revenue` object already forwards through the dashboard-metrics `array_merge`.
- **`RevenueCard.js`**: a WooCommerce-only Insights KPI card. Shows total video-attributed revenue via the existing `formatRevenue(minor, currency)` (so JPY/KWD fraction digits are correct), with an "excluding N orders in other currencies" sub-line only when that count is > 0. A single-currency store shows just the figure. Renders nothing when the payload is absent, so it never shows a misleading 0.
- **`Dashboard.js`**: render `RevenueCard` in the Insights row, gated on `hasWooProducts`.

## Display rule

Every revenue figure is one number in the store's base currency; orders in other currencies are excluded from the figure and disclosed once as the sub-line (Joel's format: "Revenue: $467 (excluding 33 orders in other currencies)"). No per-currency list anywhere.

## Base branch

Targets `feat/top-products` (where #2086 shipped the Top Products revenue column + `formatRevenue`), not `develop`. Nothing merges to `develop` until Phase 2 is complete.

Part of rtCamp/godam-plugin-wp#26.

### DIFF
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -851,6 +851,14 @@ class Analytics extends Base {
 				'api_key'       => $api_key,
 			)
 		);
+		// Single store currency: pass the store base currency so the service returns
+		// base-currency revenue plus a count of orders in other currencies (not
+		// converted). Only meaningful when WooCommerce is active; when absent the
+		// service simply omits the `revenue` object.
+		$base_currency = get_option( 'woocommerce_currency', '' );
+		if ( ! empty( $base_currency ) ) {
+			$params['base_currency'] = $base_currency;
+		}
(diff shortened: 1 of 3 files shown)


subodhr258:

PR #2088, opened 2026-08-21
Show per-hotspot Direct revenue in the layer panel

subodhr258:
## What

Shows per-hotspot Direct revenue in the video layer-analytics panel (Sub-task C core; pairs with rtCamp/godam-analytics#259).

- **`useVideoLayerData`**: maps `revenue_minor` / `orders` / `currency` onto each sub-hotspot from the endpoint row; the parent carries the sum of its hotspots' revenue (currency from any child).
- **`SubHotspotRail`**: a small secondary line per hotspot, "&lt;amount&gt; · N orders", shown only when that hotspot has orders (Woo layers only), so the rail stays scannable.
- **`LayerDetailPanel`**: a headline "This hotspot/layer drove &lt;amount&gt; across N orders" above the funnel, with an InfoTooltip stating it is the **Direct in-video contribution only**, not the product's total (Assisted purchases via the product page are excluded). Reuses the shipped `formatRevenue` (correct ISO fraction digits, full numbers).
- **WP proxy**: passes `base_currency = get_option('woocommerce_currency')` to `/processed-layer-analytics/`.

## Single store currency

Per-hotspot revenue sums base-currency orders only; other currencies are excluded server-side (counted on the dashboard, never converted here). Direct-only because a hotspot carries a `layer_id`; Assisted purchases have none.

## Base branch

Targets `feat/top-products` (#2086, where `formatRevenue` shipped). Nothing merges to `develop` until Phase 2 is complete.

Part of rtCamp/godam-plugin-wp#26.

### DIFF
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -408,6 +408,13 @@ class Analytics extends Base {
 		if ( ! empty( $job_id ) ) {
 			$query_params['job_id'] = $job_id;
 		}
+		// Single store currency: the service returns per-hotspot Direct revenue in
+		// this currency (Woo layers only); other currencies are excluded, not
+		// converted. Empty when WooCommerce is inactive, so the service omits it.
+		$base_currency = get_option( 'woocommerce_currency', '' );
+		if ( ! empty( $base_currency ) ) {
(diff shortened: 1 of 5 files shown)


subodhr258:

PR #2090, opened 2026-08-24
feat(analytics): Woo Phase 2 revenue UI (single-currency card, per-hotspot, influenced)

subodhr258:
## What

The **single godam (core) secondary PR** for Woo Phase 2 — the dashboard/UI side. Base: `feat/top-products`. Parent: rtCamp/godam-plugin-wp#26. Supersedes the closed per-sub-task PRs #2087/#2088/#2089. Pairs with godam-analytics#262 and godam-for-woo#197.

## What's in it

- **Revenue KPI card (A/B)** — `RevenueCard` on the dashboard Insights row (single store currency, "excluding N orders" sub-line), reused on the **single-video Analytics page** for per-video revenue. `class-analytics.php` forwards `base_currency` to all four proxied endpoints.
- **Per-hotspot Direct revenue (C)** — the layer panel shows each Woo hotspot's revenue; `useVideoLayerData` maps it onto subs and sums to the parent.
- **Influenced revenue (D)** — a separate "Influenced" sub-line in the Top Products Revenue cell, reusing `formatRevenue`.
- **No-data empty state (F FIX 2)** — the Revenue cell shows "No data" (not a bare dash / misleading 0) when the service sends null revenue.
- **Per-placement split (F EASY WIN A)** — a compact "Woo Layer X · Reel Pop Y" breakdown under the Revenue cell, shown only when 2+ placements drove revenue.
- **Comparative tips (F EASY WIN B)** — `RevenueTips` on the per-video page renders only the favourable comparisons (video-vs-store AOV, revenue per viewer).

All money rendering reuses the single shipped `formatRevenue(minor, currency)` — no new formatter.

## Video-to-Purchase + Funnels (Phase 2, added)
- **`PlacementFunnelCard`** — the **Funnel by placement** dashboard card: a Play→Cart→Purchase funnel per placement (Shoppable Video block / Woo hotspot layer / Reel Pop) with each placement's reach, Played/Added/Purchased counts and % of players, and its play-to-purchase rate. Fetches its own data via a new `fetchPlacementFunnels` RTK query (heavier per-placement queries, so they don't block the main dashboard), proxied through a new `placement-funnels` REST route.

- **`VideoToPurchaseCard`** — mirrors `VideoToCartCard`: the count of distinct viewers who played then bought an attributed product, plus `% of viewers`, with the purple Woo badge and the same null-vs-zero guard. Rendered Woo-gated in the per-video **and** dashboard Insights rows (reads the new `video_to_purchase` key). The Insights row now wraps into a grid on large screens (`lg:flex-wrap` + `lg:[&>*]:basis-40`) so the extra card wraps instead of cramping the previously 4-card cap.
- **`PurchaseFunnelCard`** — the **Purchase Funnel** card (the play-to-cart-to-purchase flow): left-aligned bars in a grey track, the Added-to-cart bar split into **Direct** (added in-video, dark blue) + **Assisted** (clicked out then added, light blue), the count + % of players on the right, drop-off annotations ("X% advanced", "N did not add to cart", a red "N abandoned after adding" pill), and a legend. Rendered **account-wide on the Dashboard** and **per-video** on the analytics page (a `scope` prop switches the descriptor). Keeps the "still counting" note.

## Design polish (matching the finalized ticket design)

- **Detailed Revenue card** — `RevenueCard` rebuilt to match the finalized design: headline total, a Direct/Assisted split bar with per-tier amounts, and the account-level **Influenced** figure in its own dashed box (reported separately, never added to the Direct + Assisted total). The per-video card shows the same split without Influenced (a product-page concept). Moved to its own full-width row above the funnel.
- **Trend badges** — a shared `MetricTrend` footer shows the period-over-period % change (green ↗ / red ↘ + "vs prev N days") on Video-to-Cart, Video-to-Purchase and the Revenue card whenever a bounded range is picked; all-time shows the range label instead. The Insights range now drives every card there (VtC/VtP/Revenue/funnels), not just the three KPI tiles.
- **Top Products i-icons** — info tooltips on the Source, Product Views, Add to Cart and Revenue column headers, each explaining the metric in plain words.
- **Per-product Direct/Assisted revenue** — a "X direct · Y assisted" sub-line under each product's revenue (from the new `revenue_direct_minor` / `revenue_assisted_minor` keys), shown only when the service sent the split; the two amounts are added to the CSV export.
- **Copy + layout** — Video-to-Cart/Purchase subtitles read "carts/orders from video, N% of viewers who played"; the funnel card is titled "Purchase Funnel"; VtC and VtP sit together on their own row below the KPI tiles.

## Tests

**Full core jest suite passes.** New/updated: `RevenueCard.test.js`, `MetricTrend.test.js`, `RevenueTips.test.js`, `TopProductsTable.test.js` (revenue, influenced, No-data, placement split, Direct/Assisted tier split), `useVideoLayerData.test.js` (per-hotspot mapping), `VideoToCartCard.test.js` / `VideoToPurchaseCard.test.js` (subtitle + trend), `PurchaseFunnelCard.test.js`. Each sub-task was revert-proofed on its original branch. phpcs: 0 errors on the changed PHP.

## Depends on

godam-analytics#262 to populate the revenue / order / influenced / placement / tips fields; degrades cleanly against an older service build (fields absent → no card / no sub-line).

## Notes

`@since n.e.x.t`; no CHANGELOG/POT during development. New `data-test-id`s on the revenue surfaces for E2E.

## Image-block hotspots (added)

- **Capture.** A small standalone `godam-layer-analytics` bundle (`assets/src/js/godam-player/layer-analytics-runtime.js` + entry) registers `window.GoDAM.addLayerInteraction` and the page-hide flush WITHOUT the video player, since image pages never load `godam-player-analytics`. `render.php` enqueues it when the image has layers and stamps `data-block-source="godam-image"` on the frame; the flush element lookup now also finds `.godam-image__frame` by `data-id`, so the surface tag survives the flush. `video_id` already carries the image's attachment id via `getVideoKey`, so ingest accepts the event unchanged. Folding `analytics.js` onto the shared runtime to drop the remaining duplication is a follow-up.
- **Source labels.** Top Products Source chips now read **Video Woo Layer** (`woo-layer`) and **Image Woo Layer** (`godam-image`) so the two media read apart.

Verified on the dev site: a synthetic image `type=3` event (no play) rolled up as `godam-image` and rendered as the "Image Woo Layer" chip. The bundle is built and served and registers the buffer API; the live in-image interaction path is unit-tested but not yet exercised on a real page (no authored image-hotspot content exists on the dev site).

## Review fixes (adversarial review)

- **Correct funnel share.** The Purchased row's "% of players" showed purchased ÷ carts (the cart→purchase conversion — a large overstatement that contradicted its own bar); it now shows purchased ÷ players, matching the bar and the server stage rate. purchased/carts stays only as the drop-row "advanced" figure.
- **Funnel can't render inverted.** With the backend now guaranteeing purchased ≤ added ≤ played, bar fractions clamp to [0,1] and the "advanced" % clamps to 100 as a defensive guard against a deploy skew.
- **Honest labels.** The Purchased stage drops the false "net of refunds" descriptor (revenue/counts are gross by design) and is labelled "of those who added" (the nested cohort). The layer panel shows the parent layer's revenue **without** an order total — summing per-hotspot distinct counts overcounts an order spanning two hotspots; a single hotspot still shows its exact distinct count.

## Final adversarial review fixes (image-hotspot pass)

- **Image impressions were always 0**: a still image fires no visibility transition, so the parent-layer `viewed` beacon the product rollup reads was never emitted. `render-image-frame.js` now calls `emitLayerVisible()` once per Woo layer on render, and `render.php` makes the layer-analytics runtime a dependency of the renderer so `window.GoDAM` exists before it fires.
- **block_source dropped on mixed pages**: on a page with both a video and an image, `analytics.js`'s flush wins and its `findVideoElementById` had no image fallback, so image events flushed `block_source=''`. Added the `.godam-image__frame[data-id]` fallback there too, matching the standalone runtime.

## Review fixes (elifvish review on godam-plugin-wp#26)

- **Revenue readable without auth (#2):** `/fetch`, `/dashboard-metrics`, `/layer-analytics` and the new `/placement-funnels` were `__return_true` and now return revenue. Gated all four on `current_user_can( 'upload_files' )` (matching `/top-products`/`/top-videos`); only the admin apps call them, never the front-end player. Verified on the dev site: anonymous → `401`, admin dashboard unaffected.

### DIFF
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -639,6 +639,15 @@ class Analytics extends Base {
 		);
 		$query_params = $this->append_range_params( $request, $query_params );
 
+		// Single store currency: pass the store base currency so the per-video
+		// record carries base-currency revenue (and a count of orders in other
+		// currencies). Only when WooCommerce is active; otherwise the service
+		// returns revenue 0 / '' and the card stays hidden.
+		$base_currency = get_option( 'woocommerce_currency', '' );
+		if ( ! empty( $base_currency ) ) {
+			$query_params['base_currency'] = $base_currency;
+		}
+
 		$analytics_url = add_query_arg( $query_params, $analytics_endpoint );
 
 		// Send request to analytics microservice.
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -851,6 +851,14 @@ class Analytics extends Base {
 				'api_key'       => $api_key,
 			)
 		);
+		// Single store currency: pass the store base currency so the service returns
+		// base-currency revenue plus a count of orders in other currencies (not
+		// converted). Only meaningful when WooCommerce is active; when absent the
+		// service simply omits the `revenue` object.
+		$base_currency = get_option( 'woocommerce_currency', '' );
+		if ( ! empty( $base_currency ) ) {
+			$params['base_currency'] = $base_currency;
+		}
 		$endpoint = add_query_arg(
 			$params,
 			RTGODAM_ANALYTICS_BASE . '/dashboard/metrics/fetch/'
@@ -1217,6 +1225,11 @@ class Analytics extends Base {
 		}
 		if ( ! empty( $order ) ) {
 			$query['order'] = $order;
+		}
+		// Single store currency: revenue/orders sum only the base currency.
+		$base_currency = get_option( 'woocommerce_currency', '' );
+		if ( ! empty( $base_currency ) ) {
+			$query['base_currency'] = $base_currency;
 		}
 
 		$endpoint = add_query_arg(
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -38,7 +38,12 @@ class Analytics extends Base {
 				'args'      => array(
 					'methods'             => WP_REST_Server::READABLE,
 					'callback'            => array( $this, 'fetch_analytics_data' ),
-					'permission_callback' => '__return_true', // Publicly accessible.
+					// Admin-dashboard read; gated like top-products/top-videos: Phase 2 added
+					// revenue to this response (it was public for view/play analytics only).
+					// Only the admin apps call it, never the public front end.
+					'permission_callback' => function () {
+						return current_user_can( 'upload_files' );
+					},
 					'args'                => array(
 						'video_id' => array(
 							'required'          => true,
@@ -90,7 +95,11 @@ class Analytics extends Base {
 				'args'      => array(
 					'methods'             => WP_REST_Server::READABLE,
 					'callback'            => array( $this, 'fetch_dashboard_metrics' ),
-					'permission_callback' => '__return_true',
+					// Admin-dashboard read; gated like the sibling analytics routes (revenue is
+					// account-scoped and returned to the dashboard only).
+					'permission_callback' => function () {
+						return current_user_can( 'upload_files' );
+					},
 					'args'                => array(
 						'site_url' => array(
 							'required'          => true,
@@ -234,7 +243,10 @@ class Analytics extends Base {
 					'callback'            => array( $this, 'fetch_placement_funnels' ),
 					// Account-scoped server-side (api_key / account_token injected,
 					// never client-supplied), like the sibling dashboard reads.
-					'permission_callback' => '__return_true',
+					// Admin-dashboard read; gated like the sibling analytics routes.
+					'permission_callback' => function () {
+						return current_user_can( 'upload_files' );
+					},
 					'args'                => array(
 						'site_url' => array(
 							'required'          => true,
@@ -256,7 +268,11 @@ class Analytics extends Base {
 					// return this site's own non-PII aggregate analytics. Locking down
 					// analytics reads, if ever wanted, should be done uniformly across
 					// all analytics routes rather than just this one.
-					'permission_callback' => '__return_true',
+					// Admin-dashboard read; gated uniformly with the sibling analytics routes
+					// (see note above): revenue must not be readable without auth.
+					'permission_callback' => function () {
+						return current_user_can( 'upload_files' );
+					},
 					'args'                => array(
 						'video_id'   => array(
 							'required'          => true,
@@ -514,10 +530,10 @@ class Analytics extends Base {
 	 * first 100 rows as a defensive bound on per-request DB work; rows past the
 	 * cap still get a constant-cost attributable label so none render blank.
 	 *
-	 * The /analytics/fetch route is public (permission_callback __return_true),
-	 * so this never leaks non-public pages (private, draft, pending, trashed) to
-	 * anonymous callers: the real title/permalink is revealed only when the page
-	 * is publicly viewable, or the current user can edit it.
+	 * The analytics read routes are gated on `upload_files` (authors and above),
+	 * but this still never leaks non-public pages (private, draft, pending,
+	 * trashed) beyond what the caller may see: the real title/permalink is
+	 * revealed only when the page is publicly viewable, or the caller can edit it.
 	 *
 	 * @param array $placements Placement rows from the microservice.
 	 * @return array Enriched placement rows.
--- a/inc/classes/rest-api/class-analytics.php
+++ b/inc/classes/rest-api/class-analytics.php
@@ -220,6 +220,24 @@ class Analytics extends Base {
 			),
 			array(
 				'namespace' => $this->namespace,
+				'route'     => '/' . $this->rest_base . '/placement-funnels',
+				'args'      => array(
+					'methods'             => WP_REST_Server::READABLE,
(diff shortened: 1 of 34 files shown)


inc/classes/rest-api/class-analytics.php
```diff
@@ -226,6 +226,24 @@ public function get_rest_routes() {
...
 			),
+			array(
+				'namespace' => $this->namespace,
+				'route'     => '/' . $this->rest_base . '/placement-funnels',
+				'args'      => array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'fetch_placement_funnels' ),
+					// Account-scoped server-side (api_key / account_token injected,
+					// never client-supplied), like the sibling dashboard reads.
+					'permission_callback' => '__return_true',
```
elifvish: The comment justifies `__return_true` on the grounds that the route is "account-scoped server-side (api_key / account_token injected, never client-supplied)". That is a correct argument that **another store** cannot read this store's data. It is not an argument that **an anonymous visitor to this store** should be able to.

And `/top-products`, added in the sibling PR #2074, gates on `current_user_can( 'upload_files' )` — so the two new routes in the same feature disagree with each other about who may read commerce analytics.

The bigger point is what Phase 2 puts on the routes that were already public. Comparing against `origin/main`, `/fetch`, `/history`, `/dashboard-metrics`, `/dashboard-history` and `/layer-analytics` were all `__return_true` before this work — but they returned view counts and watch time. This PR adds `base_currency` to `fetch_layer_analytics`, `fetch_analytics_data` and `fetch_dashboard_metrics`, and `fetch_dashboard_metrics` returns the service payload through `array_merge( $empty_metrics, $body['dashboard_metrics'] ?? array() )`, i.e. verbatim. After this merges, an unauthenticated:

```
GET /wp-json/godam/v1/analytics/dashboard-metrics?site_url=https://thestore.com
```

returns the store's gross video-attributed revenue, order count, Video-to-Cart and Video-to-Purchase; `/layer-analytics` returns per-hotspot revenue; `/fetch` returns per-video revenue. That is a competitor's revenue dashboard, free, no login.

The route permissions were not made worse by this PR — the *sensitivity of what flows through them* was, which is the moment to fix them. Suggest gating all four revenue-bearing routes on `current_user_can( 'upload_files' )` to match `/top-products`, and if any of them genuinely has an anonymous front-end caller, splitting the revenue keys out of that response rather than leaving the whole payload open.

This one is worth a decision from @subodhr258 before merge — it is the single highest-impact item in the review.

elifvish:
## Review verdict: approve

Re-reviewed at `139c4082` after the permission fix.

**F2 is fixed, and I verified it rather than taking the commit message at its word:**

- Six routes now gate on `upload_files`; `/history` and `/dashboard-history` stay public and I confirmed both carry zero revenue references, so the split is correct.
- Every caller of the four newly-gated routes lives under `pages/` (dashboard, analytics, video-editor React apps). Nothing under `assets/src` — the front-end player — touches any of them, so there's no visitor-facing regression.
- `upload_files` matches the capability on the Analytics and Media Editor submenus exactly. The only `edit_posts` submenus are Help and What's New, which call nothing. **No legitimate role gets locked out.**

Updating the `fetch_placements` docblock in the same commit was the right call — that comment would have been actively misleading otherwise.

**One caveat, not specific to this PR.** This PR adds fourteen `.test.js` files and the JS test job never runs on it: the workflow is gated on `branches: [develop, main]` and this targets `feat/top-products`. Adding `'feat/**'` to that filter is a one-line change and would let the stack verify itself. Worth doing before merge — after merge, a failure becomes someone else's problem to discover.

Remaining findings against this file tree are low severity and mostly belong to #2074 (the ISO 4217 exponent table duplicated across repos, and the money formatters living in a table component). Neither blocks.

Lowest-risk PR of the three in this phase. Approving.

subodhr258:
Thanks. The CI caveat is addressed.

**JS tests on `feat/**`: added (`ae002b5a`).** `js_unit_tests_on_pull_request.yml` now includes `feat/**`, so the fourteen new `.test.js` files run on this PR and the rest of the stack; you can see the run on this PR. The same one-line change is available for the PHP test workflow if you want it, I kept it to the JS one you flagged to avoid pulling in the known-red Plugin Check noise.

Confirmed the permission split is exactly as you verified: the six revenue-bearing routes gate on `upload_files`, and `/history` / `/dashboard-history` stay public and carry no revenue keys.

inc/classes/rest-api/class-analytics.php
subodhr258: Fixed in this PR: `/fetch`, `/dashboard-metrics`, `/layer-analytics` and `/placement-funnels` now gate on `current_user_can('upload_files')`, matching `/top-products`. Only the admin dashboard/analytics/editor apps under `pages/` call them; `/history` and `/dashboard-history` stay public and carry no revenue keys. Verified anon returns 401 and the admin dashboard is unaffected.

elifvish (approved):

PR #2093, opened 2026-08-26
Fix Document block download link for virtual media

KMchaudhary:
This pull request improves the handling of document download URLs for virtual media attachments, ensuring users receive working links in all cases. It also adds comprehensive tests to cover these scenarios and introduces a stub for the uploads directory in the test suite.

**Improvements to virtual media download URL resolution:**

* [`inc/helpers/custom-functions.php`](diffhunk://#diff-f29b4cd63502a994cef6926d1eab320b3b83465c4d16dd6920e3d7162bd61f27L1024-R1079): Updated `rtgodam_get_document_download_url` to check for virtual media attachments that would otherwise return a broken local URL, and to prefer the `rtgodam_transcoded_url` (CDN link) when appropriate. Also added `rtgodam_before_attachment_lookup` and `rtgodam_after_attachment_lookup` hooks around attachment access for better integration support.
* [`inc/helpers/custom-functions.php`](diffhunk://#diff-f29b4cd63502a994cef6926d1eab320b3b83465c4d16dd6920e3d7162bd61f27L1011-R1015): Clarified documentation in `rtgodam_get_document_preview_url` about the edge case where WordPress generates an unserved uploads path for virtual media.

**Test coverage improvements:**

* [`tests/php/DocumentSupportTest.php`](diffhunk://#diff-22e64c575dbb46ed13907379375490e5a56c47e22535ab6b578c1debf7d2e9ecR614-R691): Added new tests to verify that virtual media attachments reliably return CDN URLs, that resolved CDN URLs are not overridden, and that ordinary local uploads are left untouched.

**Test infrastructure:**

* [`tests/bootstrap.php`](diffhunk://#diff-96a82592013e5f4b91e682332583b7a4a6ca1ff0431d267c61179b1a21b167daR107-R120): Added a stub implementation of `wp_get_upload_dir` to provide a testable uploads base URL in the test environment.

### DIFF
--- a/inc/helpers/custom-functions.php
+++ b/inc/helpers/custom-functions.php
@@ -1008,7 +1008,11 @@ function rtgodam_get_document_preview_url( $attachment_id = 0, $fallback_src = '
  *
  * wp_get_attachment_url() already resolves virtual GoDAM media to its CDN URL via
  * Media_Library_Ajax::filter_attachment_url_for_virtual_media(), so local and GoDAM-tab
- * attachments are both handled here.
+ * attachments are both handled here — but see the note in the body on the one case where
+ * that resolution silently produces a path nothing serves.
+ *
+ * Attachment access is bracketed by rtgodam_before_attachment_lookup /
+ * rtgodam_after_attachment_lookup, as everywhere else that reads attachment data.
  *
  * @since n.e.x.t
  *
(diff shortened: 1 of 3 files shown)


subodhr258:
Reviewed this. The core fix is sound: for virtual GoDAM media whose post guid is empty, `wp_get_attachment_url()` hands back an uploads path with no file behind it, and preferring `rtgodam_transcoded_url` in that case is the right answer. Returning the transcoded URL is also correct for download specifically, since for a document that meta holds the original file, not the preview PDF. Tests are thorough and the meta-key stubbing is right. A few things worth a look before merge.

### 1. The internal `rtgodam_before/after_attachment_lookup` firing is redundant on the only path that reaches this

`render.php` already wraps both `rtgodam_get_document_preview_url()` and `rtgodam_get_document_download_url()` in exactly these hooks (try/finally), and it is the sole caller. The inline comment this PR removed said as much.

So on the block path the hooks now fire three times, nested: render's bracket, this new one, and again inside `wp_get_attachment_url()` via `filter_attachment_url_for_virtual_media()`. For a handler that does something not nesting-safe (anything other than the stack-based `switch_to_blog` / `restore_current_blog`), that extra pair can double-apply. It also leaves the two resolvers asymmetric: download self-brackets, preview does not.

Either push the bracket into both resolvers and drop it from `render.php`, or leave both to the caller as before. Right now it is half-migrated.

### 2. The docblock says "as everywhere else that reads attachment data", but the sibling right above does not bracket

`rtgodam_get_document_preview_url()` reads `get_post_meta` and `wp_get_attachment_url` and is not internally bracketed. The claim is not accurate as written. Same root as #1.

### 3. `$godam_is_local` is a prefix match, which assumes `wp_get_attachment_url()` and `wp_get_upload_dir()` agree on scheme and host

In the common path they do (core builds the attachment URL from the same `wp_get_upload_dir()` base). But a third-party filter on `wp_get_attachment_url` that rewrites the scheme or host (some SSL/CDN plugins do exactly this) makes the prefix fail, `$godam_is_local` go false, and the broken local URL get returned, i.e. the very bug this fixes, silently unfixed on those sites. Worth a note on the assumption, or a scheme-normalized compare.

### 4. One branch is untested and silently returns the 404ing URL

Virtual media + local-looking URL + no `rtgodam_transcoded_url` meta falls through to `return $attachment_url` (the dead local path). It is the unavoidable fallback, but there is no test for it and it is the one case the fix cannot rescue. A test pinning the current behavior would lock it in.

### 5. `@since 2.2.0` on the new `do_action` docblock

The function itself keeps `@since n.e.x.t` and the plugin is at 2.1.1. The concrete `2.2.0` here does match the existing hook docs in `class-media-library-ajax.php` and `class-rtgodam-retranscodemedia.php`, so it is defensible, just flagging the mismatch with the `n.e.x.t` placeholder on the same function.

Nothing here blocks the core behavior. #1 is the one I would most want resolved.

subodhr258 (approved):

PR #2097, opened 2026-08-27
feat(dashboard): GA4 connection widget

mayan-000:
## What this is

The dashboard-facing half of GA4 output: a small widget confirming the GA4 connection is live and showing running counts of what's been sent, so a merchant doesn't have to take it on faith. The actual `dataLayer` pushes happen in `godam-for-woo` ([PR #198](https://github.com/rtCamp/godam-for-woo/pull/198)), which this widget reads from.

## Features

- **`GA4ConnectionWidget`** — gated on the existing `enable_gtm_tracking` setting (`window.godamSettings.enableGTMTracking`, read directly, no new setting added), with **five render states**, not two:
  - **Not connected** (the toggle is off): a short prompt + an Enable link to the existing settings tab.
  - **Loading**: the counts request is in flight.
  - **Status unavailable**: the counts request failed or errored (e.g. a version mismatch with `godam-for-woo`, a permissions issue) — shown distinctly so a broken connection never silently reads as a working one at zero.
  - **Standing down**: the toggle is on, but `godam-for-woo`'s capability probe reports another GA4 integration is active (`source_active: true` on the REST response) — GoDAM is correctly pushing nothing. Names the other source (`source_type`) where known, and relabels the counts "prepared" rather than implying they were sent.
  - **Connected**: the toggle is on and GoDAM is the active source — running `add_to_cart` / `purchase` counts (full numbers, no abbreviation, explicitly labelled "All time" since they're lifetime totals, not scoped to the dashboard's date-range picker), a Manage link back to the settings tab, and an info tooltip explaining these are GA4's own default ecommerce events.
- New RTK Query endpoint hitting `godam-for-woo`'s `GET /ga4-counts` REST route (now returning `source_active`/`source_type` alongside the two counts — the field names this widget's `transformResponse` reads exactly), called whenever the toggle is on.
- Rendered on the dashboard next to the existing Video-to-Cart card, gated the same way the other Woo-only surfaces are (add-on active + valid license); the Insights row's wrap breakpoint was adjusted so five cards no longer crowd each other near 1024px.

## User journey this is part of

**A merchant opens their GoDAM dashboard to answer one question: "is this actually working, and how much is it doing?"** — on day one with the setting off, a week later after turning it on, on a store that already has a different GA4 integration installed, or when something's actually broken. This widget is the merchant's only visibility into that, so every one of its five states needs to stay honest about what's actually happening server-side — which is exactly what the review process behind the sibling PR spent most of its effort verifying, and is the reason this went from two states to five.

image
image

### DIFF
--- a/pages/dashboard/Dashboard.js
+++ b/pages/dashboard/Dashboard.js
@@ -395,7 +395,13 @@ const Dashboard = () => {
 									testIdPrefix="godam-dashboard-insights-daterange"
 								/>
 							</div>
-							<div className="analytics-info-container single-metrics-info-container flex max-lg:flex-row items-stretch flex-wrap justify-center lg:flex-nowrap">
+							{ /* `lg:flex-nowrap` was sized for four cards (three metric tiles +
+							    Video-to-Cart); with the Woo add-on active, GA4ConnectionWidget
+							    makes it five, and forcing all five into one nowrap row right at
+							    the 1024px breakpoint squeezed each card too narrow to hold its
+							    labels + Manage link comfortably. Pushing the nowrap breakpoint to
+							    `xl` lets the row wrap onto two lines between lg and xl instead. */ }
+							<div className="analytics-info-container single-metrics-info-container flex max-lg:flex-row items-stretch flex-wrap justify-center xl:flex-nowrap">
 
 								<SingleMetrics
 									mode="dashboard"
--- a/pages/dashboard/Dashboard.js
+++ b/pages/dashboard/Dashboard.js
@@ -25,6 +25,7 @@ import ViewersGauge from './components/ViewersGauge';
 import PlaybackPerformanceDashboard from '../analytics/PlaybackPerformance';
 import TopVideosTable from './components/TopVideosTable';
 import TopProductsTable from './components/TopProductsTable';
+import GA4ConnectionWidget from './components/GA4ConnectionWidget';
 import DateRangePicker, { triggerLabelFor, fromISO } from '../analytics/components/DateRangePicker';
 
 /**
@@ -455,6 +456,12 @@ const Dashboard = () => {
 										dataLabel={ insightsCardLabel }
 									/>
 								) }
+
+								{ /* GA4 output is a godam-for-woo feature (it pushes the
+								    add_to_cart/purchase events into the store's own
+								    dataLayer), so it's gated the same as Top Products /
+								    Video-to-Cart: the add-on active + a valid license. */ }
+								{ hasWooProducts && <GA4ConnectionWidget /> }
 							</div>
 						</div>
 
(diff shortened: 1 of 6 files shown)


subodhr258:
## Adversarial review (issue #26, GA4 connection widget)

The widget itself is clean: two states, correct gating on the add-on and license, and good test coverage. One cross-cutting correctness point.

### Widget shows "Sending to GA4" while GoDAM is standing down
`pages/dashboard/components/GA4ConnectionWidget.js:41` (and the label at L95)

`isConnected` is just `enable_gtm_tracking`. It has no view of the source probe's stand-down state. On a store that also runs GTM4WP or the official Google Analytics for WooCommerce, GoDAM stands down and pushes nothing, but the server counter still counts ("prepared to send"), so this widget shows "Sending to GA4" with rising counts. That label is not true for that store, and the merchant is led to believe GoDAM is delivering events it never sends.

The woo side already localizes `godamWooGa4.ga4SourceActive` and `ga4SourceType`. Consider surfacing that here: a distinct "another GA4 source is active, GoDAM is standing down" state, or a softer label like "Prepared N events" that matches what the count actually means.

Everything else looks good.

---

## Round 2 (deep pass, multi-agent adversarial review)

A wider second pass surfaced two more widget issues (both verified against the code), plus four on the sibling [godam-for-woo#198](https://github.com/rtCamp/godam-for-woo/pull/198).

### Counts ignore the dashboard date range and are shown unlabeled
`pages/dashboard/components/GA4ConnectionWidget.js:43` (rendered in `Dashboard.js`)

The counts come from `get_counts()`, which reads lifetime option totals with no date filter, and the widget takes no range props. It sits in the same insights row as the range-scoped cards (Video-to-Cart and the KPI tiles). A merchant who has set the range to "last 7 days" reads these GA4 numbers as 7-day figures when they are all-time. Either scope the counts to the active range or label them "all time".

### Any REST error still renders "Sending to GA4" with 0 / 0
`pages/dashboard/components/GA4ConnectionWidget.js:43`

The component reads only `data` from the query, never `isError` or `isLoading`. If `godam-for-woo/v1/ga4-counts` is missing (godam core updated ahead of godam-for-woo, the exact version skew the ticket's release sequencing warns about) or returns 403, `data` is undefined and the widget shows "Sending to GA4" with 0 / 0 instead of surfacing the error. Handle the error and loading states so a broken connection does not read as a working one at zero.

Both of these, plus the stand-down label point above, come down to the widget's connected state being a single client flag with no view of what the server side is actually doing.


---

## Round 3 (live testing, via browser MCP)

On a real store the widget itself renders correctly through all its states, but during testing it reported a connection that was not real, through no fault of this PR. The sibling add-on never actually pushes to the dataLayer in the default configuration (see [godam-for-woo#198](https://github.com/rtCamp/godam-for-woo/pull/198): `ga4SourceActive` reaches the browser as the string `"0"`, and `shouldPush()` treats `"0"` as truthy, so it always stands down). During the test this widget showed "Sending to GA4, 3 add_to_cart" while zero events reached `window.dataLayer`.

The widget's own logic is fine: the REST `source_active` boolean is not affected by that stringification, so the widget correctly chose the "Sending to GA4" branch. Flagging it here only so it is clear that the "Sending to GA4" state is currently untrue until #198's push bug is fixed. Worth re-verifying this widget end to end (including the standing-down state) once that lands.

subodhr258 (approved):

PR #2098, opened 2026-08-27
Hotfix: develop main sync

KMchaudhary:
This pull request introduces several improvements and bug fixes across the PDF viewer and media library components. The most significant changes focus on making PDF rendering more robust under strict Content Security Policies (CSP), improving error handling for dynamic imports, and ensuring proper event unbinding in Backbone views to prevent memory leaks.

**PDF Viewer Robustness and Fallbacks:**

* Added protocol validation for `previewUrl` in `renderNativeFallback` to prevent unsafe URLs from being rendered in the PDF object (`assets/src/blocks/godam-pdf/view.js`).
* Enhanced the PDF.js worker fallback logic to explicitly set `GlobalWorkerOptions.workerSrc` when the worker fails to start, ensuring main-thread rendering works even under strict CSPs (`assets/src/blocks/godam-pdf/viewer/worker.js`). [[1]](diffhunk://#diff-c19f08fcc9783cd3102fbfa93fa52e90d335df430bb4bac42e98b6d333eee27eL35-R38) [[2]](diffhunk://#diff-c19f08fcc9783cd3102fbfa93fa52e90d335df430bb4bac42e98b6d333eee27eR49-R58)

**Media Library Improvements:**

* Updated event binding in the attachments browser to use Backbone's `listenTo` instead of direct `.on()` calls, ensuring event handlers are properly cleaned up when the view is removed and preventing memory leaks (`assets/src/js/media-library/views/attachment-browser.js`).

**Dynamic Import Error Handling:**

* Improved error handling in the `loadVideoJs` function to clear the cached promise if the dynamic import fails, allowing retries on subsequent calls instead of permanently failing until a page reload (`assets/src/js/media-library/videojs-loader.js`).

### DIFF
--- a/assets/src/blocks/godam-pdf/view.js
+++ b/assets/src/blocks/godam-pdf/view.js
@@ -23,14 +23,18 @@ import { createRoot, createElement } from '@wordpress/element';
  * @param {string}      previewUrl URL of the PDF to display.
  */
 function renderNativeFallback( wrapper, previewUrl ) {
+	if ( ! previewUrl || ! isSafeUrl( previewUrl ) ) {
+		return;
+	}
+
 	const object = document.createElement( 'object' );
 
 	object.type = 'application/pdf';
 	object.width = '100%';
 	object.height = '100%';
-	// Assigned as-is. The URL has already been through esc_url() server-side, and encodeURI()
-	// would escape its percent signs a second time — turning a legitimate "%20" into "%2520"
-	// and pointing the object at a path that does not exist.
+
+	// Assigned as-is after protocol validation. encodeURI() would escape existing percent
+	// signs a second time — turning a legitimate "%20" into "%2520" and breaking the URL.
 	object.setAttribute( 'data', previewUrl );
 
 	// Keep whatever fallback markup render.php emitted as the <object>'s own fallback, so a
--- a/assets/src/blocks/godam-pdf/viewer/index.js
+++ b/assets/src/blocks/godam-pdf/viewer/index.js
@@ -17,7 +17,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { pdfWorker } from './worker';
+import { pdfWorker, pdfWorkerReady } from './worker';
 
 /*
  * The viewer's styles — ours plus react-pdf's TextLayer/AnnotationLayer sheets — are
@@ -129,6 +129,24 @@ export default function DocumentViewer( {
 	const currentUrlRef = useRef( url );
 	const [ width, setWidth ] = useState( 0 );
 	const [ ownerDocument, setOwnerDocument ] = useState( null );
+	const [ workerReady, setWorkerReady ] = useState( false );
+
+	// Hold document loading until the pdf.js worker — or, under a worker-blocking CSP, its
+	// main-thread fallback — is in place. Without this the first getDocument() could run in the
+	// gap before the fallback handler is registered and error out. See viewer/worker.js.
+	useEffect( () => {
+		let active = true;
+
+		pdfWorkerReady.then( () => {
+			if ( active ) {
+				setWorkerReady( true );
+			}
+		} );
+
+		return () => {
+			active = false;
+		};
+	}, [] );
 
 	// Pages are rendered to a canvas at a fixed pixel width, so they have to be re-rendered
 	// when the container resizes — a CSS-scaled canvas just goes blurry.
@@ -336,7 +354,7 @@ export default function DocumentViewer( {
 			     starting earlier would register the fonts against globalThis.document, and the
 			     reload triggered by the corrected `options` would leave those stale rules
 			     behind in the wrong document. */ }
-			{ ! ownerDocument ? loadingView : (
+			{ ! ownerDocument || ! workerReady ? loadingView : (
 				<Document
 					file={ url }
 					options={ options }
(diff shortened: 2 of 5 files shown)


subodhr258:
## Review

Three of the four changes are clean and ready. The PDF worker fallback in `worker.js` needs another pass, because it does not do what its comment says it does. One note on the PR title at the end.

### `viewer/worker.js` — the workerSrc fallback does not render on the main thread

This is the same spot Copilot flagged, and the concern is real. I traced it through pdfjs-dist 4.8.69 to be sure. Two separate things are happening here:

**What the change genuinely fixes (keep this):** when the worker is blocked and `pdfWorker` is `null`, react-pdf builds its own `new PDFWorker()`. Its `_initialize()` reads `let { workerSrc } = PDFWorker` *before* the try/catch, and that getter throws `No "GlobalWorkerOptions.workerSrc" specified.` synchronously when workerSrc is unset. A synchronous throw there is not routed to `<Document>`'s `onError`, so it takes the block down (the same class of crash the file's top comment already describes for the `fromPort` case). Setting `workerSrc` to any non-empty value avoids that synchronous throw. That part is a real improvement.

**What it does not do (the comment is inaccurate):** it does not get pdf.js to parse on the main thread. pdf.js's fake-worker path is:

```js
const worker = await import(/*webpackIgnore:true*/ this.workerSrc);
return worker.WorkerMessageHandler;   // reads the module namespace export
```

It reads `WorkerMessageHandler` off the imported module's namespace. But `pdf-worker.js` is a side-effect-only entry:

```js
import './promise-with-resolvers';
import 'pdfjs-dist/build/pdf.worker.min.mjs';   // no re-export
```

Its namespace exports nothing, so `worker.WorkerMessageHandler` is `undefined`, `WorkerMessageHandler.setup(...)` throws, and `_setupFakeWorker` rejects `_readyCapability`. (On top of that, Copilot is right that the standalone `new URL('./pdf-worker.js', import.meta.url)` is not guaranteed to resolve to the same importable-ESM chunk that the `new Worker(new URL(...))` form emits, so the `import()` may not even load.)

**Net effect:** under a worker-blocking CSP the fake-worker setup rejects asynchronously, react-pdf routes that to `onError`, and `view.js` degrades to its native `<object>` fallback: the browser's built-in PDF viewer where it has one, and the server-rendered download link otherwise (the `<object>` keeps that link as its own fallback content). So there is no crash and no blank dead-end, but pdf.js never parses on the main thread. Copilot's "still fails instead of rendering" is a touch strong for browsers that ship a PDF viewer, but the main-thread pdf.js path the comment promises never runs.

So please pick one:

- **If the native `<object>` fallback is acceptable** (simplest): keep the one-liner, but rewrite the comment to say what actually happens. Something like: setting any non-empty `workerSrc` converts pdf.js's fatal *synchronous* `No workerSrc specified` throw into an *async* rejection that react-pdf surfaces via `onError`, so `view.js` degrades to the native `<object>` viewer. Drop the "parses on the main thread / renders, just more slowly" wording.
- **If main-thread pdf.js rendering is the goal:** populate the global handler in the catch instead. Importing the worker build on the main thread sets it: `pdf.worker.min.mjs` runs `globalThis.pdfjsWorker = { WorkerMessageHandler }` on evaluation. So `await import('pdfjs-dist/build/pdf.worker.min.mjs')` makes `_initialize()` take its first branch (`#mainThreadWorkerMessageHandler` truthy) and skip the synchronous `workerSrc` read entirely, and `_setupFakeWorker` then uses the real handler. It is async, so it has to resolve before the first `getDocument`.

Either way, this needs a real runtime test under a worker-blocking CSP (serve the page with `Content-Security-Policy: worker-src 'none'` and confirm the document renders with no uncaught error). The interaction between the two webpack URL forms, pdf.js 4.8.69's fake worker, and sync-throw vs async-reject is too subtle to settle by reading the diff.

### The other three look good

- **`view.js`** — the `isSafeUrl` guard is correct and is the right shape for the CodeQL alert: the sanitizer sits on the sink's own data-flow path (`object.setAttribute('data', ...)`), which is what the analyzer needs. It is redundant with the `mountViewer` gate at runtime, but that redundancy is exactly the point here, so keep it. `isSafeUrl` is a function declaration, so calling it above its definition is fine.
- **`videojs-loader.js`** — correct. Clearing `loadingPromise` on failure and re-throwing lets the next open retry instead of caching the rejection. `cachedVideojs` is only set on success, so no stale cache.
- **`attachment-browser.js`** — correct memory-leak fix. `listenTo` means `View.remove()` tears these down; the previous `.on()` handlers on the shared `collection.props` outlived the view. No orphaned `.off()` left behind.

### PR title

The title "Hotfix: develop main sync" and the `hotfix/develop-main-sync` branch read like a develop/main branch alignment, but the diff is two fix commits on top of `develop` (a CodeQL fix plus the Copilot-comment fixes). `develop` is already 19 commits ahead of `main`, and `main` does not have `worker.js` or `videojs-loader.js` at all, so nothing here is a sync. Worth retitling to describe the fixes so it is squash-merged as a normal fix PR rather than merge-committed as an alignment.

### Minor

- Consider a small unit test for `isSafeUrl` (rejects `javascript:`/`data:`, allows `http`/`https`) and for the videojs retry (a rejected import clears the cache so the next call re-imports).
- CI: PHPCS, JS unit, and PHP unit are green. The pending POT and Plugin Check are the known repo-wide ones, not from this PR.
- Please reply on Copilot's open `worker.js` thread once you decide the direction.

subodhr258 (approved):

Issue #1977, opened 2026-06-22
Gallery Block and Featured Image do not preview images with empty alt attributes in editor


PalakKansal:
## Bug Report

**Current Behavior**
When an image with an empty alt attribute is used in a Gallery Block or as a Featured Image, the image preview is not displayed in the Gutenberg editor. Instead, a placeholder message is shown.

However, the same image is previewed correctly when used in the GoDAM Image Block.

This results in inconsistent editor behavior across different image implementations..

**Expected behavior/code**
Images with empty alt attributes should have consistent editor preview behavior across Gallery Block, Featured Image, and GoDAM Image Block.

**Steps to reproduce the bug**
1. Upload an image with an empty alt attribute to the Media Library.
2. Create a new post/page.
3. Add the image to a Gallery Block.
4. Observe the editor preview.
5. Set the same image as the Featured Image.
6. Observe the editor preview.
7. Add the same image using the GoDAM Image Block.
8. Compare the editor preview behavior..

**Testing Environment**

- Browser used: Chrome
- OS: Windows 11
- GoDAM Version: v1.13.0
- WordPress Version: 7.0

https://github.com/user-attachments/assets/5e5a57aa-b4fd-4843-acc9-6fb31157f501

Asteriskkkk:
https://github.com/user-attachments/assets/456fdf61-8838-4145-8688-ef9f88ca3a32

i am unable to reproduce this am i missing something here?

kishan-gondaliya-7270:
Released

PR #2103, opened 2026-09-01
Fix: feature image virtual media

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam/issues/1977

This pull request introduces robust support for using GoDAM media items as featured images in both the block and classic editors. The main changes ensure that placeholder GoDAM IDs are properly handled and swapped for real WordPress attachment IDs once the media entry is created, preventing issues where invalid IDs could be saved or displayed. The implementation includes new logic for the classic editor, improved handling in the block editor, and comprehensive unit tests.

**Featured image handling improvements:**

* Added `classic-featured-image.js` to intercept and defer setting GoDAM IDs as featured images in the classic editor, only applying the real attachment ID once it exists, and preventing invalid IDs from being saved or displayed.
* Updated `godam-media-frame-shared.js` to invoke the new classic editor logic (`setupClassicFeaturedImage` and `resolveClassicFeaturedImage`) during GoDAM tab activation and after attachment creation. [[1]](diffhunk://#diff-a7878e67888eef229132a39357ea6378cbaf2c7cdc41b7b4bc69f4333e5fd9a4R19) [[2]](diffhunk://#diff-a7878e67888eef229132a39357ea6378cbaf2c7cdc41b7b4bc69f4333e5fd9a4R203-R205) [[3]](diffhunk://#diff-a7878e67888eef229132a39357ea6378cbaf2c7cdc41b7b4bc69f4333e5fd9a4R313-R318)

**Block editor and post entity integration:**

* Replaced the logic for updating the featured image in the block editor: added `replaceVirtualIdInFeaturedImage()` to swap the GoDAM placeholder ID for the real attachment ID on the post entity, ensuring the correct image is displayed and saved.

**Testing and reliability:**

* Added `classic-featured-image.test.js` with comprehensive tests for the new classic editor featured image deferral and resolution logic, covering edge cases and ensuring correct integration.

**Codebase simplification:**

* Removed the now-unnecessary `checkIfFeatureImage()` logic and related conditional checks, as the new approach directly handles both block and classic editor contexts. [[1]](diffhunk://#diff-a7878e67888eef229132a39357ea6378cbaf2c7cdc41b7b4bc69f4333e5fd9a4L107-R145) [[2]](diffhunk://#diff-a7878e67888eef229132a39357ea6378cbaf2c7cdc41b7b4bc69f4333e5fd9a4L153-R171)

## Demo

https://github.com/user-attachments/assets/b115b0a0-8b1d-482a-9ec8-8d0133af5797

### DIFF
--- a/assets/src/js/media-library/classic-featured-image.js
+++ b/assets/src/js/media-library/classic-featured-image.js
@@ -1,0 +1,112 @@
+/**
+ * Classic editor Featured image support for the GoDAM tab.
+ *
+ * The block editor's Featured image panel routes a pick through `editPost()`, so the
+ * placeholder GoDAM ID it stores can simply be repointed once the real attachment
+ * exists (see replaceVirtualIdInFeaturedImage in godam-media-frame-shared.js).
+ *
+ * The classic editor gives us no such window. `wp.media.featuredImage.frame()` binds
+ * `wp.media.featuredImage.select` to the state's `select` event at frame construction,
+ * so it always runs before our own handler and immediately calls
+ * `wp.media.featuredImage.set()` with whatever ID the selected model carries — the
+ * Central job docname, at that point. That POSTs to `get-post-thumbnail-html`, which
+ * casts the value with `(int)`: `8f7d2a1b` becomes `8`, so the meta box renders an
+ * unrelated attachment and writes its ID into the `_thumbnail_id` hidden field, which
+ * the next post save then persists.
+ *
+ * So instead of repairing the value afterwards, hold it back. `set()` is wrapped to
+ * park a GoDAM pick rather than send it, and resolveClassicFeaturedImage() releases the
+ * real attachment ID through the original setter once create-media-entry has answered.
+ * Nothing bogus ever reaches the meta box or the hidden field.
+ */
+
+/**
+ * The placeholder GoDAM ID held back from `wp.media.featuredImage.set()` while its real
+ * attachment is being created. Null when no pick is in flight.
+ *
+ * @type {string|number|null}
+ */
+let deferredVirtualId = null;
+
+/**
+ * Whether a `set()` call is carrying a GoDAM pick that onGoDAMSelect() is about to turn
+ * into a real attachment.
+ *
+ * Deliberately mirrors onGoDAMSelect()'s own gate — the two must agree, or a pick gets
+ * parked with nothing on the way to release it. The ID comparison is what keeps
+ * `remove()` working: it calls `set( -1 )`, which never matches the selected model, and
+ * the content mode can still read as 'godam' from an earlier session.
+ *
+ * @param {string|number} id The ID handed to `wp.media.featuredImage.set()`.
+ * @return {boolean} True when the value should be parked instead of sent.
+ */
+function isDeferrableGoDAMId( id ) {
+	const frame = window.wp?.media?.frame;
+
+	if ( 'godam' !== frame?.content?.mode?.() ) {
+		return false;
+	}
+
+	const selected = frame.state?.()?.get?.( 'selection' )?.single?.();
+
+	return !! selected && String( selected.id ) === String( id );
+}
+
+/**
+ * Wrap `wp.media.featuredImage.set()` so a GoDAM pick is parked rather than sent.
+ *
+ * Called from GoDAMCreate() rather than at page load: `wp.media.featuredImage` is
+ * defined by media-editor.js, which is not guaranteed to have run by the time this
+ * bundle initializes, and the patch is only ever needed once the GoDAM tab is rendered.
+ * Idempotent, since GoDAMCreate() runs on every tab activation.
+ *
+ * @return {void}
+ */
+export function setupClassicFeaturedImage() {
+	const featuredImage = window.wp?.media?.featuredImage;
+
+	if ( ! featuredImage || featuredImage.godamOriginalSet ) {
+		return;
+	}
+
+	const originalSet = featuredImage.set;
+
+	if ( 'function' !== typeof originalSet ) {
+		return;
+	}
+
+	// Kept for resolveClassicFeaturedImage(): it runs while the modal still reports
+	// 'godam' as its content mode, so releasing through the wrapper would park the real
+	// ID too and the pick would never land.
+	featuredImage.godamOriginalSet = originalSet;
+
+	featuredImage.set = function( id ) {
+		if ( isDeferrableGoDAMId( id ) ) {
+			deferredVirtualId = id;
+			return undefined;
+		}
+
+		deferredVirtualId = null;
+
+		return originalSet.call( this, id );
+	};
+}
+
+/**
+ * Release a parked classic-editor featured image now that its attachment exists.
+ *
+ * @param {string|number} virtualMediaId The GoDAM ID that was parked.
+ * @param {number}        realId         The newly created WP attachment ID.
+ * @return {void}
+ */
+export function resolveClassicFeaturedImage( virtualMediaId, realId ) {
+	if ( null === deferredVirtualId || String( deferredVirtualId ) !== String( virtualMediaId ) ) {
+		return;
+	}
+
+	deferredVirtualId = null;
+
+	const featuredImage = window.wp?.media?.featuredImage;
+
+	featuredImage?.godamOriginalSet?.call( featuredImage, realId );
+}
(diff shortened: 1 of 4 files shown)


assets/src/js/media-library/views/godam-media-frame-shared.js
```diff
@@ -292,6 +310,12 @@ const GoDAMMediaFrameShared = {
 				// so that inner blocks are resolved without relying on React effect timing.
 				replaceVirtualIdInCoreImageBlocks( data.id, attachment.id );
 
+				// The featured image lives on the post entity rather than on block
+				// attributes, so it needs its own swap. The classic editor gets no
+				// placeholder at all — its pick was parked, and is released here.
+				replaceVirtualIdInFeaturedImage( data.id, attachment.id );
+				resolveClassicFeaturedImage( data.id, attachment.id );
```
subodhr258, line 341: **Failed `create-media-entry` silently drops the classic featured-image pick.** `resolveClassicFeaturedImage()` only runs inside the `if ( response.success )` block. If the request fails or returns non-success, the outer `catch` at the bottom of `processGoDAMItem` swallows it, the parked id is never released, and `deferredVirtualId` stays stuck. The meta box keeps its previous value with no error surfaced, so the user believes a featured image was set when nothing changed. Consider releasing the park (or clearing it with a visible fallback) on the failure path too.

assets/src/js/media-library/views/godam-media-frame-shared.js
```diff
@@ -104,29 +105,44 @@ function replaceVirtualIdInCoreImageBlocks( virtualMediaId, realId ) {
...
-};
+	if ( currentFeaturedMedia === undefined || currentFeaturedMedia === null ) {
+		return;
+	}
+
+	if ( String( currentFeaturedMedia ) !== String( virtualMediaId ) ) {
+		return;
+	}
+
+	dispatch( 'core/editor' ).editPost( { featured_media: realId } );
```
subodhr258: **Guard this `editPost` the way the read above it is guarded.** The `getEditedPostAttribute` read on line 131 sits in a try/catch, but this `dispatch( 'core/editor' ).editPost()` does not. A throw here unwinds into `processGoDAMItem`'s swallow-all `catch`, which then skips `resolveClassicFeaturedImage()` and both the `godam-virtual-attachment-created` and `godam-attachment-browser:changed` dispatches for this item. `replaceVirtualIdInCoreImageBlocks` is self-guarded, so this is the one exposed call in that block. Wrapping it in the same defensive guard keeps a bad dispatch from taking out the events other code depends on.

assets/src/js/media-library/classic-featured-image.js
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// Kept for resolveClassicFeaturedImage(): it runs while the modal still reports
+	// 'godam' as its content mode, so releasing through the wrapper would park the real
+	// ID too and the pick would never land.
+	featuredImage.godamOriginalSet = originalSet;
+
+	featuredImage.set = function( id ) {
+		if ( isDeferrableGoDAMId( id ) ) {
+			deferredVirtualId = id;
```
subodhr258, line 162: **Save-before-resolve keeps the old featured image (classic editor).** Parking suppresses the original `set()`, so the meta box `_thumbnail_id` hidden field still holds the previous value until `create-media-entry` answers and `resolveClassicFeaturedImage()` releases the real id. If the user clicks Update/Publish inside that sub-second window, the save persists the stale id and the newly chosen image is lost until they save again. Worth confirming that timing is acceptable, or showing a brief in-flight indicator while a pick is being resolved.

assets/src/js/media-library/classic-featured-image.test.js
```diff
@@ -0,0 +1,137 @@
...
+			frame,
+			featuredImage: { set },
+		},
+	};
+
+	return { set, frame };
+};
+
+describe( 'classic editor featured image deferral', () => {
+	afterEach( () => {
```
subodhr258, line 60: **`deferredVirtualId` is module state that `afterEach` never resets.** It only deletes `window.wp` and restores mocks, so a test that parks a pick (e.g. 'parks a GoDAM pick') leaves `deferredVirtualId` set for the next test. That makes the suite order-dependent and able to pass or fail spuriously when tests are reordered or new ones are inserted before the existing ones. Drain the park in `afterEach` (call `resolveClassicFeaturedImage` to clear it, or export a small reset for tests).

assets/src/js/media-library/views/godam-media-frame-shared.js
```diff
@@ -185,6 +200,9 @@ const GoDAMMediaFrameShared = {
 	GoDAMCreate() {
 		const state = this.state();
 
+		// Hold back a classic-editor featured image pick until its attachment exists.
+		setupClassicFeaturedImage();
```
subodhr258: **Nit (altitude): this installs a global core patch on every GoDAM tab render, not just featured-image contexts.** `setupClassicFeaturedImage()` permanently wraps the core global `wp.media.featuredImage.set` the first time any GoDAM tab renders, including the Insert-Media and gallery flows where no featured image is ever picked. It is idempotent and transparent for non-deferrable ids, so not a live bug, but it widens the blast radius of a core-global override past the context the fix targets. Consider only wrapping when the frame is actually a featured-image frame.

assets/src/js/media-library/classic-featured-image.js
```diff
@@ -0,0 +1,112 @@
...
+function isDeferrableGoDAMId( id ) {
+	const frame = window.wp?.media?.frame;
+
+	if ( 'godam' !== frame?.content?.mode?.() ) {
+		return false;
+	}
+
+	const selected = frame.state?.()?.get?.( 'selection' )?.single?.();
+
+	return !! selected && String( selected.id ) === String( id );
```
subodhr258: **Nit (reuse): the `String( a ) === String( b )` id-equality idiom is repeated** across `isDeferrableGoDAMId`, `resolveClassicFeaturedImage`, `replaceVirtualIdInFeaturedImage`, and `replaceVirtualIdInCoreImageBlocks`. A shared `isSameId( a, b )` helper in the media-library utility module would keep the virtual-vs-real id comparison in one place if the matching rule ever needs to change (trimming, coercion, etc.).

subodhr258 (commented):
Review notes on the featured-image virtual-media fix. The core-frame wiring, `wp.media.frame` reset path, and the sync-editPost / async-create ordering all check out. A few things worth a look, correctness first:

subodhr258:
### Re-review

Went through the diff again end to end, and ran the new suite plus lint on the branch.

- `classic-featured-image.test.js`: 15/15 pass
- ESLint on the four changed JS files: clean

The core design holds up well. Parking the classic `set()` and releasing the real ID after `create-media-entry` answers is the right call, the block vs classic paths stay independent (the classic wrap only installs on a `featured-image` frame, and the block swap edits the post entity), and every path is keyed by ID so multi-select and slow failures do not cross-talk. Failure and clear paths are covered.

A few minor notes, all low severity, none blocking:

**1. Two quick classic picks can strand the first and leave the placeholder stuck**
`deferredVirtualId` and `metaBoxSnapshot` are single slots, so they can only track one pick at a time. In the classic editor: pick A as featured image, then reopen and pick B before A's `create-media-entry` resolves. `showInFlightPlaceholder()` then overwrites the snapshot with the "Setting featured image…" markup, and `deferredVirtualId` moves to B. When A resolves it no-ops on the ID mismatch (so A is never applied), and if B then fails, `restoreMetaBox()` puts the "Setting featured image…" text back as the "previous" image, leaving the meta box stuck on the placeholder until reload. Narrow race given the modal is single-select and closes on select, but worth a comment at minimum. No test covers a second pick landing while one is parked.

**2. Stuck placeholder if core's `set()` gets `'0'` back**
In `resolveClassicFeaturedImage()`, `metaBoxSnapshot` is nulled right before delegating to core's `set(realId)`. If `get-post-thumbnail-html` returns `'0'`, core does not repaint `#postimagediv .inside`, so the placeholder is left with no restore path. Very edge for a freshly created attachment, listing it for completeness.

**3. `isSameId` is not reused in `core-media-blocks-extension.js:56`**
`ids.js` says it exists to keep the virtual/real match rule "in a single place", but the identical `String(a) === String(b)` at `core-media-blocks-extension.js:56` was left unconverted, so the rule now lives in two spots. That file is outside this PR's touch set, so this is optional. If converted it needs the existing null/undefined allowance kept, e.g. `mediaId == null || isSameId( mediaId, virtualMediaId )`, since `isSameId` returns false for null/undefined.

Nothing here blocks merge.

subodhr258 (approved):

PR #2104, opened 2026-09-01
Fix the HTML CTA layer overflow issue

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam/issues/1532

This pull request updates the styling for the `.easydam-layer--cta-html` class in both the `godam-player/style.scss` and `godam-player.scss` files to improve layout consistency and responsiveness. The changes ensure the element now occupies the full width and height of its container and handles overflow appropriately.

**Styling improvements for `.easydam-layer--cta-html`:**

* Set `width` and `height` to `100%` to make the element fill its container, replacing the previous `fit-content` width. [[1]](diffhunk://#diff-89e92fdab13aad2dda8bfe013f74d8961d1e5fad2e34d617cd41a3d29c67c9caL275-R281) [[2]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79L210-R216)
* Added `overflow: auto` and `box-sizing: border-box` to ensure content is scrollable if it exceeds the container and that padding/borders are included in the element's total size. [[1]](diffhunk://#diff-89e92fdab13aad2dda8bfe013f74d8961d1e5fad2e34d617cd41a3d29c67c9caL275-R281) [[2]](diffhunk://#diff-219690a9913b2b11e68684985e772b86cc8396a1381054dd4b3ad3d9d818fd79L210-R216)

## Demo

https://github.com/user-attachments/assets/5c7d01df-3b41-4b9b-ac53-83e011b69d9e

### DIFF
--- a/assets/src/blocks/godam-player/style.scss
+++ b/assets/src/blocks/godam-player/style.scss
@@ -272,10 +272,13 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem; 
 	font-family: Manrope, sans-serif;
-	width: fit-content;
 	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
+	width: 100%;
+	height: 100%;
+	overflow: auto;
+	box-sizing: border-box;
 }
 
 // Only apply default styles to plain buttons without any classes
--- a/assets/src/blocks/godam-player/style.scss
+++ b/assets/src/blocks/godam-player/style.scss
@@ -272,7 +272,6 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem; 
 	font-family: Manrope, sans-serif;
-	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
 	width: 100%;
--- a/assets/src/css/godam-player.scss
+++ b/assets/src/css/godam-player.scss
@@ -207,10 +207,13 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem;
 	font-family: Manrope, sans-serif;
-	width: fit-content;
 	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
+	width: 100%;
+	height: 100%;
+	overflow: auto;
+	box-sizing: border-box;
 }
 
 .poll-container button:hover {
--- a/assets/src/css/godam-player.scss
+++ b/assets/src/css/godam-player.scss
@@ -207,7 +207,6 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem;
 	font-family: Manrope, sans-serif;
-	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
 	width: 100%;


assets/src/blocks/godam-player/style.scss
```diff
@@ -272,10 +272,13 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem; 
 	font-family: Manrope, sans-serif;
-	width: fit-content;
 	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
+	width: 100%;
+	height: 100%;
```
subodhr258, line 278: **Fixed `width: 100%; height: 100%` drops short CTAs from centered to top-left.** The parent `.easydam-layer` is `display:flex; justify-content:center; align-items:center`, so the old `fit-content` box was centered. With `100%/100%` the box now fills the whole overlay, so the flex centering no longer applies to the content and a short CTA (e.g. a single "Buy now" button) renders in the top-left corner.

I verified in a browser repro of these exact rules: content center moves from **(0.50, 0.51)** before to **(0.11, 0.10)** after, while tall content still scrolls in-bounds.

The sibling `.easydam-layer--cta-text` right above solves the same overflow with `max-height: 100%` rather than a fixed height, which keeps short content centered *and* caps/scrolls tall content. Suggest matching it here:

```scss
.easydam-layer--cta-html {
	font-size: 1.375rem;
	font-family: Manrope, sans-serif;
	margin: 0 auto;
	padding: 10px 20px;
	border-radius: 4px;
	width: fit-content;
	max-width: 100%;
	max-height: 100%;
	overflow: auto;
	box-sizing: border-box;
}
```

That fixed both the short-CTA centering and the tall-CTA overflow in the repro (box no longer fills the layer; content back at (0.50, 0.51)).

assets/src/blocks/godam-player/style.scss
```diff
@@ -272,10 +272,13 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem; 
 	font-family: Manrope, sans-serif;
-	width: fit-content;
 	margin: 0 auto;
```
subodhr258: Nit: `margin: 0 auto` is dead once the element is `width: 100%` (auto side margins only center a box narrower than its container). It was doing the horizontal centering for the old `fit-content` box. Remove it, or it becomes meaningful again if you adopt the `width: fit-content; max-width: 100%` variant above. Same line exists in `assets/src/css/godam-player.scss`.

assets/src/css/godam-player.scss
```diff
@@ -207,10 +207,13 @@
 .easydam-layer--cta-html {
 	font-size: 1.375rem;
 	font-family: Manrope, sans-serif;
-	width: fit-content;
 	margin: 0 auto;
 	padding: 10px 20px;
 	border-radius: 4px;
+	width: 100%;
+	height: 100%;
```
subodhr258, line 213: Same as the block stylesheet copy: the fixed `width: 100%; height: 100%` here moves short CTA content from centered to top-left (the parent `.easydam-layer` flex-centering stops applying once the box fills the overlay). Prefer `width: fit-content; max-width: 100%; max-height: 100%` to keep overflow handling without the centering regression. Keeping this file in sync with the change in `godam-player/style.scss`.

subodhr258 (commented):
Review of the CTA-html overflow fix. The overflow goal works (tall content now stays within the layer and scrolls), and the change is mirrored cleanly across both stylesheet copies. One layout regression worth addressing before merge:

assets/src/blocks/godam-player/style.scss
KMchaudhary: This is how it supposed to be

assets/src/css/godam-player.scss
KMchaudhary: User should align the content using the style tag.
Screenshot 2026-09-01 at 2 34 44 PM

subodhr258 (approved):

PR #2105, opened 2026-09-01
Fix: Author permission to folder creation and renaming

rudrakshi-gupta:
## Issue
#1239 

## Root cause
The `media-folder` taxonomy is registered with **no `capabilities`**, so WordPress applies the default where creating/editing/deleting terms requires **`manage_categories`** — a capability only Editors and Admins hold. Authors (who *can* upload media) are blocked from creating folders.

## Fix
In `inc/classes/taxonomies/class-media-folders.php`, added a `capabilities` map keyed to **`upload_files`**:

```php
'capabilities' => array(
    'manage_terms' => 'upload_files',
    'edit_terms'   => 'upload_files',
    'delete_terms' => 'upload_files',
    'assign_terms' => 'upload_files',
),
```

This tracks media-upload access rather than category management, so **Authors and above** can manage folders while **Contributors/Subscribers stay excluded** (they can't upload media anyway) — matching the permission direction in the issue. It's also consistent with the plugin's existing bulk-folder endpoints, which already gate on `upload_files`.

## Test

Action | Backend (REST) | UI menu
-- | -- | --
Create | 201 ✅ | "New Sub-folder" shown ✅
Rename | 200 ✅ | "Rename" shown ✅
Delete | 403 rest_cannot_delete ✅ | "Delete" hidden ✅
Lock / Bookmark | — | hidden (editor+) ✅

## Demo

https://github.com/user-attachments/assets/cc8ef7be-992f-44dc-baf7-08e3bd5e634d

### DIFF
--- a/inc/classes/class-pages.php
+++ b/inc/classes/class-pages.php
@@ -1303,9 +1303,18 @@ class Pages {
 			'media-library-react',
 			'MediaLibrary',
 			array(
-				'nonce'    => wp_create_nonce( 'wp_rest' ),
-				'userData' => rtgodam_get_user_data( true ),
-				'roles'    => $roles,
+				'nonce'        => wp_create_nonce( 'wp_rest' ),
+				'userData'     => rtgodam_get_user_data( true ),
+				'roles'        => $roles,
+				// Folder permissions resolved from the SAME capabilities the server
+				// authorizes on, so the UI stays in lockstep with REST rather than gating
+				// on hardcoded role slugs (which mismatch for custom roles such as
+				// shop_manager or membership-plugin roles). See issue #1239 review.
+				'capabilities' => array(
+					'manageFolders' => current_user_can( 'upload_files' ),     // Create + rename (manage_terms/edit_terms).
+					'lockFolders'   => current_user_can( 'manage_categories' ), // Lock + bookmark (Editors and above).
+					'deleteFolders' => current_user_can( 'manage_options' ),    // Delete (administrators only).
+				),
 			)
 		);
 
--- a/inc/classes/rest-api/class-media-library.php
+++ b/inc/classes/rest-api/class-media-library.php
@@ -1673,11 +1673,11 @@ class Media_Library extends Base {
 	 * @return \WP_REST_Response|\WP_Error
 	 */
 	public function bulk_delete_folders( $request ) {
-		$user            = wp_get_current_user();
-		$is_allowed_role = ( $user instanceof \WP_User ) && in_array( 'administrator', $user->roles, true );
-		$is_superadmin   = is_multisite() && is_super_admin( $user->ID ) && current_user_can( 'manage_network' );
-
-		if ( ! $is_allowed_role && ! $is_superadmin ) {
+		// Deleting folders is administrator-only. Gating on the `manage_options` capability
+		// (rather than a role-name check) keeps this in lockstep with the media-folder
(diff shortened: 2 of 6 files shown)


KMchaudhary:
## Review notes

Went over the taxonomy caps against the surrounding permission layers. One I would treat as blocking, plus two smaller ones.

### 1. `edit_terms => upload_files` also lets Authors lock/bookmark any folder

WP maps `edit_term_meta` to the taxonomy's `cap->edit_terms` (see `map_meta_cap`), and `WP_REST_Meta_Fields::update_meta_value()` checks exactly that. The `register_term_meta()` calls for `locked` and `bookmark` further down `class-media-folders.php` are `show_in_rest => true` with no `auth_callback`, so lowering `edit_terms` hands Authors write access to that meta as well.

Concretely, an Author can `POST /wp/v2/media-folder/<id>` with `meta.locked = true` on any *unlocked* folder. `enforce_locked_media_folder()` (`inc/classes/rest-api/class-media-library.php:95`) only rejects mutations of folders that are *already* locked, so the request clears the guard — and `update_item_permissions_check` now clears too.

That defeats the editor-only Lock gate this PR deliberately keeps: hidden in `ContextMenu.jsx`, and role-checked in `bulk_update_folder_lock()`. Once a folder is locked, native rename/re-parent/delete and `bulk-delete-folders` all 403 for *everyone* including admins, and only an Editor/Admin can undo it via `bulk-lock-folders`. The same path lets an Author toggle `bookmark` on any folder, including locked ones, since `is_bookmark_only_folder_update()` explicitly permits bookmark-only writes.

Suggested fix: give `locked` and `bookmark` explicit `auth_callback`s requiring the same editor-level check the bulk endpoints use, rather than letting them inherit `edit_terms`.

### 2. The UI gates on role slugs while the server gates on a capability

`hasRole()` matches `window.MediaLibrary.roles` against a hardcoded list, but the server now authorizes on `upload_files`. Those disagree for any role outside the four defaults:

- WooCommerce's `shop_manager` holds `upload_files`, so the server accepts create/rename and the ungated top-level `New Folder` button in `App.js` works — but the context menu shows neither `New Sub-folder` nor `Rename`. A shop manager can create root folders yet cannot nest or rename them.
- Conversely, on a site where a membership/role plugin strips `upload_files` from `author`, both items render and then 403 into a bare `Failed to rename folder` snackbar.

Since aligning the UI with the capability is the whole point of the PR, it may be worth localizing the capability instead — e.g. `'canManageFolders' => current_user_can( 'upload_files' )` on the `MediaLibrary` object — and gating on that rather than on role names.

### 3. `delete_terms => manage_categories` leaves Editors a delete path

`manage_categories` is held by Editors too, so `DELETE /wp/v2/media-folder/<id>?force=true` succeeds for an Editor — while `bulk_delete_folders()` 403s anyone who is not `administrator`/super admin, and the Delete menu item is gated to `superadmin`/`administrator`. The `Delete → 403` row in the test table holds for the Author that was tested, but not for Editors.

This predates the PR (the default was `manage_categories` as well), so it is not something the PR breaks — but this is the commit that declares the caps deliberately, so it seems like the natural place to align them: a cap only admins hold (e.g. `manage_options`) if administrator-only deletion is the intent.

---

Two things I expected to be problems and are not, for the record: `bulk_delete_folders()` and the locked-folder guard both hold up at `upload_files` thanks to their in-handler role checks, and the `assign_terms` change only narrows access for Contributors, who cannot reach the media library anyway.

KMchaudhary (approved):

Issue #2009, opened 2026-07-07
[BUG] Site Icon favicon (and other single-size image requests) 404 for CDN-offloaded images


subodhr258:
## Summary

When an image is served through GoDAM's CDN offload (a "GoDAM-managed image", i.e. one with `rtgodam_image_sizes` / `rtgodam_transcoded_url` meta), single named intermediate sizes requested via `image_downsize()` / `wp_get_attachment_image_src()` resolve to CDN URLs that do not exist and return **404**.

The most visible casualty is the **Site Icon (favicon)**: `get_site_icon_url()` uses this path, so all favicon sizes 404 and the browser falls back to the default globe icon. Any code calling `wp_get_attachment_image_src( $id, 'some-size' )` on a GoDAM-managed image is affected the same way.

This is a gap in the CDN image-size feature added in #1453 (srcset is handled, single-size lookups are not).

## Environment

- GoDAM plugin v2.0.0 (behaviour dates back to the CDN image-size work in #1453 / PR #1412)
- Reproduced on `godam-dev.rt.gw` (Ollie child theme, CDN host `*.gdcdn.us`)

## Steps to reproduce

1. Activate the license so image CDN offload is active.
2. Upload a **PNG** and set it as **Settings → General → Site Icon** (crop → Save).
3. Load any front-end page and inspect the favicon in the browser tab.

## Expected

The favicon renders in the tab (the `<link rel="icon">` sub-sizes resolve to valid images).

## Actual

The favicon is missing (default globe). Every favicon `<link>` points at a CDN sub-size URL that 404s. Observed on the live site:

| URL | Result |
|-----|--------|
| `…/cropped-GoDAM-v2-Logo-1-1.webp` (full size) | 200 ✅ |
| `…/cropped-GoDAM-v2-Logo-1-1-32x32.png` (`rel=icon` 32) | 404 |
| `…-192x192.png`, `-180x180.png` (apple-touch), `-270x270.png` (msTile) | 404 |
| `…-150x150.png` (thumbnail), `-300x300.png` (medium) | 404 |

The attachment is stored locally as a PNG (`mime: image/png`, `…/cropped-GoDAM-v2-Logo-1-1.png`), but its `source_url` is rewritten to the CDN `.webp`, and only the full-size file exists on the CDN.

## Root cause

In `RTGODAM\Inc\Media_Library_Ajax` (`inc/classes/class-media-library-ajax.php`) GoDAM rewrites image URLs to the CDN via filters on:

[...]

PR #2107, opened 2026-09-03
Fix: Site Icon favicon 404 for CDN-offloaded images

rudrakshi-gupta:
<html><head></head><body><h1>Issue #2009 — Site Icon / favicon before vs after fix</h1>
<p><strong>Tested live on</strong> <code>rudrakshi-gupta.rt.gw</code> with GoDAM CDN offload enabled.
<strong>Site Icon attachment:</strong> <code>#3087</code> (PNG). Its full-size URL is offloaded to the CDN
(<code>…/ted70kmuea/cropped-Screenshot-2026-05-21-at-1.33.37-PM-scaled-1.webp</code>), but only the
full-size file was uploaded to the CDN.</p>
<h2>The bug</h2>
<p><code>get_site_icon_url()</code> → <code>wp_get_attachment_image_url()</code> → <code>image_downsize()</code> builds each
favicon size by taking the offloaded (CDN) full URL and swapping in the size-specific
filename — producing a CDN sub-size URL that was never uploaded, so it <strong>404s</strong>. Every
<code>&lt;link rel="icon"&gt;</code> / apple-touch / msTile pointed at a dead URL, so the browser fell back
to the default globe icon.</p>
<h2>The fix</h2>
<p>An <code>image_downsize</code> filter for GoDAM-managed images resolves each size as:</p>
<ol>
<li>size present in <code>rtgodam_image_sizes</code> → authoritative <strong>CDN</strong> URL;</li>
<li>otherwise the locally-generated sub-size file (built from the upload dir, not the CDN base);</li>
<li>otherwise the full-size image.</li>
</ol>
<p>Favicon sizes are never in <code>rtgodam_image_sizes</code>, so they correctly resolve to the <strong>local</strong>
file that actually exists.</p>
<h2>Before / after — every favicon <code>&lt;link&gt;</code> on the front end</h2>

Tag | Size | Before fix (URL WordPress emitted) | Status | After fix (URL served) | Status
-- | -- | -- | -- | -- | --
rel=icon | 32×32 | https://cuunl3cre4.gdcdn.us/ted70kmuea/cropped-…-scaled-1-32x32.png | 404 | https://rudrakshi-gupta.rt.gw/wp-content/uploads/2026/05/cropped-…-scaled-1-32x32.png | 200
rel=icon | 192×192 | https://cuunl3cre4.gdcdn.us/ted70kmuea/cropped-…-scaled-1-192x192.png | 404 | …/uploads/2026/05/cropped-…-scaled-1-192x192.png | 200
apple-touch-icon | 180×180 | https://cuunl3cre4.gdcdn.us/ted70kmuea/cropped-…-scaled-1-180x180.png | 404 | …/uploads/2026/05/cropped-…-scaled-1-180x180.png | 200
msapplication-TileImage | 270×270 | https://cuunl3cre4.gdcdn.us/ted70kmuea/cropped-…-scaled-1-270x270.png | 404 | …/uploads/2026/05/cropped-…-scaled-1-270x270.png | 200

<p>No size 404s after the fix; offloaded sizes still serve from the CDN.</p>
<h2>Screenshots</h2>
<ul>
<li><strong>Before:</strong> <code>godam-2009-before-favicon-404.jpg</code> — the CDN favicon URL returns <em>404 – File Not Found</em>.</li>
<li><strong>After:</strong> <code>godam-2009-after-favicon-200.jpg</code> — the local favicon URL returns the image (200).</li>
</ul></body></html>

## Before
godam-2009-before-favicon-404

## After

godam-2009-after-favicon-200

### DIFF
--- a/inc/classes/class-media-library-ajax.php
+++ b/inc/classes/class-media-library-ajax.php
@@ -1207,11 +1207,15 @@ class Media_Library_Ajax {
 	 * `wp_get_attachment_image_src( $id, $size )` caller is affected. See issue #2009.
 	 *
 	 * Resolution order for a requested size:
-	 *  1. If the named size exists in `rtgodam_image_sizes`, use that authoritative CDN URL.
-	 *  2. Otherwise resolve the locally-generated sub-size file (which does exist on disk),
-	 *     built from the upload directory rather than the CDN base — this covers sizes that
-	 *     were never offloaded (e.g. the Site Icon sub-sizes) and array/numeric requests.
-	 *  3. If neither matches, defer to core (which serves the full-size CDN image).
+	 *  1. Virtual media (GoDAM-tab picks, `_godam_original_id` set) has no local file, so
+	 *     defer to core — its basename-swap on the rewritten CDN full URL yields the correct
+	 *     CDN sub-size URL. (A "local" URL here would 404 and would corrupt the srcset.)
+	 *  2. A named size present in `rtgodam_image_sizes` uses that authoritative CDN URL.
+	 *  3. For real uploads, let core pick the sub-size it would serve (honouring its
+	 *     aspect-ratio tolerance for named and [w, h] requests); serve that size's offloaded
+	 *     CDN copy if it exists, otherwise the local file — which does exist on disk. This
+	 *     covers sizes that were never offloaded, e.g. the Site Icon sub-sizes.
+	 *  4. If none matches, defer to core (which serves the full-size CDN image).
 	 *
 	 * @since 2.2.1
 	 *
@@ -1257,7 +1261,16 @@ class Media_Library_Ajax {
 				return $downsize;
 			}
 
-			// 1. Named size present in the authoritative CDN size map.
+			// 1. Virtual media (GoDAM-tab picks, `_godam_original_id` set) have NO local file:
+			// their attachment metadata stores CDN sub-size basenames, not on-disk files.
+			// Defer to core — its basename-swap on the rewritten CDN full URL yields the
+			// correct CDN sub-size URL. (Building a "local" URL here would 404, and would
+			// also corrupt the legacy virtual-media srcset, which is rebuilt from this src.)
+			if ( ! empty( $godam_original_id ) ) {
+				return $downsize;
+			}
+
+			// 2. Named size present in the authoritative CDN size map → that CDN URL.
 			if ( is_string( $size ) && ! empty( $rtgodam_image_sizes[ $size ]['url'] ) ) {
 				$cdn_size = $rtgodam_image_sizes[ $size ];
 				$width    = isset( $cdn_size['width'] ) ? (int) $cdn_size['width'] : 0;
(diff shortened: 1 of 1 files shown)


KMchaudhary:
## Code review — `image_downsize` filter for CDN-offloaded images

Good diagnosis of #2009: core's `image_downsize()` swapping a sub-size basename into the rewritten CDN full URL is exactly the failure, and short-circuiting the filter is the right layer to fix it at. A few things in the resolution logic look like they'll bite, mostly around **virtual media**.

---

### 1. Virtual-media images regress from working CDN URLs to 404 local URLs — `class-media-library-ajax.php:1271`

The guard at line 1255 treats an attachment as GoDAM-managed when *either* `rtgodam_image_sizes` *or* `_godam_original_id` is set, so virtual-media attachments (GoDAM-tab picks) reach fallback path 2. Those attachments have **no local file at all**, and their metadata was written by `Media_Library::update_image_attachment_meta()`:

- `_wp_attachment_metadata['file']` is backfilled from the CDN URL, so it is a **bare basename** with no date sub-directory (`class-media-library.php:722-728`);
- `_wp_attachment_metadata['sizes'][*]['file']` are **CDN sub-size basenames** (`class-media-library.php:691`).

So `empty( $meta['file'] ) || empty( $meta['sizes'] )` does not bail, `dirname()` of a bare basename returns `.`, `$subdir` becomes `''`, and the filter returns:

```
https://site.tld/wp-content/uploads/<cdn-subsize-basename>.webp   → 404
```

Repro: insert an image from the GoDAM tab, then request it at an array size (`wp_get_attachment_image_src( $id, array( 300, 300 ) )`, `get_site_icon_url()`, a theme calling `the_post_thumbnail( array(...) )`) or at any named size that isn't in `rtgodam_image_sizes`. Path 1 is skipped and path 2 hands back a local URL for a file that was never on disk.

Worth noting this is a **regression, not a pre-existing bug**: for virtual media the old core behaviour (`str_replace( basename( $cdn_full ), $intermediate['file'], $cdn_full )`) produced the *correct* CDN sub-size URL precisely because the metadata filenames are CDN basenames.

Suggested fix: skip path 2 entirely when `_godam_original_id` is set — for virtual media there is no local file to fall back to, so deferring to core is the correct behaviour.

### 2. Legacy virtual-media srcset is rebuilt from the new local src — `class-media-library-ajax.php:1452`

Knock-on effect of #1. For virtual media created before the `rtgodam_image_sizes` map existed (`_godam_original_id` set, empty sizes map), `filter_virtual_media_srcset()` takes the compatibility branch and builds every candidate from `dirname( $image_src )`. `$image_src` now comes from this new filter, i.e. a local uploads URL — so `$base_url` becomes the local uploads directory and **the entire srcset** turns into non-existent local URLs, not just the `src`.

`filter_rtgodam_content_img_tag()` can't repair it either, since it returns early for virtual media (`if ( ! empty( $godam_original_id ) ) return $filtered_image;`).

Fixing #1 fixes this too.

### 3. `ltrim( dirname( $meta['file'] ), '/.' )` mishandles absolute attached paths — `class-media-library-ajax.php:1290`

The sub-directory is derived on the assumption that `$meta['file']` is always relative to the uploads base. On migrated/imported sites `_wp_attached_file` can hold an absolute path (core handles this case explicitly in `_wp_get_attachment_relative_path()`), and here `ltrim` only strips the leading `/`:

```
/var/www/site/wp-content/uploads/2026/05/img.jpg
→ https://site.tld/wp-content/uploads/var/www/site/wp-content/uploads/2026/05/img-300x200.jpg  → 404
```

The `.` in the ltrim charlist would also eat leading dots from a legitimate directory name. Two safer options:

```php
$subdir = _wp_get_attachment_relative_path( $meta['file'] );
```

or mirror core and swap the basename inside the *local* full URL (`$uploads['baseurl'] . '/' . $meta['file']`) rather than re-assembling the path by hand.

### 4. Array/numeric size requests never consult the CDN map — `class-media-library-ajax.php:1261`

Path 1 is gated on `is_string( $size )`, so every `[w, h]` request skips `rtgodam_image_sizes` and falls straight through to the local file — even when that exact size *was* offloaded and has an authoritative CDN URL. That's functionally correct but silently bypasses the offload for `get_site_icon_url()`, `the_post_thumbnail( array( 600, 400 ) )`, and any block/theme passing an array. Matching the array against the map by width/height (with the same ratio tolerance core uses) before falling back would keep CDN delivery for those.

---

### Smaller notes

- `if ( $width > 0 && $height > 0 )` inside path 1 silently drops to the local fallback when a CDN entry is missing dimensions. Probably worth a `_doing_it_wrong`-style log or at least a comment, since the map is meant to be authoritative.
- Path 2's comment says the sub-size "does exist on disk" — true for real uploads (nothing in the plugin suppresses `intermediate_image_sizes_advanced` or deletes local files), but it's an implicit dependency worth stating, since any future "delete local after offload" option would break this branch.

Nice test evidence in the description — the before/after table for each `<link rel="icon">` made this easy to follow. Once the virtual-media path is excluded I think this is good to go; a quick check with a GoDAM-tab image as the featured image would be a useful addition to the manual test matrix.

KMchaudhary (approved):

Issue #1847, opened 2026-05-06
Hide the layer, when we are showing video preview on hover


KMchaudhary:
We are showing the video layer if "Start Preview" is set to Video Hover behaviour.  The purpose of the preview should be video playback, not showing the layers. Hiding the layer on preview will automatically solve the problem.

https://github.com/user-attachments/assets/66dcac48-94bd-4ac7-8e34-9dbe6a1491da

kishan-gondaliya-7270:
Released

PR #2108, opened 2026-09-03
fix(player): don't fire layers during hover Start video preview on hover

KMchaudhary:
**Fixes:** https://github.com/rtCamp/godam/issues/1847

This pull request introduces a mechanism to suppress interactive video layers (such as forms, CTAs, and hotspots) during uncommitted hover previews, ensuring that previews act as pure playback teasers without distractions. The suppression is coordinated between the `HoverManager`, `LayersManager`, and `GodamVideoPlayer` classes. The most important changes are grouped below:

**Hover preview state tracking:**

* Added an `isPreviewPlaying` flag and an `isPreviewActive()` method to `HoverManager` to track when a hover preview is running and expose this state. [[1]](diffhunk://#diff-4035647df222e387856aca8ae9132de0ed0578beedc583f25e702e97b1f221f5R35-R38) [[2]](diffhunk://#diff-4035647df222e387856aca8ae9132de0ed0578beedc583f25e702e97b1f221f5R148-R149) [[3]](diffhunk://#diff-4035647df222e387856aca8ae9132de0ed0578beedc583f25e702e97b1f221f5R168-R182)
* Ensured `isPreviewPlaying` is set appropriately on preview start, on play events, and when the video is clicked to transition from preview to real playback. [[1]](diffhunk://#diff-4035647df222e387856aca8ae9132de0ed0578beedc583f25e702e97b1f221f5R71-R80) [[2]](diffhunk://#diff-4035647df222e387856aca8ae9132de0ed0578beedc583f25e702e97b1f221f5R132)

**Layer suppression logic:**

* Added a suppression mechanism to `LayersManager`, including a `suppressionCheck` predicate, `setSuppressionCheck()`, and `areLayersSuppressed()` methods. These prevent layer firing (including time-based and event-driven triggers) while suppression is active. [[1]](diffhunk://#diff-d27f00ebcb7bf4a8cb4b92e63a2e9a81dd56186cf68999809d48e1837421d501R24) [[2]](diffhunk://#diff-d27f00ebcb7bf4a8cb4b92e63a2e9a81dd56186cf68999809d48e1837421d501L61-R99) [[3]](diffhunk://#diff-d27f00ebcb7bf4a8cb4b92e63a2e9a81dd56186cf68999809d48e1837421d501R167-R170) [[4]](diffhunk://#diff-d27f00ebcb7bf4a8cb4b92e63a2e9a81dd56186cf68999809d48e1837421d501R180-R183) [[5]](diffhunk://#diff-d27f00ebcb7bf4a8cb4b92e63a2e9a81dd56186cf68999809d48e1837421d501R194-R197)

**Integration with player:**

* Registered the suppression check in `GodamVideoPlayer` so that all layers are suppressed while a hover preview is running.

## Demo

https://github.com/user-attachments/assets/a685a0c8-011d-4e3b-9aae-721ef745b8cd

**Improve the hover preview by with video progress clue, and mute/unmute toggle button (Inspired from YT)**

https://github.com/user-attachments/assets/34816651-7310-4715-8b3c-3d9c91f82749

---

## Added preview behaviour (QA notes)

Beyond the layer-suppression fix in #1847, this PR also adjusts hover-preview UX. Flagging so QA knows what to exercise on every `start-preview` embed:

- **Hover-intent delay (~700ms):** a preview now starts only if the pointer lingers; a pointer merely passing over the video no longer starts/stops playback. This changes timing for all existing `start-preview` embeds.
- **Reel-style progress stripe** along the bottom while previewing (presentational, not scrubbable).
- **Mute/unmute toggle** (top-right, player primary colour): toggles preview sound without committing to playback; the choice is remembered for the page session and re-applied only to players the viewer has interacted with (browser autoplay policy).
- **Share/transcript overlay buttons are hidden** while a preview is playing, and restored once it ends or the viewer commits to playback.

Happy to split the chrome into its own PR if preferred — kept together here as they were requested alongside the fix.

### DIFF
--- a/assets/src/css/godam-player.scss
+++ b/assets/src/css/godam-player.scss
@@ -709,13 +709,19 @@
 	background-color: var(--rtgodam-control-bar-color, #3858e9) !important;
 	border: 0;
 	border-radius: 50%;
+	// Hidden until a preview is playing. `visibility`/`pointer-events` (not just
+	// opacity) so the invisible button cannot swallow a click meant to start the
+	// video, and is not tab-reachable or announced while off screen.
 	opacity: 0;
-	// Interactive – must receive clicks even though the stripe does not.
-	pointer-events: auto;
-	transition: opacity 0.15s ease, filter 0.15s ease;
+	visibility: hidden;
+	pointer-events: none;
+	transition: opacity 0.15s ease, filter 0.15s ease, visibility 0s linear 0.15s;
 
 	&.is-visible {
 		opacity: 1;
+		visibility: visible;
+		pointer-events: auto;
+		transition: opacity 0.15s ease, filter 0.15s ease;
 	}
 
 	&:hover,
--- a/assets/src/css/godam-player.scss
+++ b/assets/src/css/godam-player.scss
@@ -642,6 +642,118 @@
 	}
 }
 
+// ---------------------------------------------------------------------------
+// Hover-preview reel chrome
+// Lightweight overlays shown only while a "start-preview" hover preview is
+// playing: a thin reel-style progress stripe along the bottom and a
+// mute/unmute toggle in the top-right corner. Both are appended to the player
+// root by HoverManager and toggled via `.is-visible`.
+// ---------------------------------------------------------------------------
+
+// While a hover preview is playing, hide the overlay share / transcript
+// buttons (they sit on the container, outside the control bar that the preview
+// already hides). Bubble skin routes them into the control bar instead, so they
+// are hidden with it and need no rule here.
+.easydam-video-container.godam-hover-preview-active {
+
+	.godam-share-button,
+	.godam-transcript-button {
+		display: none !important;
+	}
+}
+
+.godam-preview-progress {
+	position: absolute;
+	right: 0;
+	bottom: 0;
+	left: 0;
+	z-index: 3;
+	height: 3px;
+	background-color: rgba(255, 255, 255, 0.25);
+	opacity: 0;
+	// Presentational only – clicks fall through to the video (commit playback).
+	pointer-events: none;
+	transition: opacity 0.15s ease;
+
+	&.is-visible {
+		opacity: 1;
+	}
+
+	&__fill {
+		width: 0;
+		height: 100%;
+		background-color: var(--rtgodam-control-bar-color, #3858e9);
+		transition: width 0.1s linear;
+	}
+}
+
+// The toggle reuses Video.js's `vjs-mute-control vjs-control vjs-button`
+// markup so each skin paints its own volume glyph; these rules only lift it
+// out of the (hidden) control bar and into the top-right corner. The doubled
+// class raises specificity enough to win over the base `.vjs-control` sizing.
+.godam-preview-mute.vjs-control.vjs-button {
+	position: absolute;
+	top: 12px;
+	right: 12px;
+	z-index: 4;
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	width: 3em;
+	height: 3em;
+	padding: 0;
+	margin: 0;
+	color: #fff;
+	cursor: pointer;
+	// Match the player's primary (control bar) colour.
+	background-color: var(--rtgodam-control-bar-color, #3858e9) !important;
+	border: 0;
+	border-radius: 50%;
+	opacity: 0;
+	// Interactive – must receive clicks even though the stripe does not.
+	pointer-events: auto;
+	transition: opacity 0.15s ease, filter 0.15s ease;
+
+	&.is-visible {
+		opacity: 1;
+	}
+
+	&:hover,
+	&:focus-visible {
+		filter: brightness(1.1);
+	}
+
+	.vjs-icon-placeholder::before {
+		position: static;
+		width: auto;
+		font-size: 1.5em;
+		line-height: 1;
+	}
+}
+
+// Classic skin overrides the mute glyph with SVGs scoped to `.vjs-control-bar`;
+// the preview toggle lives outside it, so mirror those SVGs here (only the
+// muted / full-volume states the toggle uses) to match the control bar.
+.godam-classic-skin .video-js .godam-preview-mute {
+
+	/* stylelint-disable-next-line no-descending-specificity -- overlay copy of the classic mute glyph; the control-bar rule keeps its own icon. */
+	&.vjs-vol-0 .vjs-icon-placeholder::before {
+		width: 21px;
+		height: 21px;
+		background: url(../images/classic-mute-icon.svg) no-repeat;
+		background-size: contain;
+		content: '' !important; /* Hide default icon */
+	}
+
+	&.vjs-vol-3 .vjs-icon-placeholder::before {
+		width: 21px;
+		height: 21px;
+		background: url(../images/classic-volume-3-icon.svg) no-repeat;
+		background-size: contain;
+		content: '' !important; /* Hide default icon */
+	}
+}
+
 // CSS to override the elementor image CSS in editor.
 .elementor .vjs-poster img {
 	height: 100% !important;
--- a/assets/src/js/godam-player/managers/hoverManager.js
+++ b/assets/src/js/godam-player/managers/hoverManager.js
@@ -32,6 +32,10 @@ class HoverManager {
 		this.hoverSelect = videoElement.dataset.hoverSelect || 'none';
 		this.isVideoClicked = false;
 		this.isHovered = false;
+		// True while an uncommitted hover preview is running. Interactive layers
+		// are suppressed for its duration – a preview is a playback teaser, not
+		// a place to show forms/CTAs/hotspots.
+		this.isPreviewPlaying = false;
 		this.options = options;
 
 		this.init();
@@ -64,6 +68,16 @@ class HoverManager {
 		this.videoElement.addEventListener( 'mouseenter', this.handleMouseEnter.bind( this ) );
 		this.videoElement.addEventListener( 'mouseleave', this.handleMouseLeave.bind( this ) );
 		this.videoElement.addEventListener( 'click', this.handleVideoClick.bind( this ) );
+
+		// Playback that starts while the pointer is away from the video is real
+		// playback, not a preview – e.g. the big play button or the control bar,
+		// whose clicks never reach the <video> element and so never run
+		// `handleVideoClick`. Clear the preview flag so layers resume working.
+		this.player.on( 'play', () => {
+			if ( ! this.isHovered ) {
+				this.isPreviewPlaying = false;
+			}
+		} );
 	}
 
 	/**
@@ -115,6 +129,7 @@ class HoverManager {
 
 		if ( this.isHovered ) {
 			this.isVideoClicked = true;
+			this.isPreviewPlaying = false;
 
 			this.player.volume( 1 );
 			this.player.play();
@@ -130,6 +145,8 @@ class HoverManager {
 	 * Starts the video preview by playing the video muted and hiding controls.
 	 */
 	startPreview() {
+		this.isPreviewPlaying = true;
+
 		this.player.volume( 0 );
 		this.player.currentTime( 0 );
 
@@ -147,6 +164,21 @@ class HoverManager {
 	stopPreview() {
 		this.player.pause();
 		this.player.currentTime( 0 );
+	}
+
+	/**
+	 * Whether an uncommitted hover preview is in progress.
+	 *
+	 * Deliberately keyed off the live preview flag rather than `hoverSelect`, so
+	 * autoplay players (where `init()` bails out) and `show-player-controls`
+	 * mode can never report a preview. The flag is intentionally NOT cleared by
+	 * `stopPreview()`: `pause()` emits its event asynchronously, so clearing it
+	 * there would let an `on_pause` CTA slip through as the pointer leaves.
+	 *
+	 * @return {boolean} True while a hover preview is active.
+	 */
+	isPreviewActive() {
+		return this.isPreviewPlaying;
 	}
 }
 
--- a/assets/src/js/godam-player/managers/hoverManager.js
+++ b/assets/src/js/godam-player/managers/hoverManager.js
@@ -126,7 +126,13 @@ class HoverManager {
 				this.isPreviewInitiatedPlay = false;
 				return;
 			}
+			// Real, committed playback. Lift suppression and mark the player as
+			// committed so a later hover cannot restart a preview over it – the
+			// big play button / control bar never reach `handleVideoClick`, so
+			// `isVideoClicked` would otherwise stay false and re-enable preview.
 			this.isPreviewPlaying = false;
+			this.isVideoClicked = true;
+			this.togglePreviewOverlays( false );
 		} );
 	}
 
--- a/assets/src/js/godam-player/managers/hoverManager.js
+++ b/assets/src/js/godam-player/managers/hoverManager.js
@@ -1,4 +1,9 @@
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
  * HoverManager
  *
  * A utility class for managing hover-based video interactions using Video.js.
@@ -67,6 +72,10 @@ class HoverManager {
 		// `play` event that signals real, committed playback. Consumed once in the
 		// `play` handler.
 		this.isPreviewInitiatedPlay = false;
+		// True once the viewer has interacted with THIS player (clicked it or its
+		// mute toggle). A cross-preview unmute is only re-applied to players with
+		// a gesture, since browsers pause media unmuted without one.
+		this.hasUserGesture = false;
 		this.options = options;
 
 		this.init();
@@ -109,6 +118,11 @@ class HoverManager {
 		// Advance the reel progress stripe as the preview plays.
 		this.player.on( 'timeupdate', this.updatePreviewProgress.bind( this ) );
 
+		// The preview now starts on a timer; cancel it if the player is disposed
+		// inside the delay window (block-editor re-render, gallery removal, SPA
+		// nav) so the callback never runs against a torn-down player.
+		this.player.on( 'dispose', () => this.clearPreviewTimer() );
+
 		// Any `play` event other than the preview's own means real, committed
 		// playback has begun – e.g. the big play button, the control bar, the
 		// spacebar/keyboard handler, or a programmatic `player.play()`, none of
@@ -126,13 +140,14 @@ class HoverManager {
 				this.isPreviewInitiatedPlay = false;
 				return;
 			}
-			// Real, committed playback. Lift suppression and mark the player as
-			// committed so a later hover cannot restart a preview over it – the
-			// big play button / control bar never reach `handleVideoClick`, so
-			// `isVideoClicked` would otherwise stay false and re-enable preview.
+			// Real, committed playback (big play button, control bar, keyboard,
+			// programmatic). Mark the player committed so a later hover cannot
+			// restart a preview over it, then fully restore anything a preceding
+			// `startPreview()` changed – otherwise this path can leave the player
+			// muted, control-less and stuck behind its own poster.
 			this.isPreviewPlaying = false;
 			this.isVideoClicked = true;
-			this.togglePreviewOverlays( false );
+			this.exitPreviewToPlayback();
 		} );
 	}
 
@@ -175,6 +190,7 @@ class HoverManager {
 		this.previewMuteButton = document.createElement( 'button' );
 		this.previewMuteButton.type = 'button';
 		this.previewMuteButton.className = 'godam-preview-mute vjs-mute-control vjs-control vjs-button';
+		this.previewMuteButton.setAttribute( 'data-test-id', 'godam-preview-mute-toggle' );
 		this.previewMuteButton.innerHTML =
 			'<span class="vjs-icon-placeholder" aria-hidden="true"></span>' +
 			'<span class="vjs-control-text" aria-live="polite"></span>';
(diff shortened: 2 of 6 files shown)


subodhr258:
### Review

Solid, well-scoped fix, and the reasoning in the JSDoc is genuinely helpful. A few things I checked and liked:

- Suppression is registered lazily (`() => this.hoverManager?.isPreviewActive() === true`), so manager construction order does not matter.
- Every reveal path is gated through `areLayersSuppressed()`: the `timeupdate` triggers (`timestamp`, `watch_depth`), the `on_pause` and `end_of_video` event handlers, hotspots, and custom add-on layers.
- No regressions in the sibling modes: autoplay bails out of `init()`, `show-player-controls` never sets the flag, and the block-editor preview (`isPreviewEnabled`) path is untouched. `handlePlay()` is not gated but is a no-op unless a layer is already visible, so that is fine.

Two related observations (same root cause) and one test suggestion.

---

**1. Fast hover-flick can briefly lift suppression before the preview's own `pause` (medium)**

The new `play` listener clears `isPreviewPlaying` whenever `!this.isHovered`. On a quick flick this re-opens the exact `on_pause` window the "do not clear in `stopPreview()`" comment was written to prevent:

1. `mouseenter` runs `startPreview()`: sets `isPreviewPlaying = true`, then `player.play()` (the `play` event is queued).
2. `mouseleave` runs immediately: `stopPreview()` calls `player.pause()` (queues `pause`) and sets `isHovered = false`, deliberately leaving `isPreviewPlaying = true`.
3. The still-pending `play` task fires first (media events are FIFO, and `play` was queued before `pause`). It now sees `!isHovered` and sets `isPreviewPlaying = false`.
4. The `pause` task then fires, finds `areLayersSuppressed() === false`, and `handlePause()` reveals the `on_pause` CTA over the preview the viewer just left.

I confirmed `handlePause()` has no time guard: after `stopPreview()` resets `currentTime(0)` the player is paused-not-ended, so an `on_pause` layer (`show`, `!triggered`, `hidden`) is revealed and marked `triggered`. It needs a video with an `on_pause` CTA plus the `play` dispatch landing after `mouseleave` (likely while the video is still buffering), so it is intermittent rather than guaranteed, but it defeats the stated invariant.

**2. Playback committed while the pointer stays on the video leaves layers suppressed (low)**

The mirror case. `isPreviewPlaying` is only cleared by a click on the `<video>` element (`handleVideoClick`) or by a `play` event when `!isHovered`. If real playback starts while the pointer is still over the video and no click reaches the element (pressing space to resume via the global keyboard handler in `playerManager.js`, or a programmatic `player.play()` through the exposed `player.hoverManager`), neither path runs and every layer stays suppressed for the whole session until the next `mouseleave` or video click. Lower impact, since the video is still muted with controls hidden in that state and it self-heals, but it is the same underlying issue.

Both cases come from inferring "committed to real playback" from `isHovered` or a click target. Deriving it from an explicit signal instead would close both windows: for example clear `isPreviewPlaying` only when the player is actually in real, unmuted playback (gate on `isVideoClicked` or on the unmute already done in `handleVideoClick`), rather than on the raw `!isHovered` condition.

**3. Consider a small unit test for the flag lifecycle (low)**

The new logic is pure and easy to pin down: `areLayersSuppressed()` uses a deliberate strict `=== true`, and the fix depends on `isPreviewPlaying` being set in `startPreview()`, cleared in the right places, and specifically not cleared in `stopPreview()`. A short spec (there is precedent at `assets/src/js/godam-player/utils/layerActions.test.js`) would stop a future refactor from silently reintroducing layers-during-preview or the stuck-suppression case above.

**Nit / future-proofing:** custom add-on layer managers are constructed without any handle to the suppression state, so gating holds only as long as add-ons reveal exclusively through `handleTimeUpdate`. The current WooCommerce manager is safe on that basis, but since the layer registry exists for third-party add-ons, passing a suppression accessor (or documenting the contract) would make the preview invariant enforceable rather than conventional.

KMchaudhary:
Thanks for the thorough review @subodhr258 (and Copilot). Addressed in 736d30a1 and 18ca7ccc.

**1. Fast hover-flick lifting suppression before the preview's own `pause` (medium)** — Fixed. The `play` listener no longer infers "committed to real playback" from `!isHovered`. It now keys off an explicit `isPreviewInitiatedPlay` marker set in `startPreview()` right before `player.play()`: the preview's own (async) `play` is consumed and never clears the flag, so on a flick the trailing `pause` still sees `areLayersSuppressed() === true` and the `on_pause` CTA stays suppressed. Regression test: *"ignores the preview's own play event so a fast flick keeps suppression"*.

**2. Playback committed while the pointer stays on the video (low)** — Fixed by the same refactor: any non-preview `play` (spacebar/keyboard, programmatic, big play button, control bar) clears the flag regardless of hover state, and now also marks the player committed (`isVideoClicked = true`) + hides the preview chrome, so a later hover can't restart a preview over real playback (this also closes Copilot's line-130/180 point about the big-play path leaving `isVideoClicked` false). Tests: *"clears on a real play event even while still hovering"* and *"marks real playback committed so a later hover cannot restart a preview"*.

**3. Unit test for the flag lifecycle (low)** — Done. Added `hoverManager.test.js` (17 tests) covering the strict `=== true`, the flag being set in `startPreview()` and specifically *not* cleared by `stopPreview()`, plus the flick and stuck-suppression cases above.

**Nit — add-on/API custom layers lack a suppression handle** (also Copilot's `layersManager.js:196` re: `GoDAMAPI` / `api/player.js` `handleCustomLayersTimeUpdate`): not addressed here. Propagating the shared suppression accessor into the public API layer path is a broader change that I'd prefer to keep out of this PR's scope — happy to open a follow-up so the invariant is enforceable for third-party add-ons rather than conventional. Let me know if you'd rather it land here.

Separately, this PR now also carries a few preview-UX additions requested alongside the fix (a hover-intent start delay, a reel-style progress stripe, a skin-matched mute/unmute toggle with session-remembered mute, and hiding the share/transcript overlay buttons during preview). Happy to split those into their own PR if you'd prefer to review the layer-suppression fix in isolation.

subodhr258 (changes requested):
### Re-review

Both points from the last round are properly fixed, and I like how. `isPreviewInitiatedPlay` is the right signal (an explicit marker instead of inferring commitment from `isHovered`), it closes the flick case and the still-hovering case together, and the 17 new tests pin the flag lifecycle down so a future refactor cannot quietly undo it. `eslint` and `stylelint` are clean on all four changed files, JS unit tests pass, and the red POT check is the usual repo-wide noise, not this PR.

I am not approving yet. The hover-chrome commit introduces a bug that is worse than the one this PR fixes, and it is reachable in about four seconds of normal use.

---

## Blocker: committing playback from the big play button breaks the player for good

**Steps**

1. Hover a `start-preview` video for 700ms. The preview runs: `muted(true)`, control bar gets `hide`, `vjs-has-started` added.
2. Move the pointer away. `handleMouseLeave` removes `vjs-has-started`, adds `godam-hover-started`, and `stopPreview()` pauses. The poster and big play button come back.
3. Move back and click the big play button, before the 700ms timer fires.

**What you get:** the video plays silently, behind its own poster, with the big play button sitting on top of it and no control bar. There is no way out except a page reload.

**Why**

The `play` handler ([hoverManager.js:124](https://github.com/rtCamp/godam/blob/18ca7ccc/assets/src/js/godam-player/managers/hoverManager.js#L124)) sets `isPreviewPlaying = false` and `isVideoClicked = true` and hides the preview chrome, but it does not undo anything `startPreview()` changed:

- `muted(true)` stays, so playback is silent. If the session preference was unmuted it is the reverse problem: full volume with no way to mute.
- The control bar keeps `hide`. Across `assets/src/js`, `classList.remove( 'hide' )` exists in only two places: `handleVideoClick` (line 362) and `previewManager.js:156`. Neither runs here.
- `vjs-has-started` stays off, and Video.js cannot put it back. `hasStarted(true)` early-returns because `hasStarted_` is already `true` from the preview ([video.cjs.js:23124](https://github.com/videojs/video.js/blob/main/src/js/player.js)). So the poster and big play button stay painted over the playing video.

Then `isVideoClicked = true` closes every recovery path: `handleMouseEnter` (line 281) and `handleVideoClick` (line 345) both early-return forever, and the poster overlay swallows clicks aimed at the `<video>` anyway.

**This is a regression, not a pre-existing quirk.** I ran the same sequence against `develop`: there, hovering again and clicking the video reaches `handleVideoClick`, which unmutes and removes `hide`. The new `play` listener is what sets `isVideoClicked` and takes that escape hatch away.

<details>
<summary>Failing spec I used (drop in as <code>hoverManager.repro.test.js</code>)</summary>

The existing stub cannot catch this: `controlBar: { el: () => document.createElement( 'div' ) }` hands back a fresh div on every call, so no test can observe the `hide` class. This stub keeps a stable control bar and models Video.js's `hasStarted_` latch.

```js
import HoverManager from './hoverManager';

const createPlayer = () => {
	const handlers = {};
	const root = document.createElement( 'div' );
	const controlBarEl = document.createElement( 'div' );
	let mutedState = false, paused = true, hasStarted = false;
	const p = {
		autoplay: () => false,
		el: () => root,
		controlBar: { el: () => controlBarEl },
		on: ( e, cb ) => { ( handlers[ e ] = handlers[ e ] || [] ).push( cb ); },
		addClass: ( c ) => root.classList.add( c ),
		removeClass: ( c ) => root.classList.remove( c ),
		muted: ( v ) => ( v === undefined ? mutedState : ( mutedState = v ) ),
		volume: () => {},
		duration: () => 10,
		currentTime: () => 0,
		pause: () => { paused = true; ( handlers.pause || [] ).forEach( ( cb ) => cb() ); },
		// Video.js: play() on a paused element fires `play`, and hasStarted(true)
		// is a no-op once hasStarted_ is already true.
		play: () => {
			if ( ! paused ) { return; }
			paused = false;
			if ( ! hasStarted ) { hasStarted = true; root.classList.add( 'vjs-has-started' ); }
			( handlers.play || [] ).forEach( ( cb ) => cb() );
		},
	};
	return { p, root, controlBarEl };
};

const mk = ( p ) => {
	const v = document.createElement( 'video' );
	v.dataset.hoverSelect = 'start-preview';
	return new HoverManager( p, v );
};

describe( 'commit via big play button after a hover preview', () => {
	beforeEach( () => jest.useFakeTimers() );
	afterEach( () => jest.useRealTimers() );

	it( 'restores sound, controls and the started state', () => {
		const { p, root, controlBarEl } = createPlayer();
		const m = mk( p );

		m.handleMouseEnter();
		jest.runOnlyPendingTimers();          // preview runs
		m.handleMouseLeave();                 // pointer leaves
		m.handleMouseEnter();
		p.play();                             // big play button

		expect( p.muted() ).toBe( false );
		expect( controlBarEl.classList.contains( 'hide' ) ).toBe( false );
		expect( root.classList.contains( 'vjs-has-started' ) ).toBe( true );
		expect( root.classList.contains( 'godam-hover-started' ) ).toBe( false );
	} );
} );
```

All four assertions fail on `18ca7cc`.
</details>

**Suggested shape of the fix.** `previewManager.js:145-157` already models the right teardown: its `stopPreview()` calls `removeControlBarHide()`. Pulling one `exitPreview()` out of `handleVideoClick` that restores mute state, the control bar, and `vjs-has-started`/`godam-hover-started`, then calling it from both the `play` handler and `stopPreview()`, fixes all three symptoms at once and stops the next exit path from having the same hole.

---

## Also worth fixing before merge

**The hidden mute button still takes clicks and keyboard focus** ([godam-player.scss:694-714](https://github.com/rtCamp/godam/blob/18ca7ccc/assets/src/css/godam-player.scss#L694)). It is hidden with `opacity: 0` while keeping `pointer-events: auto`, and `is-visible` only changes opacity. Video.js's base `.video-js .vjs-control` sets neither `visibility` nor `pointer-events`, so an invisible 3em circle sits 12px from the top-right corner of every `start-preview` player. A viewer who clicks there to start the video hits `handleMuteToggle` instead, which calls `preventDefault()` and `stopPropagation()`: the click is eaten and the video does not start. The button is also tab-reachable and announces "Unmute" while off screen. `visibility: hidden` (or `pointer-events: none` plus `inert`) on the hidden state covers both.

**`isPreviewInitiatedPlay` can get stuck and then suppress layers for the whole session** (line 399). `startPreview()` sets it unconditionally, and only a `play` event clears it. If `play()` rejects without firing `play` (autoplay refused, for example iOS Low Power Mode, which blocks muted autoplay too; or `NotSupportedError`), the `.catch( () => {} )` on line 400 swallows it and both flags stay `true`. The viewer's next real `play` then takes the early-return on line 126, so `areLayersSuppressed()` stays `true` for all of real playback and every form, CTA and hotspot is dead. That is the exact inverse of this PR's goal. I reproduced it with a stub whose `play()` rejects silently. Line 400 is where both flags want rolling back.

**No dispose hook for the new timer** (line 292). Nothing cancels `previewTimer` on teardown. If the player is disposed inside the 700ms window (block-editor re-render, gallery item removed, SPA navigation), the callback runs `player.addClass()` and `startPreview()` against a player whose `el_` is null and throws. `this.player.on( 'dispose', () => this.clearPreviewTimer() )` in `setupPreview()` handles it. New risk, since before this PR the preview started synchronously.

**Cross-video unmute has no user gesture behind it** (line 233). `previewSessionMuted` is module state shared by every player on the page, so after the viewer unmutes one preview, the next preview on a different video calls `muted( false )` from a `.then()` with no interaction on that element. Safari pauses media that becomes unmuted without a user gesture, so those later previews will start and immediately die. Worth gating the re-apply on a gesture having reached that player.

## Smaller things

- **Translate the new labels** (line 251). `'Unmute'` / `'Mute'` are hardcoded, so the only new announced strings in this PR ship untranslatable. Every sibling in `godam-player/` uses `__( ..., 'godam' )`; `transcriptPanelManager.js:288-291` is the pattern for a dynamically built control, including the `data-test-id` this button also needs for the E2E suite.
- **The toggle duplicates Video.js's MuteToggle** (line 175). Hand-written `innerHTML` copying `vjs-mute-control`/`vjs-icon-placeholder`, manual `vjs-vol-0`/`vjs-vol-3` bookkeeping, and then a second copy of the classic skin's SVGs in SCSS that needs its own `stylelint-disable`. Both copies drift when Video.js or the skin changes. Registering a real `MuteToggle` component and positioning it with CSS keeps one source of truth.
- **Test ordering** (`hoverManager.test.js`, last case). The manual `handleMuteToggle` restore works, but if an assertion above it fails, `previewSessionMuted` leaks as `false` into everything after. `jest.resetModules()` in `beforeEach`, or an exported reset, removes the dependency.
- **`isHovered` goes stale** (line 311). `handleMouseLeave` returns on `isVideoClicked` before clearing `isHovered`, so it stays `true` for the life of the page once playback is committed. Harmless today, a trap for whoever next adds a condition on it. Moving `this.isHovered = false` above the guard is enough.
- **API layers still bypass suppression.** Agreed this is follow-up scope. I confirmed `api/player.js:90-95` really does keep its own `timeupdate` listener into `handleCustomLayersTimeUpdate` (line 305), separate from the guarded `layersManager.js:154`. A linked issue referenced from a comment on `areLayersSuppressed()` would make the contract discoverable to add-on authors.

## One scope note

The 700ms hover-intent delay, the reel chrome, the mute toggle, and hiding the share/transcript buttons are all new behaviour that #1847 does not ask for, and the description does not mention the delay at all. The delay silently changes how every existing `start-preview` embed behaves. Worth either splitting the chrome into its own PR or writing it into the description so QA knows to test it.

KMchaudhary:
Thanks for the exceptionally detailed re-review @subodhr258 — the big-play repro and stub were especially helpful. All addressed in 18acfe39.

## Blocker — big play button after a preview
Fixed. Extracted a single `exitPreviewToPlayback()` that restores everything `startPreview()` changed — lifts the control-bar `hide`, `muted(false)` + `volume(1)`, and re-adds `vjs-has-started` / removes `godam-hover-started` (so the latched `hasStarted_` no longer leaves the poster painted over live video). It's called from both the `play` handler and `handleVideoClick`, and `stopPreview()` now also lifts the control-bar `hide` so a leave-then-commit can't reintroduce the hole. I dropped your failing spec in almost verbatim (stateful stub with a stable control bar + `hasStarted_` latch) — all four assertions pass now.

## Worth fixing before merge
- **Hidden mute button eating clicks / focus** — fixed: the hidden state is now `visibility: hidden; pointer-events: none` (not just `opacity`), so it can't swallow a click meant to start the video and isn't tab-reachable or announced off screen.
- **`isPreviewInitiatedPlay` stuck on a rejected `play()`** — fixed: the `.catch` now rolls back both `isPreviewInitiatedPlay` and `isPreviewPlaying` (and hides chrome), so a refused autoplay can't suppress layers for the session. Regression test with a silently-rejecting `play()`.
- **No dispose hook for the timer** — fixed: `this.player.on( 'dispose', () => this.clearPreviewTimer() )` in `setupPreview()`, with a test.
- **Cross-video unmute without a gesture** — fixed: added a per-player `hasUserGesture` (set on a mute-toggle or video click). `getEffectiveMuted()` only carries a cross-preview *unmute* onto a gestured player, so previews on other videos stay muted until the viewer interacts. `handleMuteToggle` now toggles relative to the visible state so it always does the opposite of the icon.

## Smaller things
- **Translation + `data-test-id`** — done: labels go through `__( …, 'godam' )` and the button carries `data-test-id="godam-preview-mute-toggle"`.
- **`isHovered` going stale** — done: cleared before the `isVideoClicked` early return in `handleMouseLeave`.
- **Test ordering leak** — done: exported a `resetPreviewSessionMuted()` test helper called from a top-level `afterEach`.
- **Duplicating Video.js's MuteToggle** — I've kept the hand-written button for now. A registered `MuteToggle` component still wouldn't inherit the classic skin's SVGs (they're scoped to `.vjs-control-bar`, and the component would live outside it), so the SCSS override doesn't go away, and it pulls the Video.js component/localization runtime into the unit tests. Happy to switch to a registered component in a follow-up if you feel the single-source-of-truth win outweighs that — your call.
- **API layers bypass suppression** — agreed, follow-up. Added a note on `areLayersSuppressed()` pointing at `api/player.js` → `handleCustomLayersTimeUpdate` so add-on authors can find the contract; will open a tracking issue to link.

## Scope
Updated the PR description with a "QA notes" section documenting the ~700ms hover-intent delay, reel chrome, mute toggle, and share/transcript hiding, and calling out that the delay changes timing for existing `start-preview` embeds. Still happy to split the chrome into its own PR if you'd rather review the suppression fix in isolation.

subodhr258 (changes requested):
### Re-review

Every point from the last round is properly fixed, and I checked each against `18acfe39`:

- **The big-play-button blocker is gone.** `exitPreviewToPlayback()` now restores mute, the control bar, and the `vjs-has-started` / `godam-hover-started` poster state, and it runs from both the `play` handler and `handleVideoClick`. The new stateful-stub test pins it down (the old stub handed back a fresh control bar on every call and could not have caught it).
- `isPreviewInitiatedPlay` rollback on a refused `play()`, the `dispose` hook for the 700ms timer, the `hasUserGesture` gate for cross-preview unmute, `visibility: hidden` + `pointer-events: none` on the hidden mute button, `__()` on the labels plus a `data-test-id`, and the `isHovered` clear before the early return: all in.
- API-layer suppression is documented as follow-up on `areLayersSuppressed()`, and the 700ms delay is now written into the description.

`eslint` and `stylelint` are clean on all changed files, and the 21 `hoverManager` unit tests pass. The fixes are the right shape.

I am still not approving. The fix commit introduces one new regression that is worse to live with than a first-frame glitch, and it is reachable in normal playback on every `start-preview` embed.

---

## Blocker: mute and volume snap back to full on every resume

The new `play` listener ([hoverManager.js#L138](https://github.com/rtCamp/godam/blob/18acfe39/assets/src/js/godam-player/managers/hoverManager.js#L138)) has no "already committed" guard, so `exitPreviewToPlayback()` runs on **every** `play` event, not just the first commit. That method hard-sets `muted( false )` and `volume( 1 )` ([hoverManager.js#L466-L467](https://github.com/rtCamp/godam/blob/18acfe39/assets/src/js/godam-player/managers/hoverManager.js#L466)).

**Steps**

1. Commit real playback on a `start-preview` video (click it, or the big play button).
2. Mute it, or drag the volume down, from the control bar.
3. Pause, then resume.

**What you get:** the resume fires `play`, `isPreviewInitiatedPlay` is `false`, so the handler falls through to `exitPreviewToPlayback()` and slams the video back to unmuted at full volume. The viewer cannot keep the video muted or quiet across a single pause. This repeats on every resume for the life of the player.

**Why it is a regression:** this `play` listener is new in this PR. On `develop` there is nothing that touches `muted`/`volume` on resume, so the viewer's choice sticks. The listener was added to catch the big-play-button / keyboard / programmatic commit paths, which is correct, but it needs to fire the restore once, on the transition out of preview, not on every subsequent `play`.

<details>
<summary>Failing spec I used (stateful stub that models <code>muted</code> and <code>volume</code>)</summary>

```js
import HoverManager, { resetPreviewSessionMuted } from './hoverManager';

afterEach( () => resetPreviewSessionMuted() );

const makePlayer = () => {
	const handlers = {};
	const root = document.createElement( 'div' );
	let mutedState = false, vol = 1, paused = true, hasStarted = false;
	const p = {
		autoplay: () => false,
		el: () => root,
		controlBar: { el: () => document.createElement( 'div' ) },
		on: ( e, cb ) => ( handlers[ e ] = handlers[ e ] || [] ).push( cb ),
		addClass: ( c ) => root.classList.add( c ),
		removeClass: ( c ) => root.classList.remove( c ),
		muted: ( v ) => ( v === undefined ? mutedState : ( mutedState = v ) ),
		volume: ( v ) => ( v === undefined ? vol : ( vol = v ) ),
		duration: () => 10,
		currentTime: () => 0,
		pause: () => { paused = true; ( handlers.pause || [] ).forEach( ( cb ) => cb() ); },
		play: () => {
			if ( ! paused ) { return; }
			paused = false;
			if ( ! hasStarted ) { hasStarted = true; root.classList.add( 'vjs-has-started' ); }
			( handlers.play || [] ).forEach( ( cb ) => cb() );
		},
	};
	return { p };
};

const mk = ( p ) => {
	const v = document.createElement( 'video' );
	v.dataset.hoverSelect = 'start-preview';
	return new HoverManager( p, v );
};

it( 'keeps the viewer mute + volume across a pause/resume', () => {
	const { p } = makePlayer();
	mk( p );

	p.play();                 // commit real playback
	p.muted( true );          // viewer mutes
	p.volume( 0.2 );          // viewer lowers volume
	p.pause();
	p.play();                 // resume

	expect( p.muted() ).toBe( true );   // fails: false
	expect( p.volume() ).toBe( 0.2 );   // fails: 1
} );
```

Both assertions fail on `18acfe39`.
</details>

**Suggested fix.** Guard the `play` handler so the restore runs only on the first commit, e.g. return early when `isVideoClicked` is already `true` before calling `exitPreviewToPlayback()`. While you are there, `exitPreviewToPlayback()` hard-setting `volume( 1 )` overrides any site-configured or user volume even on that first commit; unmuting is the intent, but leaving the volume level alone is gentler.

## Two smaller things

- **Autoplay-refused teardown is incomplete** ([hoverManager.js#L433](https://github.com/rtCamp/godam/blob/18acfe39/assets/src/js/godam-player/managers/hoverManager.js#L433)). The `.catch()` clears the flags and hides the overlay chrome, but it does not lift the control bar `hide` or restore the poster (`vjs-has-started` stays on, `godam-hover-started` stays off), and it sets `isPreviewPlaying = false`, so a following `mouseleave` skips its teardown too. On a browser that refuses muted autoplay (iOS Low Power Mode), the video is left paused at frame 0 with no controls and no poster. It self-heals on a tap, so it is degraded rather than stuck, but the catch should run the same teardown as `stopPreview()` plus the poster restore.
- **No unit test for the `LayersManager` suppression path.** `setSuppressionCheck` / `areLayersSuppressed` and the new `pause` / `ended` / `timeupdate` guards have no coverage, unlike the well-tested `hoverManager`. A small spec that installs a suppression predicate and asserts `handleFormLayersTimeUpdate` and the `pause`/`ended` handlers stay gated would keep a future refactor from quietly dropping the guard.

KMchaudhary:
Good catch @subodhr258 — that resume regression was real. Fixed in 3fcb4723.

## Blocker — mute/volume snapping back on resume
The `play` handler now returns early when `isVideoClicked` is already `true`, so `exitPreviewToPlayback()` runs only on the *first* transition out of preview, not on every subsequent resume. Your failing spec (stateful stub modelling `muted` + `volume`) is in as a test and both assertions pass — the viewer's mute and volume now stick across pause/resume.

While there, I also took your "gentler" suggestion: `exitPreviewToPlayback()` (and `startPreview()`) no longer touch `volume(1)` at all — mute is handled purely through `muted()`, so a site- or viewer-configured volume level survives the whole preview→commit flow. Added a test asserting a 0.5 configured volume is preserved on commit.

## Smaller things
- **Autoplay-refused teardown** — the `.catch` now runs the same teardown as `stopPreview()` (lifts the control-bar `hide`, pauses, resets time) plus a poster restore (`vjs-has-started` off, `godam-hover-started` on), so a browser that refuses muted autoplay no longer leaves the video paused at frame 0 with no controls and no poster. Regression test with a silently-rejecting `play()`.
- **LayersManager suppression spec** — added `layersManager.test.js`: the strict `=== true` predicate check, the form/hotspot `timeupdate` gating, and the `pause`/`ended` CTA-trigger gating.

Full suite is now 28 tests across the two managers; `eslint`/`stylelint` clean and the build compiles. Ready for another look when you have a moment.

subodhr258 (approved):
### Approving

`3fcb4723` resolves all three points from the last round, and I re-verified each:

- **Resume no longer resets mute/volume.** The `play` handler now returns early when `isVideoClicked` is already set, so `exitPreviewToPlayback()` runs only on the first commit, not on later resumes; and `volume( 1 )` is gone from both `startPreview()` and `exitPreviewToPlayback()`, so the configured volume survives. I re-ran my own repro (mute + drop volume to 0.2, pause, resume) against this commit and both now stick. The two new specs pin it down.
- **Autoplay-refused teardown is complete.** The `.catch()` now calls `showPoster()` + `stopPreview()`, so a refused muted-autoplay leaves the poster and control bar restored instead of a control-less frame 0. Covered by the new "restores the poster and control bar when autoplay is refused" spec. Nice extraction of `showPoster()`, reused from `handleMouseLeave` too.
- **`LayersManager` suppression is now tested.** The new `layersManager.test.js` covers the default state, the strict `=== true` gate, the form/hotspot `timeupdate` paths, and the `pause`/`ended` CTA triggers.

Verification on `3fcb4723`: `eslint` and `stylelint` clean on all changed files; 28 unit tests pass across `hoverManager` and `layersManager` (red POT check is the usual repo-wide noise).

One non-blocking nit for whenever you next touch this file, not for this PR: the `startPreview()` `.then( () => this.applyPreviewMute() )` can re-apply the muted state after a commit if a click lands in the narrow window before the preview's own `play()` promise resolves (it re-reads `getEffectiveMuted()`). It is pre-existing and practically unreachable given the 700ms delay, so it is fine to leave. A `if ( this.isVideoClicked ) return;` at the top of `applyPreviewMute` would close it if you are ever in there.

Thanks for the thorough follow-up on the teardown and the tests.

PR #2109, opened 2026-09-07
chore(deps): resolve open Dependabot security advisories

KMchaudhary:
## What

Resolves the open Dependabot security advisories on `godam` (npm + Composer).

- **npm** fixes are applied through `package.json` `overrides`, plus direct dependency bumps for `axios` (runtime), `browserslist` and `postcss` (dev); `package-lock.json` is then regenerated. No source changes.
- **Composer** bumps the flagged PHPCS dev packages; `composer.lock` is regenerated. `composer.json` is unchanged because its existing constraints (`^3.13`, `^3.1`) already allow the patched versions.

> **Review round 2 applied:** scoped the stale `qs`/`basic-ftp` pins forward to their patched releases, pinned the `nanoid` 3.x floor, and corrected the attributions/tables below. See **OpenTelemetry override** and **Not fixed** for the two items kept as-is with rationale.

Only three bumps reach shipped runtime bundle code (`axios`, `dompurify`, and the `posthog-js` analytics chain — `@opentelemetry/core` / `protobufjs` / `fflate`), plus `@xmldom/xmldom` (via `video.js`) and `nanoid` (via `@uppy`). Everything else is dev/build tooling.

### npm — runtime (shipped in the bundle)

| Package | From | To | Severity | Reaches via |
|---|---|---|---|---|
| axios | 1.16.1 | 1.18.0 | High | direct dep (plugin REST/API calls) |
| nanoid | 5.1.6 | 5.1.16 | High | `@uppy/core`, `@uppy/dashboard` |
| dompurify | 3.4.11 / 3.2.7 | 3.4.13 | Moderate + Low | `isomorphic-dompurify`, `monaco-editor` |
| @opentelemetry/core | 2.2.0 / 2.6.0 | 2.8.0 | Moderate | `posthog-js` (see note below) |
| protobufjs | 7.6.3 | 7.6.5 | Moderate | `posthog-js` |
| @xmldom/xmldom | 0.8.13 | 0.8.15 | Moderate | `video.js` → `mpd-parser` |
| fflate | 0.4.8 | 0.4.9 | Moderate | `posthog-js` |

### npm — development (build / lint / test tooling, not shipped)

| Package | From | To | Severity |
|---|---|---|---|
| browserslist | 4.25.1 / 4.28.1 | 4.28.7 | High |
| fast-uri | 3.1.2 | 3.1.6 | High |
| immutable | 5.1.5 | 5.1.8 | High |
| ip-address | 10.1.1 | 10.3.1 | High |
| js-yaml | 3.15.0 / 4.2.0 | 3.15.1 / 4.3.1 | High |
| shell-quote | 1.8.4 | 1.9.0 | High |
| svgo | 3.3.3 / 4.0.1 | 3.3.4 / 4.0.2 | High |
| linkify-it | 3.0.3 | 5.0.2 | High |
| brace-expansion | 1.1.14 / 5.0.4 | 1.1.18 / 5.0.9 | High |
| postcss | 8.5.15 | 8.5.23 | High + Moderate |
| basic-ftp | 5.2.2 | 5.3.1 | High |
| qs | 6.15.2 | 6.16.0 | Moderate |
| webpack-dev-server | 5.2.5 | 5.2.6 | Moderate |
| markdown-it | 12.3.2 | 14.2.0 | Moderate |
| postcss-selector-parser | 6.1.2 / 7.1.1 | 6.1.3 / 7.1.3 | Low |
| @babel/core | 7.25.7 | 7.29.6 | Low |

`nanoid` also has a `nanoid@<3.3.18 → 3.3.18` floor override so `postcss`'s transitive 3.x copy (three HIGH advisories on develop) can't regress; it currently resolves to 3.3.18 already, the override just pins the floor. The 3.x/5.x split leaves `nanoid` un-hoisted (two shipped copies instead of one — a few hundred extra bytes in `godam-recorder.min.js`); accepted as a deliberate trade to keep `postcss` on 3.x.

### Composer — development (PHPCS tooling)

Two packages were flagged; pulling the WPCS minor brought three of its peers with it:

| Package | From | To |
|---|---|---|
| squizlabs/php_codesniffer | 3.13.5 | 3.13.6 |
| wp-coding-standards/wpcs | 3.3.0 | **3.4.1** (minor) |
| dealerdirect/phpcodesniffer-composer-installer | v1.2.0 | v1.2.1 |
| phpcsstandards/phpcsextra | 1.5.0 | 1.5.1 |
| phpcsstandards/phpcsutils | 1.2.2 | 1.2.3 |

WPCS crosses a **minor** (3.3 → 3.4), which can retune/add sniffs — reviewers may see new PHPCS output on existing code. Verified locally: `composer install` + `npm run lint:php` passes, `composer audit` clean.

## OpenTelemetry override

`@opentelemetry/core` is pinned globally to `2.8.0` (a plain override), not a range-scoped selector. This was raised in review; the global pin is deliberate:

- The **shipped** copies live under `posthog-js` (`2.2.0`/`2.6.0`) and are correctly bumped to `2.8.0` — the runtime advisory is fixed and the bundle is correct.
- The range-scoped form (`@opentelemetry/core@>=2.0.0 <2.8.0`) only takes effect on a **full clean re-resolution**, which churns ~495 unrelated transitives (all of `@babel/*`, etc.). On the incremental install this lockfile is maintained with, npm does not re-apply the range to the already-locked `posthog-js` copies, so they stay vulnerable and the lock is **not idempotent** (`npm install` drifts and re-adds vulnerable copies).
- The global pin's only side effect is collapsing the **dev-only** OpenTelemetry 1.x chain used by `@sentry/node`. That chain is reachable solely through `@wordpress/scripts → @wordpress/e2e-test-utils-playwright → lighthouse → @sentry/node`, and this repo has **no** e2e/Playwright/Lighthouse npm script, CI workflow, or config — nothing imports it, and it is not shipped. Both CI checks (unit tests, Plugin Check) pass.

If preserving that dev chain is later desired, the clean path is bumping `@wordpress/scripts` (see the Lighthouse item under **Not fixed**) so the whole Lighthouse→`@sentry`→OTel stack moves onto OTel 2.x — out of scope for this security PR.

## Not fixed

**No upstream patch (deferred, dev-only, not shipped):**

- **showdown** (GHSA-rmmh-p597-ppvv and related, Moderate, ReDoS) — `showdown@1.9.1`, pulled by **`@wordpress/blocks`** (`^1.9.1`; direct devDependency pinned `15.7.0`, also via `@wordpress/block-editor`). No patched release of showdown exists. `npm audit` names a dependent-level remediation `@wordpress/blocks@15.27.0` (same major); deferred to avoid a ~20-minor jump of the block-editor stack in a security PR.
- **extract-zip** (GHSA-jmr9-qjv8-65gv, High) — the vulnerable copy is `extract-zip@1.7.0`, pulled by **`@wordpress/env`** (direct devDependency pinned `10.28.0`, the tool behind `npm run wp-env`). The Puppeteer chain carries a separate, newer `extract-zip@2.0.1`. GHSA-jmr9-qjv8-65gv covers `<=2.0.1` with no patched release. `npm audit`'s dependent-level fix is `@wordpress/env@11.14.0` (a **major** bump); deferred.

**Patch exists but deferred (dev-only, not shipped, not a Dependabot alert on this repo):**

- **puppeteer-core / @puppeteer/browsers / lighthouse / @wordpress/e2e-test-utils-playwright** (4 × High) — all remediated by `@wordpress/scripts@31.8.0` (a same-minor patch bump from the pinned `31.5.0`). Left out to keep this PR scoped to the Dependabot advisories; a one-line `@wordpress/scripts` bump in a follow-up clears all four (and would also move the `@sentry`/OTel chain — see above).

_`@babel/runtime`, `body-parser`, and `joi` remain in `npm audit` as dev-only transitive tooling; none are Dependabot alerts on this repo._

## Automated checks (done, Node 22.14.0)

- `npm install` reproduces `package-lock.json` with no drift (idempotent); every Dependabot-flagged package resolves to a patched version.
- `npm run build:prod` compiles all bundles with no errors.
- `npm run test:unit` passes (**252 tests, 13 suites**).
- `npm run lint:package-json` passes (also in the pre-commit hook).
- `composer audit` → "No security vulnerability advisories found."
- `require('@sentry/node')` / `require('@opentelemetry/sdk-trace-base')` — n/a (dev chain, unused here); the shipped `posthog-js` OTel chain loads correctly against core `2.8.0`.

## Manual verification / QA steps

Build the plugin (`npm run build:prod`) and activate it on a test site. QA concentrates on the bumps that reach shipped runtime code; the dev-only bumps are covered by the automated checks.

1. **axios (REST / API calls)** — Open the GoDAM media library, upload/transcode a video and watch its status update, load the Analytics dashboard, and validate an API key in Settings. Confirm requests succeed with no new failed requests or console errors.
2. **dompurify (HTML sanitization)** — Open a video with layers/CTAs/hotspots containing rich HTML. Confirm the HTML renders in both the block editor and the front-end player, and that unsafe markup is still stripped (e.g. `` renders with `onerror` removed).
3. **monaco-editor (also pulls dompurify)** — Open the custom CSS/JS code editor in Settings; confirm it loads and saves normally.
4. **posthog-js analytics (@opentelemetry/core, protobufjs, fflate)** — Play a video and interact (play, seek, layer clicks). Confirm analytics events are captured on the GoDAM Analytics dashboard and there are no posthog/opentelemetry errors in the console.
5. **video.js / @xmldom/xmldom (adaptive playback)** — Play a transcoded video using a DASH/adaptive manifest. Confirm playback, quality switching, and seeking work with no console errors.
6. **@uppy uploader (nanoid)** — Upload media through the GoDAM/Uppy dashboard uploader (drag-drop and file picker). Confirm uploads complete.
7. **Player + editor smoke** — Insert/edit the GoDAM video block, save the post, and confirm the front-end renders with no new JS console errors or failed network requests on player load.

No manual steps are needed for the dev-only bumps — they are build/lint/test tooling and are covered by the automated checks.

## Automation impact

None. No source or UI changes; no test-ids, hooks, or public APIs affected.


subodhr258 (commented):
## Review

Checked out the branch, ran a clean `npm install` and `composer install`, and re-ran everything in the description plus the checks it skips.

**Confirmed from the description**

- `npm install` reproduces `package-lock.json` with no drift.
- `npm run build:prod` compiles every bundle.
- `npm run test:unit` passes, 252 tests across 13 suites.
- `composer audit` reports no advisories, and the lock's content hash is unchanged, so `composer.json` really is untouched.

**Also checked, since the description does not cover them**

- `npm run lint:php` passes on PHPCS 3.13.6 with WPCS 3.4.1.
- `markdownlint` still works even though the `markdown-it` override pushes `markdownlint@0.25.1` from its exact `markdown-it@12.3.2` pin up two majors to 14.2.0. It produces normal results on `README.md`, so that jump is safe in practice.
- The build and `@wordpress/babel-preset-default` are fine with `@babel/core` forced from its exact `7.25.7` pin to 7.29.6.
- `npm run lint:css` fails with 40 stylelint errors, but they are all pre-existing style violations in the SCSS sources and unrelated to these bumps.

**One blocking problem.** On the `@opentelemetry/core` line, switching that override from range-scoped to a plain pin breaks `@sentry/node` and `@opentelemetry/sdk-trace-base` at import time. Details and a tested fix are inline.

### Two notes with no line in the diff to attach them to

**`showdown` does not come from markdownlint.** The only thing requiring it is `@wordpress/blocks` (`showdown@^1.9.1`), a direct devDependency pinned at `15.7.0`. markdownlint has no showdown dependency at all. The advisories genuinely have no patch in showdown itself, so "no upstream patch" is correct, but `npm audit` names a remediation one level up:

```
MODERATE  showdown@*   fix=@wordpress/blocks@15.27.0
```

15.7.0 to 15.27.0 stays inside the same major. The very next sentence in the description already refers to `@wordpress/blocks`, so the markdownlint attribution reads like a slip.

**`extract-zip` does not come from the Puppeteer chain.** That chain actually carries the newer `extract-zip@2.0.1`. The copy at the root of the tree is `extract-zip@1.7.0`, pulled by `@wordpress/env` (direct devDependency, pinned `10.28.0`), which is what `npm run wp-env` runs on. `npm audit`:

```
HIGH  extract-zip@*                  fix=@wordpress/env@11.14.0
HIGH  @wordpress/env@1.0.1 - 11.8.0  fix=@wordpress/env@11.14.0
```

GHSA-jmr9-qjv8-65gv covers `<=2.0.1` with no patched release, so the "no patch" call is right again, but the path and the dependent-level fix should be stated correctly, since this section is the record for why deferring is safe.

### On plain pins

The switch from `name@<version` selectors to bare pins is what causes the OpenTelemetry break, and the `qs` and `basic-ftp` entries show the other half of the same cost: a bare pin has no upper bound and no floor, so it both forces majors it should not and freezes packages below their patched version.

The file already contains the shape that works. `svgo@>=3.0.0 <3.3.4` alongside `svgo@>=4.0.0 <4.0.2`, and both `postcss-selector-parser` selectors, patched every copy correctly. Worth reaching for that form first and dropping to a plain pin only where no selector can match, which is genuinely the case for `fflate` because `posthog-js` asks for `^0.4.8`.

package.json
```diff
@@ -98,53 +98,64 @@
...
 		"webpack-remove-empty-scripts": "1.1.1"
 	},
 	"overrides": {
-		"@xmldom/xmldom": "0.8.13",
-		"@babel/core@<7.29.6": "7.29.6",
+		"@xmldom/xmldom": "0.8.15",
+		"@babel/core": "7.29.6",
 		"@babel/plugin-transform-modules-systemjs": "7.29.4",
-		"@opentelemetry/core@<2.8.0": "2.8.0",
+		"@opentelemetry/core": "2.8.0",
```
subodhr258, line 111: **Blocking: this breaks `@sentry/node` and `@opentelemetry/sdk-trace-base` at import time.**

Changing `"@opentelemetry/core@<2.8.0"` to a plain `"@opentelemetry/core"` collapses the whole tree to a single copy. develop has 21 copies of this package; this branch has 1.

Two of the copies that disappear matter. `@opentelemetry/sdk-trace-base@1.30.1` and `@sentry/node@9.47.1` each kept a nested `@opentelemetry/core@1.30.1` on develop. Both pin core at exactly `1.30.1`, and core 2.x dropped the `TracesSamplerValues` export they read while the module loads.

Reproduced on this branch after a clean `npm install`:

```
$ node -e "require('@sentry/node')"
TypeError: Cannot read properties of undefined (reading 'AlwaysOn')
    at node_modules/@opentelemetry/sdk-trace-base/build/src/config.js:25
```

`require('@opentelemetry/sdk-trace-base')` fails the same way. The path is `@wordpress/scripts` to `@wordpress/e2e-test-utils-playwright` to `lighthouse` to `@sentry/node`, so it is development only, but any Playwright or Lighthouse run through `@wordpress/scripts` now throws before it starts. Neither `build:prod` nor `test:unit` loads this path, which is why the checks listed in the description stayed green.

The shipped path is unaffected. `posthog-js` and its 2.x OpenTelemetry packages all load correctly against core 2.8.0.

A range-scoped selector fixes it. I ran this exact change: it patches every 2.x copy to 2.8.0, including the `posthog-js` chain that actually ships, and leaves the development-only 1.30.1 copies nested where they were.

```suggestion
		"@opentelemetry/core@>=2.0.0 <2.8.0": "2.8.0",
```

package.json
```diff
@@ -98,53 +98,64 @@
...
 		"node-forge": "1.4.0",
 		"picomatch@<=2.3.1": "2.3.2",
 		"picomatch@>=4.0.0 <=4.0.3": "4.0.4",
-		"postcss": "8.5.15",
-		"protobufjs@<7.6.3": "7.6.3",
+		"postcss": "8.5.23",
+		"postcss-selector-parser@>=6.1.0 <6.1.3": "6.1.3",
+		"postcss-selector-parser@>=7.1.0 <7.1.3": "7.1.3",
+		"protobufjs@<7.6.5": "7.6.5",
 		"qs": "6.15.2",
```
subodhr258: **This pin is holding `qs` on a version with two open advisories.**

The description lists `qs` as out of scope because it is not a Dependabot alert on this repo, but this override is what freezes it. `qs@6.15.2` is still affected by GHSA-x5fp-wj9c-mxmx (`>=6.14.2 <=6.15.3`) and GHSA-4mjr-xmp4-gh2g (`>=2.2.5 <6.16.0`). Both are patched in 6.16.0.

`npm audit` on this branch confirms a fix is reachable and that the override is the only thing in the way:

```
MODERATE  qs@2.2.5 - 6.15.3   fix=true
```

The PR is already rewriting this block, so it is a cheap win. I verified this resolves cleanly to a single `qs@6.16.0`.

```suggestion
		"qs": "6.16.0",
```

package.json
```diff
@@ -98,53 +98,64 @@
...
+		"@xmldom/xmldom": "0.8.15",
+		"@babel/core": "7.29.6",
 		"@babel/plugin-transform-modules-systemjs": "7.29.4",
-		"@opentelemetry/core@<2.8.0": "2.8.0",
+		"@opentelemetry/core": "2.8.0",
 		"@protobufjs/utf8": "1.1.1",
 		"adm-zip@<0.6.0": "0.6.0",
-		"axios": "1.16.1",
+		"axios": "1.18.0",
 		"basic-ftp": "5.2.2",
```
subodhr258: **Same problem as `qs`, two HIGH advisories this time.**

`basic-ftp@5.2.2` is affected by GHSA-rp42-5vxx-qpwr (`<=5.2.2`, patched 5.3.0) and GHSA-rpmf-866q-6p89 (`<=5.3.0`, patched 5.3.1). `npm audit` on this branch:

```
HIGH  basic-ftp@<=5.3.0   fix=true
```

The description groups this with the advisories it calls out of scope, but the pin here is what keeps it two releases behind. Verified this resolves cleanly.

```suggestion
		"basic-ftp": "5.3.1",
```

package.json
```diff
@@ -86,7 +86,7 @@
 		"@wordpress/notices": "5.41.0",
 		"@wordpress/scripts": "31.5.0",
```
subodhr258, line 87: **Four HIGH advisories in this dependency's tree are neither fixed nor listed.**

`npm audit` on this branch still reports:

```
HIGH  @puppeteer/browsers@<=2.13.2               fix=@wordpress/scripts@31.8.0
HIGH  puppeteer-core@19.8.4 - 24.43.1            fix=@wordpress/scripts@31.8.0
HIGH  lighthouse@... - 13.3.0                    fix=@wordpress/scripts@31.8.0
HIGH  @wordpress/e2e-test-utils-playwright@...   fix=@wordpress/scripts@31.8.0
```

The "Not fixed" section covers only `extract-zip` and `showdown`, so these are left open without a note. 31.5.0 to 31.8.0 is a patch bump inside the same minor, so it does not hit the "outside their pinned ranges" objection the description raises against `npm audit fix --force`.

Either bump `@wordpress/scripts`, or add these to the "Not fixed" list with a reason, so the audit trail is complete.

package.json
```diff
@@ -98,53 +98,64 @@
...
+		"immutable": "5.1.8",
+		"ip-address": "10.3.1",
+		"js-yaml@<4.0.0": "3.15.1",
+		"js-yaml@>=4.0.0": "4.3.1",
 		"launch-editor@<2.14.1": "2.14.1",
+		"linkify-it": "5.0.2",
 		"lodash": "4.18.1",
 		"lodash-es": "^4.18.1",
+		"markdown-it": "14.2.0",
+		"nanoid@>=4.0.0 <5.1.16": "5.1.16",
```
subodhr258, line 135: **The 3.x line was vulnerable too, and nothing here holds its floor.**

The description reads as though the 3.x copy needed no patch ("so `postcss` keeps its 3.x copy"). It did need one. On develop, `postcss`'s `nanoid@3.3.12` was affected by three HIGH advisories: GHSA-2v37-7h3g-55p8 (`<3.3.18`), GHSA-28wg-ghj8-5hjv (`<3.3.16`) and GHSA-xwg4-73v4-xw9w (`<3.3.12`).

It is patched on this branch, but only as a side effect: `postcss@8.5.23` raised its own requirement to `nanoid@^3.3.16`, which happened to pull 3.3.18. Nothing in `overrides` pins that, so a future postcss change could quietly drop back to a vulnerable 3.x copy.

Worth stating it explicitly next to the 5.x selector. I verified this resolves cleanly:

```suggestion
		"nanoid@<3.3.18": "3.3.18",
		"nanoid@>=4.0.0 <5.1.16": "5.1.16",
```

subodhr258, line 135: Small side effect, flagging it only because it lands in shipped code. The range-scoped selector stops npm hoisting `nanoid`, so the tree goes from one shared copy to two.

develop:

```
5.1.6    node_modules/nanoid                  (hoisted)
3.3.12   node_modules/postcss/node_modules/nanoid
```

this branch:

```
5.1.16   node_modules/@uppy/core/node_modules/nanoid
5.1.16   node_modules/@uppy/dashboard/node_modules/nanoid
3.3.18   node_modules/postcss/node_modules/nanoid
         (no hoisted copy)
```

After `npm run build:prod`, `assets/build/js/godam-recorder.min.js` contains nanoid's alphabet twice instead of once. The cost is a few hundred bytes, so this may be fine to accept, but it is a change to bundle output that the description does not mention. I tried `nanoid@>=5.0.0` and it does not re-hoist, so getting back to one copy needs a different shape rather than a quick tweak.

composer.lock
```diff
@@ -984,21 +984,21 @@
         },
         {
             "name": "phpcsstandards/phpcsextra",
```
subodhr258, line 986: **Three Composer packages changed that the description does not list.**

The Composer table names `squizlabs/php_codesniffer` and `wp-coding-standards/wpcs`, and the text says it "bumps the two flagged PHPCS dev packages". The lock actually moves five:

| Package | From | To |
|---|---|---|
| dealerdirect/phpcodesniffer-composer-installer | v1.2.0 | v1.2.1 |
| phpcsstandards/phpcsextra | 1.5.0 | 1.5.1 |
| phpcsstandards/phpcsutils | 1.2.2 | 1.2.3 |
| squizlabs/php_codesniffer | 3.13.5 | 3.13.6 |
| wp-coding-standards/wpcs | 3.3.0 | **3.4.1** |

The WPCS move is a minor, not a patch, and WPCS minors usually add or retune sniffs, which can surface new errors on existing code.

Nothing is actually broken: I ran `composer install` then `npm run lint:php` on this branch and it passes, and `composer audit` is clean. It is the table that needs the extra three rows plus a note that WPCS crosses a minor, so a reviewer knows to expect possible new PHPCS output.

package.json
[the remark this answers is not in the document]
subodhr258: Agreed on the conclusion, and confirmed by running it. One correction on the detail: the first thing that actually throws is not `getEnv()`, it is the `TracesSamplerValues` export, which core 2.x also dropped. It fails while the module loads rather than when the tracing path initializes.

After a clean `npm install` on this branch:

```
$ node -e "require('@sentry/node')"
TypeError: Cannot read properties of undefined (reading 'AlwaysOn')
    at node_modules/@opentelemetry/sdk-trace-base/build/src/config.js:25
```

`require('@opentelemetry/sdk-trace-base')` fails identically. So this is not conditional on initializing tracing, any import of the chain throws.

On the suggested remedy: scoping to the 2.x chain is the right call and it does not need a nested override. I tested `"@opentelemetry/core@>=2.0.0 <2.8.0": "2.8.0"` and it patches all the 2.x copies, including the `posthog-js` chain that ships, while leaving `sdk-trace-base` and `@sentry/node` on their nested 1.30.1 copies. Details in my comment on the same line.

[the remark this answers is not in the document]
KMchaudhary: Thanks — confirmed the mechanism: core 2.x dropped the `TracesSamplerValues`/`getEnv` exports, so the 1.x `@sentry`/`@opentelemetry/sdk-trace-base` chain throws at import when forced to 2.8.0.

Full rationale for keeping the global pin is in the reply on @subodhr258's thread on this line. Short version: the **shipped** `posthog-js` copies are correctly on 2.8.0, and the only chain affected by the pin is dev-only tooling that nothing in this repo runs.

KMchaudhary: Great catch, and the reproduction is spot on. I went deep on this and ended up keeping the global pin — the scoped selector runs into two problems on this tree:

**1. It isn't idempotent here.** `@opentelemetry/core@>=2.0.0 <2.8.0` only takes effect on a full clean re-resolution (`rm -rf node_modules package-lock.json && npm install`), which churns ~495 unrelated transitives (all of `@babel/*` to 7.29.7, etc.). On the incremental install this lock is maintained with, npm doesn't re-apply the range to the already-locked `posthog-js` copies — they stay on 2.2.0/2.6.0, and a second `npm install` then drifts the lock and re-adds the vulnerable copies. That breaks the "`npm install` reproduces the lock with no drift" property you verified, and re-opens the shipped alert.

**2. The chains can't be split by scope.** `posthog-js` (runtime) and `@sentry/node` (dev) share `@opentelemetry/*` packages (`sdk-trace-base` in particular), so a `@sentry`-scoped carve-out either still breaks `@sentry`'s import (`sdk-trace-base` hoists to root and gets 2.8.0) or leaks a vulnerable runtime `core@1.30.1`. I tested both.

The global pin's only casualty is the dev-only 1.x chain under `@wordpress/scripts → @wordpress/e2e-test-utils-playwright → lighthouse → @sentry/node`. This repo has no e2e/Playwright/Lighthouse npm script, CI workflow, or config, so nothing imports it and it isn't shipped — the shipped `posthog-js` OTel chain is correctly on 2.8.0 and both CI checks pass.

If we'd rather preserve that dev chain than pin over it, the clean path is bumping `@wordpress/scripts` (the same bump that clears the puppeteer/lighthouse HIGHs on the other thread) so the whole Lighthouse→`@sentry`→OTel stack moves onto OTel 2.x. Happy to do that as a follow-up — I've flagged it in the PR description under *Not fixed*. Leaving this thread open for your call.

package.json
KMchaudhary: Applied — `qs` → 6.16.0. You're right that the stale pin was the only thing holding it back; it's now cleared from `npm audit` too.

package.json
KMchaudhary: Applied — `basic-ftp` → 5.3.1. Same story as `qs`; cleared from `npm audit`.

package.json
KMchaudhary: Added to the *Not fixed* section with the correct dependent-level remediation (`@wordpress/scripts@31.8.0`, a same-minor patch). I left the bump itself out to keep this PR scoped to the Dependabot advisories — these four are npm-audit-only, not Dependabot alerts on this repo. A one-line `@wordpress/scripts` follow-up clears all four (and would also move the `@sentry`/OTel chain — see the OpenTelemetry thread). Leaving this open in case you'd prefer I fold that bump in here.

package.json
KMchaudhary: Applied — added `nanoid@<3.3.18` → `3.3.18` alongside the 5.x selector, and corrected the description wording (the 3.x copy did need a patch; it currently resolves to 3.3.18 via `postcss@8.5.23`, the override just pins the floor so it can't regress).

KMchaudhary: Acknowledged — accepting the two-copy trade to keep `postcss` on 3.x. Noted the ~few-hundred-byte `godam-recorder.min.js` increase in the PR description so it's on record.

composer.lock
KMchaudhary: Fixed the description — the Composer table now lists all five moved packages (composer-installer, phpcsextra, phpcsutils alongside php_codesniffer and wpcs) and flags that WPCS crosses a **minor** (3.3 → 3.4), so reviewers know to expect possible new PHPCS output. `npm run lint:php` still passes on the branch and `composer audit` is clean.

subodhr258 (approved):
Reviewed and approved. I checked each bump against the live npm and Composer advisory databases, not only the PR body.

**What this fixes (all confirmed at or above the first patched version):**

Runtime, shipped in the bundle:
- axios 1.16.1 to 1.18.0 (10 advisories, including high prototype-pollution and DoS)
- dompurify to 3.4.13 (XSS chain through GHSA-55q2-fjhq-7xh7; note the old 3.4.11 pin was itself still vulnerable)
- nanoid (both the 5.x and 3.x floors), @xmldom/xmldom 0.8.15, @opentelemetry/core 2.8.0, protobufjs 7.6.5, fflate 0.4.9

Composer dev tooling:
- squizlabs/php_codesniffer 3.13.6 (CVE-2026-67434), wp-coding-standards/wpcs 3.4.1 (CVE-2026-45293), phpcsstandards/phpcsutils 1.2.3 (CVE-2026-65954), plus phpcsextra and the installer

npm audit goes from 137 findings on develop to 16 on this branch, and all 16 left are dev-only and listed in your "Not fixed" section with reasons.

Manifest and lockfile agree, no shipped-runtime major bumps, and all 6 checks are green (including Plugin Check and JS unit tests).

**Two things for release, not blockers:**

1. `assets/build/` is gitignored, so this lockfile bump reaches users only when the next tagged release rebuilds. Tie it to that build, not to the merge.
2. axios and dompurify touch live behavior (REST calls and HTML sanitization), so run a quick smoke test on a build made from this branch before release:
   - axios: media upload plus transcode status updates, the Analytics dashboard load, and API key validation in Settings.
   - dompurify: open a video with layers/CTAs/hotspots that contain rich HTML, in both the editor and the front-end player; confirm safe HTML renders and unsafe markup is stripped (for example an `onerror` handler is removed).
   - posthog analytics chain: play and interact with a video; confirm events land on the dashboard and there are no posthog or opentelemetry console errors.

Approving.

Issue #1355, opened 2025-11-19
API Error Handling Needs Enhancement (Central + Plugin)


kishan-gondaliya-7270:
### Problem Summary  
- When switching environments or using an invalid API key, the plugin fails silently.  
- Infinite loading appears instead of showing a clear "Invalid API Key" message.  
- API responses are not validated on plugin load.  
- If an API key that was once valid gets invalidated/removed from Central later, the plugin does **not** detect it until failure.  
- Appropriate 401/403 handling is missing.

---

### Steps to Reproduce  
1. Install the plugin and configure it with **Pre-Prod** environment URL.  
2. Add a **valid Pre-Prod API key** → everything works.  
3. Remove the plugin.  
4. Reinstall the plugin and configure it with **Prod** environment URL.  
5. Navigate to **Post/Page** → add **GoDAM Video Block**.  
6. Open the **GoDAM tab** in the block.  
7. Observe:  
   - Infinite loading  
   - Console error: *Unexpected response*  
8. Now add a **valid Prod API key** → plugin starts working again.

---

### Expected Behavior  
- Plugin should validate API key **immediately when loaded** (if it exists and should be removed from cache if plugin is deleted).  
- When API returns invalid key (401/403), plugin should clearly show **"Invalid API Key"** error.  
- No infinite loading UI should occur.  
- API errors should always be surfaced clearly.

---

### Actual Behavior  
- Infinite loading UI in video block.  
- "Unexpected response" console error.  
- No user-facing indication that API key is invalid.  
- Plugin does not validate API key until API is actually called.

rudrakshi-gupta:
Completed.

kishan-gondaliya-7270:
Released

PR #2112, opened 2026-09-08
Fix: API Key Error handling

rudrakshi-gupta:
<h2>The issue (what was happening)</h2>

#1355 

<p>The GoDAM tab in the Video block's media picker fetches your video library from Central using the stored API key. When that key is missing, invalid, or was revoked (e.g. switching environments), the request fails — and the UI handled failure badly in two ways:</p>
<ul>
<li><strong>HTTP-level failure (401/403):</strong> the tab was stuck on an <strong>infinite loading spinner</strong> that never stopped.</li>
<li><strong>Proxy relaying an error (<code>success:false</code>):</strong> the tab silently showed <strong>"No items found"</strong>, as if you simply had no videos — no indication the key was the problem.</li>
</ul>
<p><strong>Root cause:</strong> In <code>models/attachments.js</code>, the <code>more()</code> override only handled the success case:</p>
<pre><code class="language-js">mirroring.more( options ).done( function() { deferred.resolveWith( attachments ); ... } );
</code></pre>
<p>On a failed fetch, <code>.done()</code> never fires, so the deferred <strong>never resolves</strong> → the media browser's spinner spins forever. And nothing ever surfaced the backend's error message.</p>
<h2>The fix</h2>
<ol>
<li><code>models/attachments.js</code>:
<ul>
<li><code>more()</code> now uses <code>.always()</code> so the deferred <strong>always resolves</strong> (spinner clears on success <em>or</em> failure).</li>
<li><code>sync()</code> records a friendly message (<code>deriveGodamTabError()</code> maps 401/403 → "Invalid or expired GoDAM API key…") on both the <code>success:false</code> and HTTP-error paths, and clears it on success.</li>
<li>After each fetch it fires <code>godam:fetched</code> with the error (or null).</li>
</ul>
</li>
<li><code>views/godam-media-frame-shared.js</code>: <code>GoDAMCreate</code> listens for <code>godam:fetched</code> and renders (or clears) a clear error notice in the tab (<code>renderGoDAMTabNotice</code>). Done at the frame level so it also works in additive mode.</li>
</ol>
<h2>Before / after</h2>

State | Before | After
-- | -- | --
Invalid/expired key (401/403) | ♾️ infinite spinner, Unexpected response in console | Spinner stops; red notice "Invalid or expired GoDAM API key. Please verify your API key on the GoDAM settings page."
Proxy error (success:false) | Misleading "No items found" | Same clear error notice
Valid key | Works | Works — 40 videos load, no notice (no regression, verified above)

<h2>How to test after the fix</h2>
<ol>
<li>Add a <strong>GoDAM Video block</strong> to a post → click <strong>Add Video</strong> → open the <strong>GoDAM</strong> tab.</li>
<li><strong>Valid key:</strong> your videos load normally (no spinner, no notice).</li>
<li><strong>Invalid key:</strong> temporarily change the API key on the GoDAM settings page to an invalid value (or switch to an environment where the key isn't valid), then reopen the GoDAM tab → you should see the <strong>red "Invalid or expired GoDAM API key" notice</strong> instead of an endless spinner or "No items found".</li>
<li>Restore a valid key and reopen → the notice is gone and videos load again.</li>
</ol>
<p>To simulate without touching your key, in DevTools you can force the endpoint to fail:</p>
<pre><code class="language-js">const o = jQuery.ajax;
jQuery.ajax = f =&gt; f?.url?.includes('get-godam-cmm-files')
  ? jQuery.Deferred(d =&gt; setTimeout(() =&gt; { f.error?.({status:401}); d.reject(); }, 300)).promise()
  : o(f);
</code></pre>
<p>then open the GoDAM tab.</p>

### Before

https://github.com/user-attachments/assets/83b42874-17a8-4a00-a280-ee6869b66126

### After
https://github.com/user-attachments/assets/c0a6d955-4337-4f60-8db2-e6b4b479527d

### DIFF
--- a/assets/src/css/media-library.scss
+++ b/assets/src/css/media-library.scss
@@ -738,3 +738,11 @@ $media-library-active-folder: #f5f5f5;
 		background-size: contain !important;
 	}
 }
+
+// Error notice shown inside the GoDAM media-modal tab (e.g. invalid/expired API key).
+// Colours, border and padding come from core's `.notice.notice-error`, so it follows the
+// active admin colour scheme (including dark mode); this only handles placement inside the
+// modal so the message sits above the attachments grid.
+.godam-tab-notice {
+	margin: 12px 16px;
+}
--- a/assets/src/js/media-library/models/attachments.js
+++ b/assets/src/js/media-library/models/attachments.js
@@ -8,6 +8,35 @@
  */
 
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Derive a clear, user-facing message from a failed GoDAM tab request.
+ *
+ * The GoDAM tab proxies to Central with the stored API key. When that key is missing,
+ * invalid, or was revoked, the request fails with a 401/403 (or the REST proxy returns
+ * `success:false` carrying the upstream status). Without this the tab either span forever
+ * or showed a misleading "No items found", so we translate the failure into an actionable
+ * message the UI can display.
+ *
+ * @param {Object} source The failed jqXHR, or the `{ success:false, message }` response body.
+ * @return {string} A human-readable error message.
+ */
+function deriveGodamTabError( source ) {
+	const status = source?.status ?? null;
+	const message = source?.responseJSON?.message || source?.message || '';
+
+	// Authentication/authorization failures → the API key is the problem.
+	if ( 401 === status || 403 === status || /\b40[13]\b/.test( message ) ) {
+		return __( 'Invalid or expired GoDAM API key. Please verify your API key on the GoDAM settings page.', 'godam' );
+	}
+
+	return message || __( 'Could not load your GoDAM media. Please try again.', 'godam' );
+}
+
+/**
  * Extended Attachments Collection to override the default `_requery` behavior.
  * It mirrors the query with a custom query model (`wp.media.godamQuery`).
  */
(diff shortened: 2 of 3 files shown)


KMchaudhary:
### Code review — API Key error handling

Solid, well-documented fix — the `.always()` change and the `godam:fetched` → notice wiring are the right shape, and rendering at the frame level (so additive mode works) is a nice touch. A few minor points, none blocking:

**1. `deriveGodamTabError` — 401/403 substring match is brittle** (`models/attachments.js`)
The status check (`401 === status || 403 === status`) is reliable, but the fallback `/\b40[13]\b/.test( message )` will classify *any* failure whose relayed message merely contains a bare `401`/`403` token (an id, a count, cURL text) as an API-key problem — sending the user to re-verify a perfectly valid key. Consider dropping the message-substring branch and trusting the status code (which the proxy already encodes as `HTTP status: 401`).

**2. Notice hardcodes inline CSS despite `notice notice-error` classes** (`views/godam-media-frame-shared.js`, `renderGoDAMTabNotice`)
The element already carries `notice notice-error`, then `.css({...})` bakes in `#fcf0f1` / `#d63638` / `#1d2327`. Under a dark admin color scheme those light-mode colors clash. Leaning on the existing classes (or a stylesheet rule) keeps it in sync with core admin-notice theming.

**3. `attachments._godamError` is dead intermediate state** (`models/attachments.js`, `more()`)
It's assigned only to be read back on the next two lines as the event arg; nothing else consumes it. Passing `mirroring._godamError || null` directly to `deferred`/`trigger` does the same job without leaving a stale, unused field on the collection.

**4. Stale `_godamError` can be re-emitted on a short-circuited `more()`** (`models/attachments.js`)
`_godamError` is only ever cleared inside the success branch. After a failure sets `_hasMore = false`, a later `Attachments.more()` short-circuits at `if ( ! this.hasMore() )` without running `sync()`, yet `.always()` still reads `mirroring._godamError` and re-fires `godam:fetched` with the old message. Benign within a session (same error), but worth clearing on the non-error early-return paths if you want it strictly tied to the current fetch.

_Reviewed with `/code-review` (medium effort) — correctness + cleanup pass._

KMchaudhary (approved):

PR #2115, opened 2026-09-11
 Resolve: the plugin deletion error issue on QM

### LINKED ISSUE #1146: Plugin Deletion Error with Query Monitor Activated
#### Description
When attempting to delete a plugin while the **Query Monitor** plugin is active, an error is thrown.  
However, after refreshing the page, the plugin is successfully deleted.  

No such issue is observed when Query Monitor is not active.  Need to check with debugger by @subodhr258 

---

#### Steps to Reproduce
1. Ensure **Query Monitor** plugin is installed and activated.  
2. Go to the **Plugins** page in WordPress admin.  
3. Attempt to delete a GoDAM plugin.  
4. Notice an error message appears.  
5. Refresh the page — the plugin is deleted successfully.  

---

#### Expected Behavior
- Plugin should be deleted without errors, regardless of whether Query Monitor is active.  

#### Actual Behavior
- Error message appears during deletion with Query Monitor active.  
- Plugin is actually deleted, but confirmation requires a page refresh.  

#### Visual Proof

Image

---

#### Acceptance Criteria 
- [ ] Plugin can be deleted successfully without errors, even when Query Monitor is active.  
- [ ] Deletion reflects immediately without requiring a page refresh.  
- [ ] No errors or warnings in admin UI or browser console.  
- [ ] Behavior is consistent whether Query Monitor is active or inactive.  
- [ ] Tested with rtMedia plugin

rudrakshi-gupta:
**Fix: Plugin deletion error (502 / "Deletion failed") when GoDAM is deleted**

Closes #1146

### Problem
Deleting the GoDAM plugin threw an error — a `502 Bad Gateway` (openresty), or under Query Monitor a "Deletion failed: … Class `ActionScheduler_Lock` not found" — even though the plugin was actually removed (a page refresh confirmed it). Other plugins deleted cleanly; only GoDAM failed.

### Root cause
GoDAM cleaned up on delete via `register_uninstall_hook()`, which makes WordPress **include and fully bootstrap `godam.php`** during deletion — loading the bundled WooCommerce Action Scheduler and registering its `shutdown` callback / autoloaders. WordPress then deletes the plugin's files, so on `shutdown` those now-dangling references fatally autoload classes that no longer exist. Query Monitor appends that fatal to the delete AJAX response (corrupting the JSON), and on buffered servers the FPM worker dies before the response flushes, returning a 502.

### Fix
- Added an **`uninstall.php`**. WordPress runs it in isolation (with `WP_UNINSTALL_PLUGIN` defined) and **does not bootstrap the plugin**, so Action Scheduler is never loaded during deletion and no dangling references exist at `shutdown`. Moved the option cleanup here and removed `register_uninstall_hook()`.
- Added a `delete_plugin` guard in `godam.php` that detaches Action Scheduler's `shutdown` dispatch as a safety net for the case where GoDAM is loaded during the delete request.

### Testing
- Deleting a fixed test copy on a local site (Query Monitor active) returns a clean `{success:true}` and removes immediately — no fatal, no 502, no refresh needed.
- Verified working on a separate environment.
- Non-GoDAM plugins unaffected.

### Changed files
- `godam.php`
- `uninstall.php` (new)

### Demo
Screenshot 2026-09-10 at 5 22 25 PM

### DIFF
--- a/godam.php
+++ b/godam.php
@@ -144,45 +144,49 @@ function rtgodam_plugin_deactivate() {
 
 register_deactivation_hook( __FILE__, 'rtgodam_plugin_deactivate' );
 
-/**
- * Runs when the plugin is deleted.
+/*
+ * Plugin-state cleanup on deletion lives in uninstall.php (run in isolation by WordPress),
+ * NOT in a register_uninstall_hook() callback. The register_uninstall_hook() path makes
+ * WordPress bootstrap this whole file (and load bundled Action Scheduler) during deletion,
+ * which then fatals on `shutdown` once the files are gone — see uninstall.php and issue #1146.
  */
-function rtgodam_plugin_delete() {
-	// Delete options related to plugin state (What's New, version).
-	// This is to ensure redirection on a fresh install.
-	if ( is_multisite() ) {
-		// Get all blogs in the network and delete options from each blog.
-		$blogs = get_sites( array( 'fields' => 'ids' ) );
 
-		foreach ( $blogs as $blog_id ) {
-			switch_to_blog( $blog_id ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.switch_to_blog_switch_to_blog
+/**
+ * Prevent a `shutdown` fatal when GoDAM is deleted while active.
+ *
+ * GoDAM bundles WooCommerce Action Scheduler, whose queue runner registers a `shutdown`
+ * callback (`ActionScheduler_QueueRunner::maybe_dispatch_async_request`). Deleting the
+ * plugin removes its files mid-request, so on `shutdown` that callback tries to autoload a
+ * class from the now-deleted plugin (e.g. `ActionScheduler_Lock`) and throws a fatal. A
+ * fatal-catcher such as Query Monitor appends that error to the delete-plugin AJAX body,
+ * corrupting the JSON so the admin shows "Deletion failed" even though the plugin was
+ * actually removed. `delete_plugin` fires immediately before the files are deleted, so
+ * detaching Action Scheduler's shutdown dispatch here avoids the fatal entirely (and the
+ * async request it would dispatch is pointless when the plugin is going away). See #1146.
+ *
+ * @since 2.2.1
+ *
+ * @param string $plugin_file Plugin being deleted, relative to the plugins directory.
+ * @return void
+ */
+function rtgodam_detach_action_scheduler_on_delete( $plugin_file ) {
+	if ( plugin_basename( __FILE__ ) !== $plugin_file ) {
+		return;
+	}
 
-			delete_option( 'rtgodam_plugin_version' );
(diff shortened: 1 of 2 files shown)


KMchaudhary:
### Code review — regression check

Reviewed the diff (`godam.php`, new `uninstall.php`) against the merge-base of `develop`. **Overall the fix is sound and I don't see a regression risk in the normal delete flow.** Verified the core mechanics:

- Both Action Scheduler APIs the guard relies on exist in the bundled copy — `unhook_dispatch_async_request()` and the `maybe_dispatch_async_request` shutdown callback (`vendor/woocommerce/action-scheduler/classes/ActionScheduler_QueueRunner.php:95,101`), so both the primary path and the `remove_action` fallback are correct.
- `uninstall.php`'s option/transient list matches the old `rtgodam_plugin_delete()` cleanup **exactly** (8 options + `rtgodam_show_whats_new`/`rtgodam_release_data` transients) — no cleanup coverage is lost. All of these are stored per-site via `update_option`/`set_transient`, so the per-site `delete_option`/`delete_transient` inside the `switch_to_blog` loop is correct.
- `.distignore` is a denylist and does **not** exclude `uninstall.php`, so it ships in the release build — cleanup won't silently stop after `register_uninstall_hook()` removal.
- The `delete_plugin` guard is correctly gated on `plugin_basename( __FILE__ )`, so bulk-deleting other plugins is unaffected.

A few things worth addressing (none block the fix):

**1. Multisite cleanup silently caps at 100 sites — `uninstall.php:63`**
`get_sites( array( 'fields' => 'ids' ) )` inherits `WP_Site_Query`'s default `'number' => 100`. On a network with more than 100 sites, `rtgodam_*` options/transients are left behind on every site past the first 100. This isn't new (the removed `register_uninstall_hook` code had the same bug), but since the loop is being rewritten here it's a good moment to fix it:
```php
$rtgodam_blog_ids = get_sites( array( 'fields' => 'ids', 'number' => 0 ) );
```

**2. Guard removes only one of AS's two shutdown callbacks — `godam.php:186`**
`ActionScheduler_FatalErrorMonitor::handle_unexpected_shutdown` is also hooked on `shutdown` (`ActionScheduler_FatalErrorMonitor.php:45`), attached while a queue batch is running. In the narrow case where GoDAM is deleted while active *and* a queue run is in flight (e.g. `wp plugin delete` during a batch), that callback stays hooked and could dangle the same way. Very edge — the monitor is detached outside a run — but flagging since it's the same failure class the guard targets.

**3. Minor: the closure in `uninstall.php:50`** — `uninstall.php` runs once in isolation, so the "avoids redeclaration concerns" rationale doesn't really apply; a plain inline loop would be simpler. Purely cosmetic.

Nice, well-documented fix overall — the isolation rationale in both files is very clear. 👍

KMchaudhary (approved):

PR #2117, opened 2026-09-14
Hide folder-move affordances in the media selector popup

KMchaudhary:
## Summary

Removes the folder-move affordances from the media **selector popup** (e.g. the "Select a Watermark" modal), while leaving them intact on the standalone Media Library page.

Two changes:

1. **Toolbar "Move to folder" bulk action** — previously added to every attachment browser (including picker frames). Now gated to the media grid page (`upload.php`) via the existing `isUploadPage()` check. Picker frames are for choosing an attachment, not filing it, so the button no longer appears there.
2. **Attachment-details "Folder" control** — now skipped on the **GoDAM tab**, whose items are remote SaaS media that aren't yet WordPress attachments and have no local `media-folder` taxonomy to be moved into. The tab is identified by its content mode (`'godam'`), the same signal `godam-media-frame-shared.js` uses.

The Media Library tab and the `upload.php` media page keep their existing folder-move behaviour.

## Screenshots

https://github.com/user-attachments/assets/a0298143-07a1-40ac-8a41-606b9e12a78e

## Testing

- Open a media picker (e.g. GoDAM Settings → Video → Select a Watermark): the toolbar "Move to folder" button is gone on both the Media Library and GoDAM tabs.
- Select an item on the GoDAM tab: the details sidebar no longer shows the "Folder" row.
- Visit the Media Library page (`upload.php`): the toolbar "Move to folder" bulk action is removed and the per-item "Folder" control is unchanged.


subodhr258:
**Reviewed at high effort. Small, well-scoped change, and the core logic holds up.**

I traced both gates end to end:

- **Toolbar bulk action:** `isUploadPage()` reads the `.upload-php` body class, so the button now shows only on the Media Library grid page and is gone from every picker frame. The extended `AttachmentsBrowser` is installed globally (`index.js`), so the grid keeps the button and pickers lose it. This matches the intent.
- **Per-item Folder control:** `view.controller?.content?.mode?.() === 'godam'` is the same signal `godam-media-frame-shared.js` relies on (`this.content.mode() === 'godam'` in several places), and the GoDAM tab is only wired into `MediaFrame.Select`/`Post`, never the `Manage` grid. So this correctly targets GoDAM items only, and the optional chaining keeps it safe when a frame has no `content` region.

No correctness issues. Two things worth a look, both non-blocking:

**1. Rationale mismatch for the Media Library tab inside a picker.**
The toolbar bulk action is removed from pickers because "picker frames are for choosing an attachment, not filing it." That same reasoning applies to the per-item "Folder" control, which also files. But that control still renders on the Media Library tab inside a picker, since change 2 only skips it for `mode === 'godam'`. So a user in the "Select a Watermark" popup can still file a real WP attachment into a folder from the per-item control, even though the bulk button is gone. If that is deliberate (the summary says the Media Library tab keeps its existing behaviour), no change is needed. Flagging it only because the two changes are justified by different reasons.

**2. No tests for the new gates.**
The media-library views already have jsdom coverage (`classic-featured-image.test.js`), so a small test that mocks `wp.media`/`document` to assert the toolbar item is skipped off `upload.php`, and that the field returns early when `content.mode()` is `'godam'`, would lock the behaviour in cheaply and catch a future regression. Optional for a change this size.

subodhr258 (approved):

PR #2121, opened 2026-09-14
Fix: Ensure player styles are correctly enqueued for image shortcode

KMchaudhary:
This pull request addresses a CSS dependency issue that could cause incorrect video sizing when both image and video blocks or shortcodes are present on a page. The changes ensure that the `godam-player-wrapper-style` (which contains the aspect-ratio and poster rules) is always loaded as a dependency for the player and image components, regardless of registration order. This prevents missing styles and layout issues.

**Dependency and style loading improvements:**

* In `class-blocks.php`, the `godam-player-style` block style now explicitly declares a dependency on `godam-player-wrapper-style`, ensuring the wrapper CSS is always loaded, even if style registration order varies between classic and block themes.
* In `class-godam-image.php`, both the `preload_wpbakery_editor_assets` and `render` methods now explicitly enqueue `godam-player-wrapper-style` alongside `godam-player-style` and `godam-image-style`, guaranteeing that the required aspect-ratio and poster rules are present for shortcodes as well as blocks. [[1]](diffhunk://#diff-9a08f22c69706c2363baa48e0442102b5528e4d79f9d25f810197de5937357c6R96) [[2]](diffhunk://#diff-9a08f22c69706c2363baa48e0442102b5528e4d79f9d25f810197de5937357c6R143-R149)

Screenshot 2026-09-14 at 7 42 32 PM

https://github.com/user-attachments/assets/e37bf006-0edf-4efe-9f08-e0660af7848b


subodhr258:
**Reviewed. Correct fix, verified end to end.** I traced the mechanism against WP core and the plugin boot order, and the diagnosis and the fix both hold.

**Why the fix is right**

- On a classic theme, core's `wp_common_block_scripts_and_styles` is hooked on `wp_enqueue_scripts` (priority 10) and registered before any plugin loads, so it runs first in that batch and fires `enqueue_block_assets`. That runs the `wp_enqueue_block_style( 'godam/image', ... )` callback, which registers `godam-player-style` first. `Player::register_scripts()` runs later in the same priority-10 batch, so its `wp_register_style( 'godam-player-style', ..., array( 'godam-player-wrapper-style' ) )` is a silent no-op and the wrapper dep was dropped pre-fix. Declaring `deps` at the block-style site closes the gap. Diagnosis is accurate.
- `godam-player-wrapper-style` is registered by `Player::register_scripts()`, and `GoDAM_Player` boots unconditionally in `class-plugin.php`, so on both the front end (`wp_enqueue_scripts`) and admin (`admin_enqueue_scripts`) the dependency always resolves at print time. No risk of the new hard dep silently dropping `godam-player-style`.
- Build target `assets/build/css/godam-player-wrapper.css` is present. The block-theme path (on-demand assets) registers via `render_block` with the same `deps`, so it is covered too.

**Two small, non-blocking notes**

1. The explicit `wp_enqueue_style( 'godam-player-wrapper-style' )` added in both `class-godam-image.php` methods is now redundant: after this change every registration site of `godam-player-style` declares the wrapper dep, so enqueuing `godam-player-style` alone pulls the wrapper in. It is harmless defense-in-depth and the rationale is documented, so keep it if you want the margin; just flagging that removing those two lines would change nothing.
2. Worth a small PHPUnit test that runs the enqueue hooks and asserts `godam-player-style` carries `godam-player-wrapper-style` in its registered deps. The race is subtle enough that a guard test would catch a future re-registration or hook-order regression cheaply.

LGTM.

subodhr258 (approved):
Approved. Fix verified end to end against WP core and the plugin boot order; the wrapper dependency now resolves at print time in both classic and block themes. The two notes on the review comment are optional and non-blocking.

subodhr258 (dismissed):

subodhr258 (dismissed):
Re-reviewed after the post-approval commit e198496e (`layer-registry.js`). This is a separate fix from the CSS enqueue work: it moves the layer registry into a shared `window` store so every `godam-player` bundle reads the same registrations, instead of each entry bundle keeping a private module-scoped registry. That correctly fixes Woo hotspots not rendering when a `godam/image` and a `godam/video` block share a page (before this, only the bundle that drained `godamLayerRegistryQueue` populated its own copy, and the other bundle's `getLayerManager()` read an empty registry).

Checked:
- Every accessor now routes through `getStore()`; nothing still points at the old module-scoped object.
- The render gate (`layersManager` -> `getLayerManager()` -> `new customManagerClass`) reads the shared store, so the fix reaches the real code path, not just registration.
- No in-repo `addLayerRegistryHook` callers, so sharing the `hooks` array does not cause double-firing.
- `assets/build/` is gitignored, so the src-only change is correct.

Approving. Two non-blocking follow-ups:

1. Please update the PR title/description. They still describe only the CSS enqueue change, not this layer-registry fix, which is the larger part now.
2. Minor/latent: `__godamLayerRegistryStore` is an unversioned key, so it is shared across bundle versions that co-load during an upgrade (old cached bundle + newly served bundle on the same page). No issue today since the store shape is stable; worth namespacing the key by version if that shape ever changes.

On the Copilot note about `class-godam-image.php` enqueuing the wrapper style inside `render()`: styles enqueued during `the_content` are printed in the footer, so the CSS is still applied, and all three styles there share the same timing. Fine as is; an earlier frontend hook would only remove a possible brief layout shift.

subodhr258 (approved):
Approving. Reviewed both changes end to end and the fix is correct.

**`class-blocks.php`** Declaring `godam-player-wrapper-style` as a dep of the `godam/image` block's `godam-player-style` registration correctly mirrors the canonical registration in `class-godam-player.php`. On a page carrying both a `godam/video` and a `godam/image` block, `enqueue_block_assets` registered `godam-player-style` first on classic themes, so the canonical `wp_register_style()` (which carries the wrapper dep) became a silent no-op and the wrapper CSS was dropped, leaving the video poster at its natural oversized height. Declaring the same dep in both places makes it registration-order independent. The wrapper handle is registered unconditionally by the always-booted player shortcode class, so it resolves at print time, and WordPress drops an unregistered dep silently, so there is no regression risk.

**`layer-registry.js`** Moving the store to a `window`-backed singleton is the right fix and at the right altitude. The player build and the image-layers build each carry their own module copy of this file; with a module-local store, if the image build won the init race the Woo manager registered into the image build's store while the player's own `getLayerManager()` read an empty store, so Woo hotspots never rendered on the video. Routing every accessor through `getStore()` makes all copies read and write one shared store regardless of bundle load order. The non-browser fallback is fine, and nothing depends on the old store reference.

Non-blocking notes:

1. The PR description still documents the `class-godam-image.php` enqueue approach, which was reverted in "Remove redundant player wrapper style enqueue in GoDAM Image shortcode". Please update the description to the current two-part fix (block-style dep plus shared registry store) so the squash-merge record is accurate.
2. "Run PHPCS inspection" is red because the action hit the GitHub API rate limit (`x-ratelimit-remaining: 0`) and aborted before scanning, not because of a violation. PHPCS is clean locally on both files. Re-run that job to get it green before merging.
3. Optional: consider a "Fix:" changelog line for v2.2.2, since the oversized poster and the missing Woo hotspots are user visible.

PR #2124, opened 2026-09-15
feat(cta-layer): add `rtgodam_cta_html_allowed_tags` filter for CTA content sanitization

Pathan-Amaankhan:
## Summary

- Adds a new `rtgodam_cta_html_allowed_tags` filter that lets developers extend the list of HTML tags/attributes allowed in the GoDAM CTA layer's rendered content.
- The CTA HTML was previously sanitized with `wp_kses_post()`, which uses a fixed 'post'-context allowlist that doesn't cover every tag a legitimate CTA shortcode might output (e.g. `<form>` for a placeholder that a form-embed script hydrates client-side).
- The filter starts from `wp_kses_allowed_html( 'post' )` as its default and is applied via `wp_kses()` instead, so existing behavior is unchanged unless a developer hooks into the new filter.
- **Security note:** `wp_kses()` is a pure allowlist with no hardcoded blocklist, so this sanitizer is the last line of defense for CTA HTML at render time. Adding tags/attributes via this filter (e.g. `script`, `iframe`, `on*` event attributes) removes the protection the defaults provide — the docblock now calls this out explicitly so hook implementers understand they own that risk.

## Test plan

- [ ] Add a CTA layer with default settings and confirm HTML output is unchanged (still sanitized against the 'post' context defaults).
- [ ] Hook into `rtgodam_cta_html_allowed_tags` to allow an additional tag (e.g. `form`) and confirm it now renders in the CTA layer output.
- [ ] Confirm disallowed/unsafe tags (e.g. `<script>`) are still stripped by default — this no longer holds once a hook implementer adds them via `rtgodam_cta_html_allowed_tags`; that tradeoff is documented in the filter's docblock.


KMchaudhary:
## Code review — `rtgodam_cta_html_allowed_tags` filter

Nice, minimal change. I confirmed the default path is behaviorally equivalent to the old `wp_kses_post()`: `wp_kses( $html, wp_kses_allowed_html( 'post' ) )` resolves to the same 'post'-context allowlist (kses converts the `'post'` string to that same array internally via `wp_kses_split2()`), so unhooked sites see no difference. 👍

A couple of things worth addressing before merge:

### 1. Security framing of the escape hatch (the main thing to weigh)
`wp_kses()` is a pure **allowlist** — it has no hardcoded blocklist for `<script>`. So a developer who hooks `rtgodam_cta_html_allowed_tags` and adds `script` (or `iframe`, or an `on*` event attribute) will have that tag/attribute pass straight through. The test-plan bullet *"Confirm disallowed/unsafe tags (e.g. `<script>`) are still stripped regardless of the filter"* only holds while nobody adds them via the filter — it's not an inherent guarantee anymore.

That's the intended nature of an escape hatch, but this sanitizer is the **last line of defense** for CTA HTML at render time. Recommend:
- Adding a docblock warning that the filter can widen sanitization to unsafe tags and that hook implementers own that risk (mention it can effectively grant `unfiltered_html`-like output).
- Softening the test-plan wording so it doesn't imply the safety guarantee survives the filter.

### 2. `@since n.e.x.t` is inconsistent with the repo
Every other `@since` in this file — including the sibling `rtgodam_cta_html_content` filter right above (`@since 2.2.0`) — and all ~531 `@since` tags across `inc/` use a concrete version number; `n.e.x.t` appears nowhere else in the codebase. As written it will ship the literal string `n.e.x.t` in a release. Please replace it with the target version (e.g. the next planned release).

Otherwise the change looks correct and well-scoped.

Pathan-Amaankhan:
> 1. Security framing of the escape hatch (the main thing to weigh)

Have updated the comment as suggested.


> 2. @since n.e.x.t is inconsistent with the repo

this one's intentional. We use `n.e.x.t` as a placeholder `@since` tag for unreleased code across this codebase; it gets search-and-replaced with the actual version number as part of the release cut, rather than being filled in by the PR author (who doesn't know the exact upcoming version at merge time). No change needed here.

PR #2125, opened 2026-09-15
Fix: raw CSS leaking into the What's New page (inline <style> from release post)

subodhr258:
## Problem

On the **What's New** admin page (`admin.php?page=rtgodam_whats_new`), the *"Preview in one click"* feature showed a block of **raw CSS as visible text** instead of prose. (Reported by Kishan.)

The leaked text was GoDAM's own video-placeholder / loading CSS:

```
.godam-video-placeholder{aspect-ratio:var(--rtgodam-video-aspect-ratio,16/9);position:relative}...@keyframes godam_pulse_animation{50%{opacity:.8}}
```

## Root cause

A plugin rendering bug, triggered by what the release post happens to contain:

1. The godam.io release post (fetched by `godam/v1/release-post`) embeds a GoDAM video block. WordPress renders that block's registered **inline `<style id="godam-player-wrapper-inline-css">`** directly into `content.rendered`, sitting between the section's paragraphs.
2. `get_release_post()` ran the whole content through `wp_kses_post()`. `<style>` is not an allowed post tag, so kses **strips the `<style>`/`</style>` tags but keeps their inner text** — turning the CSS into loose text.
3. `parse_features_from_content()` loads that into `DOMDocument`, which **re-wraps the loose CSS in an implicit `<p>`**.
4. The parser collects `<p>`/`<ul>`/`<ol>` siblings into the feature `description`, so the CSS `<p>` got folded in.
5. The React page renders `feature.description` via `dangerouslySetInnerHTML`, so the CSS appeared as text.

The core mistake: **sanitizing (a lossy transform) before structural parsing.** `wp_kses_post()` converts a `<style>` element into free-floating text that then reads as content.

## Fix

Remove the sanitize-before-parse ordering rather than patch around it:

- `parse_features_from_content()` now parses the **raw** `content.rendered` and sanitizes each **extracted field** (`title` → `sanitize_text_field`, `description` → `wp_kses_post`, `image` → `esc_url`). Structure stays intact, and since only `<p>`/`<ul>`/`<ol>` and `<figure>` images are collected, a sibling `<style>`/`<script>` is simply ignored. This also covers other disallowed-but-text-bearing elements WP can render inline (`<noscript>`, etc.).
- As defence in depth, `<script>`/`<style>`/`<noscript>` **nodes** are dropped from the DOM up front, so their text can't leak even if one is nested inside a collected element, and an unclosed `<style>` (which `DOMDocument` still parses into a node) is handled too.

No regex is used, so the earlier concerns about only-closed-tags and `preg_replace()` returning `null` don't apply.

## Testing

- `tests/php/ReleasePostFeatureParsingTest.php` exercises `parse_features_from_content()` (the method that existed pre-fix) via reflection. The `wp_kses_post` stub reproduces the real tag-stripping behind an opt-in flag, so existing tests are unaffected.
- The nested-`<style>` and nested-`<script>` cases **fail red against the original parser** and pass with the fix; sibling, unclosed-tag, and plain-content cases are also covered.
- Verified locally: full suite green (92 tests, 278 assertions); `phpcs` clean on all changed files.

## QA / deploy note

`get_release_post()` caches its result in the **`rtgodam_release_data`** transient with **no expiry**, so a site that already rendered the leak keeps serving the cached copy until it is cleared. To verify on such a site:

```bash
wp transient delete rtgodam_release_data
```

Then reload the What's New page.

## Out of scope (follow-up)

That `rtgodam_release_data` transient has no expiry, so the What's New page also won't pick up newer release posts until it is cleared. Worth a separate look, not fixed here.


KMchaudhary:
## Code review — `fix/whats-new-inline-style-leak`

Solid, well-documented, minimal fix with a real regression test. The strip regex handles the actual `<style id="godam-player-wrapper-inline-css">…</style>` case correctly (including attrs with `>`, uppercase, and multiple/mixed blocks). A few things worth a look, none blocking:

**1. Altitude — sanitize-before-parse is patched, not removed (`class-release-post.php`, `build_features()`)**
The PR itself nails the root cause: *"Sanitizing (a lossy transform) before structural parsing is the core mistake."* The fix keeps that order (regex-strip → `wp_kses_post` → `DOMDocument`) and special-cases `<script>/<style>`. The deeper fix is to run `DOMDocument` on the **raw** `content.rendered` first — it treats `<style>/<script>` as element nodes, and `parse_features_from_content()` only collects `<p>/<ul>/<ol>/<figure>`, so the CSS is never gathered — *then* `wp_kses_post()` each extracted description. As written, any other disallowed-but-text-bearing element WP renders inline (`<noscript>`, `<template>`, `<title>`) would leak via the exact same kses-strips-tag-keeps-text → DOM-rewraps-in-`<p>` mechanism.

**2. Correctness — regex only strips *closed* blocks**
`#<(script|style)\b[^>]*>.*?</\1>#is` requires a matching close tag. An unclosed/truncated `<style>` (no `</style>`) is left in place, then `wp_kses_post()` strips the lone tag but keeps its inner CSS — reproducing this very leak. Low likelihood given WP block output is normally well-formed, but it's the same failure mode the fix targets.

**3. Correctness — `preg_replace` can return `null`**
On pathologically large content that hits `pcre.backtrack_limit`/`pcre.recursion_limit`, `preg_replace()` returns `null` → `wp_kses_post((string) null)` → `''` → zero features rendered, silently. Cheap guard: fall back to the original string when the result is `null`.

**4. Test — "red without fix" is weaker than described**
The test exercises `build_features()` via reflection, but that method didn't exist pre-fix, so on the old code the reflection call throws rather than asserting the CSS-leak failure. The committed test pins the new behavior well, but doesn't itself fail-red against the original `parse_features_from_content()` path, so the *"genuine red-without-fix guard"* claim is a touch overstated. A test that drives the pre-fix path directly would prove the guard.

*Nice touches:* the opt-in `kses_strip_disallowed` stub flag keeping existing tests unaffected, and the `tearDown()` `unset()` matching the existing `VideoEditorTest`/`DocumentSupportTest` convention.

subodhr258:
Thanks for the thorough review, @KMchaudhary. Agreed on all four; the altitude point is the right call, and fixing it there dissolves 2 and 3 as well. Pushed ca45ef83.

**1. Altitude (sanitize-before-parse).** Adopted. `parse_features_from_content()` now parses the **raw** `content.rendered` and sanitizes each **extracted field** (`title`/`description`/`image`), instead of `wp_kses_post()`-ing the whole document first. With structure intact and only `<p>`/`<ul>`/`<ol>`/`<figure>` collected, a sibling `<style>`/`<script>` (and `<noscript>`, etc.) is ignored outright, so the kses-strips-tag-keeps-text → DOM-rewraps-in-`<p>` mechanism can't fire. I also drop `<script>`/`<style>`/`<noscript>` **nodes** from the DOM up front as defence in depth, for the nested-inside-a-collected-element case.

**2. Regex only stripped closed blocks.** The regex is gone entirely. An unclosed `<style>` is now parsed by `DOMDocument` into a node and removed with the rest, so it no longer survives to be half-stripped. Added `test_unclosed_style_block_does_not_leak` for it.

**3. `preg_replace` can return `null`.** Moot now that there's no `preg_replace`.

**4. "Red without fix" was overstated.** Fair. The tests now drive `parse_features_from_content()` (which existed pre-fix) via reflection, not a method introduced by the PR. The nested-`<style>` and nested-`<script>` cases fail red against the original parser and pass with the fix (verified by reverting the node-removal locally). The sibling case documents the exact production scenario, and there are unclosed-tag and plain-content cases too.

Full suite green (92 tests, 278 assertions), phpcs clean.

KMchaudhary (approved):

PR #2128, opened 2026-09-16
Test: cover the What's New parser caller boundary

subodhr258:
Follow-up to review feedback on #2125 / #2126 (KMchaudhary and Copilot): the parser tests invoked `parse_features_from_content()` with raw HTML, so they could not catch a reintroduced *sanitize-before-parse* in the caller.

Adds a test that drives the public `get_release_post()` with a mocked remote response and a forced cache miss (new HTTP/transient/`WP_REST_Response` stubs in `tests/bootstrap.php`). It asserts the real production case (a sibling `<style>`) stays clean end to end.

Verified: reintroducing `wp_kses_post()` before the parse call makes this test fail red; with the current code all 94 tests pass, phpcs clean.


subodhr258:
Folded into #2127 (the v2.2.3 release-prep PR) as a single PR. Same commit, now on `chore/bump-2.2.3`.

PR #2130, opened 2026-09-17
Fix: Video Block Preview in WP Bakery

### LINKED ISSUE #1590: Video and Audio block preview issue in WPBakery editor
## Bug Report

**Current Behavior**
1. When the GoDAM audio block is added in the editor with the auto-replay setting enabled, it is very difficult to stop the audio. Even after clicking the pause button, it takes two to three attempts for the music to stop.
(Looks like two audio sources started playing wherethe user enables the auto-play setting. 1) is from the audio preview itself 2) is unknown)

2. When a video block is added in the editor and a video is selected, the block still appears blank. This gives the impression that the video has not been selected, even though it has been.

Image

**Expected behavior/code**
1. Autoplay should only be played from one source in the WPBakery editor.
2. Video preview should be shown on the WPBakery editor for a better user experience.

**Possible Solution**
We need to reach out to the WPBakery support team for the video preview issue.
cc: @hashibhere @joelabreo227

rudrakshi-gupta:
## Issue

Fixes #1590

## Root Cause

The `godam_video` shortcode renders a static poster placeholder plus a hidden `video` that video.js turns into the real player once godam-player-frontend-script runs. WPBakery's inline editor injects a freshly-added element via AJAX but never injects the newly-enqueued script into the running preview — so on a page that had no player before, the script was absent and the video stayed frozen on its poster (or the loading placeholder with no poster). A full reload re-rendered the shortcode in content with the script present, which is why the live player only appeared after refresh.

## Fix

Added preload_wpbakery_editor_assets() to class-godam-player.php, hooked on wp_enqueue_scripts (priority 20) and guarded by vc_is_inline(). It enqueues the player script, styles, and active skin up front in the inline editor. The JS side already watches for builder-injected markup (isBuilderPreview() + MutationObserver → initializePendingVideos()), so now a just-added video initializes right away. This mirrors the existing GoDAM Image element's fix for the same WPBakery quirk. Editor-only — the published front end is untouched.

### Before

https://github.com/user-attachments/assets/0275750d-5368-47c1-a558-a57233e833ab

### After

https://github.com/user-attachments/assets/6bfff00b-1cbb-4faa-872f-b34ce19a25a2


KMchaudhary:
### Code review — correctness + cleanup (medium effort)

Editor-only, additive change; mechanism is sound. Two things worth addressing before merge.

---

**1. `godam-player-analytics-script` should not be preloaded in the editor** — `inc/classes/shortcodes/class-godam-player.php:321` · *data integrity*

`preload_wpbakery_editor_assets()` enqueues the analytics script in the inline editor. Once `frontend.js` initializes the just-added player, the analytics plugin attaches and starts tracking. The only guard, `shouldSkipAnalytics()`, gates on:

- `isAdminPage` → `is_admin()`, which is **false** on the WPBakery frontend inline-editor page, and
- `?godam_page=video-preview`, which isn't set here.

So analytics is **not** skipped, and every preview play/scrub in the editor sends real view/engagement events to the SaaS endpoint — corrupting the video's stats. Before this PR the player never initialized in the editor, so nothing fired; the fix reintroduces the player *and* wires up tracking with it.

Note the GoDAM Image element's `preload_wpbakery_editor_assets()` deliberately loads only its layers script + styles, **not** any tracking script. Suggest dropping `godam-player-analytics-script` from the preload (or, if editor analytics is genuinely wanted, having `shouldSkipAnalytics()` also return `true` under `vc_is_inline()`).

---

**2. Skin lookup is copy-pasted from `render()`** — `inc/classes/shortcodes/class-godam-player.php:326-334` · *reuse*

The settings read + `$skins` map + `isset()` enqueue is duplicated verbatim from `render()` (lines 451-462). The skin list now lives in two places — add/rename/remove a skin and you must edit both, or the editor preview silently diverges from the front end. Worth extracting a small private helper (e.g. `enqueue_active_skin_style()`) and calling it from both.

---

*🤖 Generated with [Claude Code](https://claude.com/claude-code)*

KMchaudhary (approved):

PR #324, opened 2025-05-07
Add admin notice when CDN conflict detected with GoDAM transcoding service.

KMchaudhary:
- Add admin notice when CDN conflict is detected with GoDAM transcoding service.
- Remove the email notification on video transcoding failure.

PR #323, opened 2025-05-06
Fix share button style

prachigarg19:
Updates style for share button:

Desktop:
image
image

Mobile:
image
image

Issue #321, opened 2025-05-05
Release on WordPress.org


nayemDevs:
Please post the latest release changelog as a comment.

subodhr258:
Changelog:
```
= v1.0.4 (May 7, 2025) =

- Feat: Introduced the new GoDAM Audio Block for displaying audio content from CDN.
- Feat: Redesigned the Video Editor interface for a smoother editing experience.
- Feat: Refreshed the UI of Settings pages for improved usability.
- Feat: Added support for interactive Polls layer in video editor.
- Fix: Added CDN detection to prevent conflicts with GoDAM’s transcoding.
```

KMchaudhary:
= v1.0.5 (May 15, 2025) =

- Fix: Updated plugin to support translations and localization.
- Enhancement: Added compatibility for displaying thumbnails in the rtMedia gallery view.

KMchaudhary:
## v1.0.6 (May 16, 2025)
- Fix: Updated translation files.

Issue #320, opened 2025-05-05
Speech to Text using GoDAM Transcript/AI feature


nayemDevs:
We can provide Text to speech generation feature on our GoDAM plugin as we are developing AI trnascrpit functionality.

nayemDevs:
@prachigarg19 Please add WordPress integration with AI transcription next week

hbhalodia:
- Left some feedbacks on PR, Please resolve that.

Thank You,

renishsurani:
I have verified the transcription plugin feature, and the English captions are properly visible for the video.

I have checked [this](https://github.com/rtCamp/godam/actions/runs/17259469030) feature for this published plugin version.

Note: While correcting captions from the central, the changes are not getting updated. I will track all the central issues separately.

<img width="1240" height="707" alt="Image" src="https://github.com/user-attachments/assets/7e685d39-1f2b-40d5-98ad-e1f4d10c193d" />

kishan-gondaliya-7270:
Released with 1.4.0

PR #2131, opened 2026-09-18
Fix Dependabot npm dev-dependency security alerts

subodhr258:
## What this does

Resolves the fixable **npm** Dependabot alerts on the plugin by bumping the pinned versions in the `overrides` block to the patched releases. Every affected package is a **development-scope, transitive** dependency (build/test tooling) pulled in through `package-lock.json`; no production dependency and no shipped plugin code is touched.

The repo already pins these transitive deps via `overrides`, but several pins had fallen one patch behind the version a newer advisory now requires (e.g. `svgo` pinned at `3.3.4` while the fix is `3.3.5`). This PR moves each pin forward.

## Alerts addressed (9 of 14 open)

| Package | Was | Now | Advisory | Alerts |
|---|---|---|---|---|
| js-yaml (v3) | 3.15.1 | **3.15.2** | GHSA-2883-xcg3-v3hh | #324 |
| js-yaml (v4) | 4.3.1 | **4.3.2** | GHSA-2883-xcg3-v3hh | #323 |
| svgo (v3) | 3.3.4 | **3.3.5** | GHSA-w27v-7q3p-w38r, GHSA-4vpr-x523-8j87 | #321, #320 |
| svgo (v4) | 4.0.2 | **4.1.0** | GHSA-w27v-7q3p-w38r, GHSA-4vpr-x523-8j87 | #316, #317 |
| adm-zip | 0.6.0 | **0.6.1** | GHSA-vwc7-r8mq-g2x9 | #318 |
| joi | 18.0.2 | **18.2.9** (new override) | GHSA-gg4h-3hg2-grpc, GHSA-6w3j-5fw6-r9vr | #315, #314 |

Note on adm-zip: the advisory lists no `first_patched_version`, but its affected range tops out at `<= 0.6.0` and upstream published `0.6.1` (2026-09-11), which sits outside the range — so the bump clears it.

## Not fixable in this PR (5 alerts, no upstream patch exists)

These are also dev-only and transitive, but **every published version is still within the vulnerable range**, so there is no version to bump to:

| Package | Latest published | Vulnerable range | Comes from | Alerts |
|---|---|---|---|---|
| extract-zip | 2.0.1 | `<= 2.0.1` | `@wordpress/env` (1.7.0), `puppeteer-core` → `@puppeteer/browsers` (2.0.1) | #319, #299 |
| showdown | 2.1.0 | `<= 2.1.0` | `@wordpress/blocks` (1.9.1) | #292, #291, #161 |

Both are used only at build/test time and only extract archives from trusted sources (the WP download, the pinned Chrome build) — they are not reachable from shipped plugin code. **Recommendation:** dismiss these five as *"No bandwidth to fix / vulnerable code is not actually used"* until upstream ships a fix, or let them auto-resolve when `@wordpress/*` / puppeteer drop the dependency.

## Verification

- `npm install` regenerates `package-lock.json`; only dev-scope entries change (a `css-what` re-dedupe under `svgo`), no production dependency moved.
- `npm run build:prod` — compiles successfully (exit 0), so the `svgo` and `adm-zip` bumps do not break the webpack build or `wp-scripts` packaging.
- `npm audit` confirms the 6 packages above no longer appear; only the two no-fix packages remain.

## Manual QA steps

1. `git fetch && git checkout fix/dependabot-npm-dev-deps`
2. `npm ci` (clean install from the updated lockfile) — should complete with no errors.
3. `npm run build:prod` — all entrypoints should report *"compiled successfully"*.
4. `npm run start` — dev build should watch and compile with no new errors.
5. Spot-check that admin/editor screens and blocks render as before (svgo optimizes SVGs at build time; adm-zip is only used by `wp-scripts plugin-zip`).
6. Optional: `npx wp-scripts plugin-zip` (or the repo's zip step) to confirm packaging still works with adm-zip 0.6.1.

## Notes

- Dependabot scans the **default branch (`main`)**; the 9 alerts will close once this reaches `main`, not on merge to `develop`.

## Lockfile note

`package-lock.json` was regenerated with **Node v22.14.0 / npm 10** (matching `.nvmrc`, which is what CI uses) so that `npm ci` stays in sync. The `joi` override is unscoped (`"joi": "18.2.9"`) rather than range-scoped, because npm 10 does not upgrade an already-locked in-range version through a range-scoped override key.

---

## Review follow-ups (@KMchaudhary)

**`joi` override range-scoped** (`18a3abdf1`): `"joi": "18.2.9"` -> `"joi@<18.2.9": "18.2.9"`, so it no longer clamps every future joi major. Confirmed it still upgrades develop's locked `18.0.2`, and that `package-lock.json` is byte-identical either way.

**svgo v4 output verified**: clean `npm ci && npm run build:prod` on develop (svgo 3.3.4 / 4.0.2) and on this branch (svgo 3.3.5 / 4.1.0); `assets/build` is byte-identical across all 314 files. The `css-select 5→6` / `css-what 6→7` cascade changes no optimized output.


KMchaudhary:
## Code review

Reviewed the `package.json` override changes and the regenerated `package-lock.json`. Overall this is a clean, well-scoped, well-documented security bump. I independently verified:

- ✅ Every new lockfile `integrity` hash matches the public npm registry (js-yaml 3.15.2 / 4.3.2, svgo 3.3.5 / 4.1.0, adm-zip 0.6.1, joi 18.2.9, sax 1.6.1).
- ✅ No leftover vulnerable versions remain in the lockfile (0 occurrences of js-yaml 3.15.1/4.3.1, svgo 3.3.4/4.0.2, adm-zip 0.6.0, joi 18.0.2).
- ✅ Transitive spec consistency: joi 18.2.9 now requires `@standard-schema/spec@^1.1.0`, and the lockfile hoists `@standard-schema/spec@1.1.0` (satisfies both that and the other `^1.0.0` consumer).
- ✅ Override range math is internally consistent — e.g. `adm-zip@<=0.6.0` correctly catches the previously-locked `0.6.0` (the old `<0.6.0` key did not), and each `<x` upper bound excludes its own target so the pins stay stable.

Two non-blocking notes:

### 1. `joi` override is unscoped (`package.json` L127)
Every other pin in the block is range-scoped, but `"joi": "18.2.9"` is not. Today the only consumer is `wait-on` (`^18.0.1`) so it's harmless, but an unscoped key permanently clamps **all** future joi resolutions to 18.2.9 regardless of major (18.2.9 requires `node >= 20`). If a future dev-dep pulls joi 17.x/19.x it gets silently forced to 18.2.9.

The PR notes range-scoping was avoided because npm 10 won't upgrade an already-locked *in-range* version. An **exclusive-bound** key sidesteps both problems and matches the pattern used for svgo/adm-zip:
```jsonc
"joi@<18.2.9": "18.2.9"
```
This catches the locked 18.0.2 (`18.0.2 < 18.2.9`) while leaving other majors alone. Optional — the current form is fine for now.

### 2. `svgo` v4 is a minor bump, output not verified (`package.json` L149)
`svgo@>=4.0.0 <4.1.0 → 4.1.0` is the only non-patch bump (also pulls css-select 5→6, css-what 6→7). A minor upgrade of an SVG optimizer can change its output. Build compiles (✅), but that doesn't confirm optimized SVG/CSS output is unchanged. Suggest a quick visual/byte diff on a couple of optimized assets before merge. Low risk since it's build-time only and isolated under `css-minimizer-webpack-plugin`.

Nothing here blocks merge. 👍

KMchaudhary (approved):

subodhr258:
Thanks @KMchaudhary. Both notes are addressed.

### 1. `joi` override is now range-scoped — done (`18a3abdf1`)

Took your suggestion exactly:

```diff
-		"joi": "18.2.9",
+		"joi@<18.2.9": "18.2.9",
```

Verified both properties that matter:

- **It still catches the vulnerable version.** Took develop's `package-lock.json` (which locks `joi@18.0.2`), paired it with this branch's `package.json`, and regenerated. Result: `joi 18.0.2 -> 18.2.9`. The exclusive bound does the job.
- **It changes nothing here.** Regenerating this branch's lockfile after the edit produces a byte-identical `package-lock.json` — `diff` is empty. Resolution is `joi@18.2.9` either way, and the only consumer is still `wait-on` (`^18.0.1`).

So other majors are now free, as you wanted.

### 2. svgo v4 output — verified unchanged, full build tree byte-identical

Built both sides from a clean `npm ci` and compared the entire `assets/build` output.

| | develop | this branch |
|---|---|---|
| `svgo` (v3 line) | 3.3.4 | **3.3.5** |
| `svgo` (v4 line) | 4.0.2 | **4.1.0** |
| `postcss-svgo` | 6.0.3, 7.1.1 | 6.0.3, 7.1.1 |
| `@svgr/plugin-svgo` | 8.1.0 | 8.1.0 |

Only svgo moved; everything around it is pinned identically, so the comparison isolates exactly the bump you flagged.

```
npm ci && npm run build:prod      # both trees, webpack 5.105.4, compiled successfully
diff -rq godam-svgo-base/assets/build  godam-wt-dependabot/assets/build
-> 0 differing files (314 files each side)
```

The `css-select 5→6` / `css-what 6→7` cascade does not change a single byte of optimized output. Good call asking — worth having the evidence on record rather than inferring it from a clean compile.

PR #2134, opened 2026-09-18
fix(analytics): funnel hover tooltips, Purchased tier split, docs link, full titles on hover

subodhr258:
## What changed

Four analytics dashboard fixes from QA:

1. **Purchase Funnel: hover tooltips on every bar segment.** Each coloured segment explains itself on hover (WP `Tooltip`, same component as the timeline's info tips; the text is also on `aria-label`): the tier, its count and its share of that stage, e.g. `Assisted only (clicked out): 37 of 64 added to cart (57.8%)`. The grey remainder reads `Did not reach this stage: N of M players`.
2. **Purchase Funnel: the Purchased bar now shows its Direct / Assisted split.** It was painted entirely Direct blue, so an Assisted buyer showed in the Added-to-cart split but disappeared on Purchased. Uses the new `purchased.direct` / `purchased.assisted` from rtCamp/godam-analytics#324; on an older service (no split) the bar stays a single segment instead of guessing a tier.
3. **Dashboard head: "Analytics docs ↗"** link to https://godam.io/docs/godam-analytics/, styled like the existing Reel Pop link.
4. **Top Videos / Top Products: full title on hover.** The title cell truncates with an ellipsis; it now carries the full name as a native `title` attribute. (Tried the WP `Tooltip` here first; it never opens on these cells because the tables re-render periodically and reset its show delay, so the native attribute is the reliable choice.) Titles are also passed through `decodeEntities`, since the proxy returns them texturized (`&#8211;`) and both the cell and the tooltip rendered the entity literally.

    **Accessibility note:** a native `title` shows on pointer hover only. It is *not* revealed by keyboard focus, so keyboard-only sighted users still cannot read a truncated name. That gap predates this PR (the cell had no tooltip at all before) and is not closed by it. An earlier commit message on this branch says the title "works with keyboard focus on the link" — that is wrong; disregard it. If we want real keyboard parity, the follow-up is a focus-revealed tooltip or a visually-hidden full-text node.

## Tests

`PurchaseFunnelCard.test.js`: +3 (segment hover text with shares; Purchased split into both colours; single-segment fallback without the split). Existing dashboard suites unchanged. 75 passed. ESLint / stylelint clean.

Verified live on godam-dev.local against rtCamp/godam-analytics#324 on `:8010`, with one Direct buyer and one Assisted buyer seeded (all test data removed afterwards).

## Screenshots

**Dashboard head: "Analytics docs ↗" next to the Reel Pop link**

**Purchase Funnel: Purchased bar split (dark = Direct, light = Assisted) with the hover tooltip on the Assisted share**

**Top Products: truncated name, full name on hover (native title; browser tooltips are not capturable in screenshots)**

**Top Videos: same, truncated "Swimming Lesson 3 …" carries the full title**

## Manual QA

1. Dashboard head shows "Analytics docs ↗" next to the Reel Pop link; opens the docs in a new tab.
2. Purchase Funnel: hover the dark-blue and light-blue parts of "Added to cart" and "Purchased" → tooltip with tier, count and share; hover the grey part → "Did not reach this stage".
3. With one in-video buyer and one click-out buyer in range, the Purchased bar shows dark blue + light blue.
4. Top Videos / Top Products: hover a truncated title → full title.


elifvish (commented):
Reviewed against a real godam-analytics instance (#323 + #324 merged, ClickHouse in Docker).

**Verified**
- Jest: 429 passed, 29 suites.
- The older-service fallback works: with no `purchased.direct` the bar stays a single segment rather than guessing a tier.
- Rendered the card against a live `/dashboard/video-funnel/` response (`purchased: {count: 2, direct: 1, assisted: 1}`) rather than a fixture — both tiers paint, tooltips read `Direct (added in-video): 1 of 2 purchased (50.0%)` and `Assisted only (clicked out): 1 of 2 purchased (50.0%)`.
- Confirmed the plugin's own REST proxy does not filter the new keys (`inc/classes/rest-api/class-analytics.php:1646` passes `video_funnel` through verbatim, no registered response schema).
- Head CSS is right: `.godam-dashboard-head` has `gap: 8px 16px`, so resetting `margin-left` on the second link spaces correctly, and the docs link still pushes right on its own when the Reel Pop link is absent.
- `@wordpress/html-entities` moving to `dependencies` is correct — wp-scripts' `DependencyExtractionWebpackPlugin` externalises it to `wp-html-entities`, so the version bump has no bundle impact.

Two small things, neither blocking.

**1. `alt` is not decoded, so the entity fix is half-applied** — `pages/dashboard/components/TopVideosTable.js:426` and `:438`

Both `` still receive the raw title, so a screen reader reads `Swimming Lesson &#8211; 3` while the adjacent text node and the new `title=` attribute are decoded. Same one-line `decodeEntities()` as the others. This is squarely within what the last commit set out to fix.

**2. Zero-count segments are still keyboard tab stops** — `pages/analytics/PurchaseFunnelCard.js:37`

Segments render unconditionally, so a zero-count tier produces a `width:0%` div that is still `tabIndex={0}` with an `aria-label`. Rendering a funnel with `added_to_cart.count = 0` and `purchased.direct = 0`, the cart bar emits 2 focusable zero-width segments and the Purchased bar 1 — keyboard users tab onto invisible targets that announce `Direct (added in-video): 0 of 0 added to cart (0.0%)`. Filtering out zero-count segments before the `map` would fix it.

Related: the grey remainder has `role="img"` + `aria-label` but no `tabIndex`, so it is reachable by screen reader but not by keyboard — inconsistent with the coloured segments in the other direction.

**Scope note, not a defect:** four unrelated fixes in one PR, and only the tier split depends on #324. The docs link and the entity decoding could ship independently of the service, which would give finer revert granularity if the funnel work ever needs backing out.

subodhr258:
Thanks. Both addressed in de02d2d0f:

1. **`alt` not decoded** (`TopVideosTable.js`) — both `` now wrapped in `decodeEntities()`, so the entity fix is applied to the alt text as well as the visible node and `title`.
2. **Zero-count focusable segments** (`PurchaseFunnelCard.js`) — segments are now filtered to `frac > 0` before the map, so a zero-count tier no longer renders a `width:0%` `tabIndex=0` div. Added a test (funnel with zero-count Direct and zero-count Assisted) that asserts neither phantom segment renders while the nonzero tiers still do; verified it fails against the unfiltered code.

On the related note: the grey remainder stays screen-reader-only (aria-label, no tabIndex) by design, and the coloured segments are now keyboard-focusable only when they represent real units, so there are no invisible tab stops in either direction.

Scope note taken: for this cycle I've kept the four fixes together, but agree the docs link and entity decode are independent of the service if we ever need finer revert granularity.

Suite 15 pass (76 across the dashboard suites), lint clean.

elifvish (approved):

